PR #12, opened 2015-04-07
Pass SyntheticEvent to onPrev/NextMonthClick callbacks

### LINKED ISSUE #10: Call stopPropagation on nativeEvent on next and previous month click
In `handleNextMonthClick` and `handlePrevMonthClick` event propagation is not being stopped because `stopPropagation` is being called on the synthetic event, not on the native event. 

https://github.com/gpbl/react-day-picker/blob/master/src/DayPicker.js#L58 and https://github.com/gpbl/react-day-picker/blob/master/src/DayPicker.js#L70 should be:

```js
e.nativeEvent.stopPropagation();
```

Alternatively, you could pass the event to the `onNextMonthClick` callback (as you do with `onDayClick`) and let the component user decide whether to stop propagation or not. I'd be happy to PR either fix. Let me know which you prefer.

iancmyers:
Pass the SytheticEvent to the onPrevMonthClick callback and the onNextMonthClick callback so that the developer can access it as needed.

Updated the corresponding documentation in the README.

My editor also automagically cleaned up some whitespace. I can undo that if you'd like.

Fixes #10

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ npm install react-day-picker --save
 
 ### Use of modifiers
 
-This date picker works with modifiers, as in [BEM-like syntax](http://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/). You set the modifiers as functions returning `true` or `false`. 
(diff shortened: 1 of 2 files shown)


gpbl:
Thanks! (also for the whitespaces :) )

Issue #11, opened 2015-04-06
Support rendering multiple months at the same time


iancmyers:
Would you be open to a PR that added support for rendering multiple months side-by-side?

gpbl:
:+1:  How could the API work? I'd deprecate `initialMonth` and add `startMonth` and `endMonth` props. I'm also thinking that splitting the component in multiple components could help: I mean `<Calendar />`, `<Week/>`, `<WeekHeader/>`, `<Day />`, `<NavButton />` and `<Header />`.
I needed that also to add some unit tests. What do you think?

gpbl:
Or maybe we could leave `initialMonth` and add a `numberOfMonths` prop (defaults to 1)

iancmyers:
I think `initialMonth` and `numberOfMonths` will make the logic for paging through months easier–we'll just have to increment/decrement `month` in the state.

gpbl:
Published in v0.8.0

PR #13, opened 2015-04-13
Adds 'numberOfMonths' option to the DayPicker, allowing multiple months to be displayed side-by-side.

majapw:
In reference to https://github.com/gpbl/react-day-picker/issues/11, @iancmyers @gpbl

I separated out the logic to render a single month into its own method, and simply called it as many times as was indicated by the new property. Each month is its own table, and I added some custom styling to the example css to make the months render side-by-side (which I think is desirable). 

Here are some screenshots of the example with multiple months:
screen shot 2015-04-13 at 3 21 01 pm
3-months-visible

It was super awesome how easy it was to add this functionality! :) Let me know what you guys think? This does not affect the default calendar whatsoever. It still functions exactly as it did before.

### DIFF
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -9,6 +9,8 @@ const DayPicker = React.createClass({
     enableOutsideDays: React.PropTypes.bool,
 
     initialMonth: React.PropTypes.object, // default is current month
+    numberOfMonths: React.PropTypes.number, // default is 1
+
     modifiers: React.PropTypes.object,
 
     onDayClick: React.PropTypes.func,
@@ -22,7 +24,7 @@ const DayPicker = React.createClass({
   },
 
   getDefaultProps() {
-    return { initialMonth: moment(), enableOutsideDays: false };
+    return { initialMonth: moment(), numberOfMonths: 1, enableOutsideDays: false };
   },
 
   getInitialState() {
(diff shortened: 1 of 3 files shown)


src/DayPicker.js
```diff
@@ -92,19 +94,36 @@ const DayPicker = React.createClass({
   },
 
   render() {
-    const { month } = this.state;
+    var { month } = this.state;
```
gpbl: @majapw could you replace `var` with `let`? (here and in the following rows) thanks!

src/DayPicker.js
```diff
@@ -92,19 +94,36 @@ const DayPicker = React.createClass({
...
+  },
+
+  renderMonth(month, monthIndex) {
+    var isFirstMonth = (month === this.state.month);
+    var isLastMonth = (monthIndex === this.props.numberOfMonths - 1);
     return (
       <table className="DayPicker">
         <caption className="DayPicker-caption">
-          { this.renderNavButton('left') }
+          { isFirstMonth ? this.renderNavButton('left') : null }
```
gpbl: Instead of a ternary operator, here I'd write:
```js
 { isFirstMonth && this.renderNavButton('left') }
```

src/DayPicker.js
```diff
@@ -92,19 +94,36 @@ const DayPicker = React.createClass({
...
           { month.format('MMMM YYYY') }
-          { this.renderNavButton('right') }
+          { isLastMonth ? this.renderNavButton('right') : null }
```
gpbl:
```js
 { isLastMonth && this.renderNavButton('right') }
```

gpbl:
@majapw awesome, thanks a lot! Would you mind to update your pull request with the small changes I commented above?

majapw:
Definitely. Made all of the changes that you asked for, @gpbl.

gpbl:
Thank you!

PR #14, opened 2015-04-17
Persist events on month changes

iancmyers:
Because React reuses event objects by the time the event reaches the onPrev/NextMonthClick callback all of the attributes are null.

This change has React persist the events so that they can actually be used in the callback.


gpbl:
Thank you 👍

PR #20, opened 2015-05-12
Allow className attribute

zzarcon:
Allow to pass **className** to the component

### DIFF
--- a/dist/DayPicker.js
+++ b/dist/DayPicker.js
@@ -1,28 +1,30 @@
 'use strict';
 
-var _interopRequireWildcard = function (obj) { return obj && obj.__esModule ? obj : { 'default': obj }; };
-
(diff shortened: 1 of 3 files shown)


zzarcon:
@gpbl should I include the `dist` folder in the PR? thanks!

gpbl:
Hi @zzarcon thanks... question: shouldn't we also use `props.style` then?

(i'm not sure if I'm doing the right thing with that `dist` folder... I would anyway overwrite it before publishing to npm)

zzarcon:
@gpbl Well, if you see the https://github.com/js-next/react-style approach, definitely yes :). 
But in my opinion it's enough that a vendor component only admits the **className**, since you can do whatever you want from the stylesheet file :)

PR #30, opened 2015-07-23
Feature: caption click -> show this month

adambbecker:
While testing this component, I navigated pretty far away from the current month and noticed that clicking the month display on the header didn't bring me back to today's month.

This PR enables a click on the displayed month `.DayPicker-Caption` to return the picker to the current month (at the time of the click).

### DIFF
--- a/site/src/style/DayPicker.scss
+++ b/site/src/style/DayPicker.scss
@@ -51,6 +51,7 @@
     display: table-caption;
     height: 1.5rem;
     text-align: center;
+    cursor: pointer;
   }
 
   .DayPicker-Weekdays {
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -123,7 +123,7 @@ class DayPicker extends Component {
       <div
         className="DayPicker-Month"
         key={i}>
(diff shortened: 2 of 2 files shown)


gpbl:
Hi @adambbecker, thanks for the contribution! I wonder – is that an expected behavior? It feels weird that clicking on "August 2016" it displays the current month. Or do date pickers usually do that? 
Also, the click imho should display the `initialMonth` instead of the current month.

An alternative could be to add a `onCaptionClick` prop so that the dev chooses what to do when the user clicks the caption.

adambbecker:
Hey @gpbl, no problem!

> is that an expected behavior? It feels weird that clicking on "August 2016" it displays the current month. Or do date pickers usually do that? 

Haha, so apparently I thought it was, but it looks like it's not as expected as I thought. In that case I think your suggestion of:

> An alternative could be to add a onCaptionClick prop so that the dev chooses what to do when the user clicks the caption.

is a really good one and I've opened #31 (which you've already noticed :smile: ) to add that functionality. Going to close this one in favor of #31.

Issue #33, opened 2015-07-28
Add ARIA attributes for better accessibility


basham:
While `react-day-picker` works with the keyboard, its semantics could be improved by adding the appropriate ARIA attributes, so screen readers will better interpret how the component works.

Probably the best React exemplar I've found which achieves this nicely is [`react-widgets` Calendar](http://jquense.github.io/react-widgets/docs/#/calendar). However, I see the customizability of the `react-day-picker` project being much more ideal. I really like the clean, small ES6 implementation with no dependencies outside of React.

The following are some suggestions based on my research. Some suggestions are a result of adding back certain semantics which were lost due to [switching from `<table>` to `<div>`](https://github.com/gpbl/react-day-picker/issues/25). Suggestions are organized by the element to which it affects, identified by its class name.
## `.DayPicker-Month`
- Add [`role="grid"` attribute](http://www.w3.org/TR/wai-aria/roles#grid); grid cells need an ancestor `grid`.

> Authors MUST ensure that elements with role gridcell are owned by elements with role row, which in turn are owned by an element with role rowgroup, grid or treegrid.
- Add [`aria-readonly="true"` attribute](http://www.w3.org/TR/wai-aria/states_and_properties#aria-readonly) to indicate that the child grid cells will never be edited; by default, grid cells are assumed to be editable.

> A grid is considered editable unless otherwise specified. To make a grid read-only, set the aria-readonly attribute of the grid to true. The value of the grid element's aria-readonly attribute is implicitly propagated to all of its owned gridcell elements, and will be exposed through the accessibility API.
- Add [`aria-labelledby` attribute](http://www.w3.org/TR/wai-aria/states_and_properties#aria-labelledby) to explicitly indicate the `.DayPicker-Caption` element is labeling the month grid.
- Add [`aria-activedescendant` attribute](http://www.w3.org/TR/wai-aria/states_and_properties#aria-activedescendant) to explicitly indicate which `.DayPicker-Day` is in focus.
## `.DayPicker-Caption`
- Add unique id to work with `.DayPicker-Month` enhancements.
## `.DayPicker-Weekdays`
- Add [`role="rowgroup"` attribute](http://www.w3.org/TR/wai-aria/roles#rowgroup), which is equivalent to using `<thead>` element.
## `.DayPicker-Weekday`
- Add [`role="columnheader"` attribute](http://www.w3.org/TR/wai-aria/roles#columnheader), which is equivalent to using the `<th>` element.
[...]

Issue #35, opened 2015-07-29
New labels prop


gpbl:
A `labels` prop may help to implement `title` and `aria-label` attributes (see #33). Example:

```js
// default labels
let labels = { "today": "Today" }
<DayPicker labels={labels} />

// localized, modifier specific labels
let modifiers = { "weekend": day => day.getDate() === 5 || day.getDate() === 6 }
let labels = { "weekend": { en: "Weekend", it: "Fine settimana", de: "Wochenende" }
<DayPicker labels={labels} modifiers={modifiers} />
```

gpbl:
Not sure if call the prop `dayLabels` or just `labels`

basham:
I'm a little concerned about having this project be opinionated about the best way to implement this feature. I could see situations in which both `title` and `aria-label` could co-exist on the same calendar. But if providing both options, the API could get rather messy.

I wonder if what's really needed is a hook to add any additional attributes to a given day element. That way, the author could determine how `title`, `aria-label`, or anything else not currently managed by this project could be integrated. Yes, this could be accomplished by having a custom day component render with these attributes on child elements, but it may be nicer having the ability to attach it directly on the day element to reduce unnecessary DOM nesting.

basham:
In this way, localization concerns for this feature are offloaded to the implementation, instead of this project.

gpbl:
I agree with your point. There's a `renderDay` prop rendering the *content* of the day cell. Maybe it would work better rendering the whole cell?

```js
render() {
  function renderDay(day, modifiers) {
    return <div title="MyCustomLocalizedTitle">{ day.getDate() }</div>;
  }
  return <DayPicker renderDay={renderDay} />;
}
```

The component would still append the proper `className` to it.

basham:
I like where that's going. But in that solution, you'd be responsible for adding additional element properties (`key`, `className`, `tabIndex`, `role`, `onKeyDown`, `onMouseEnter`, `onMouseLeave`, `onClick`, `onTouchTap`) after the `renderDay` function has worked. Meaning that you're enhancing and potentially overriding implementation details. Perhaps it would be better for all the element properties to be wrapped in an object and sent as a prepared set of attributes that the implementation has to deal with. Yes, the implementation could mess things up, but it would also be within the control of the implementation to fix it.

In this way, the implementation would be responsible for transferring the generated properties on the new element. And it gives opportunity for the props to be modified as needed.

```js
function renderDay(day, props) {
  // Modify props, as needed.
[...]

PR #41, opened 2015-07-29
add: ARIA attributes to the nav buttons

joeybaker:
This is related to #33. I think this as more of a continuation of that discussion than anything else.

Basically: the nav buttons aren't ARIA compliant, and this solves that problem, but there's a better option: use real `<button>` elements instead of `<span>`s. This is more correct, but might be a breaking change. What do you think?

### DIFF
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -262,16 +262,23 @@ class DayPicker extends Component {
 
   renderNavBar() {
     const baseClass = "DayPicker-NavButton DayPicker-NavButton";
+    const { localeUtils, locale } = this.props;
     return (
       <div className="DayPicker-NavBar">
-        <span
+        <button
           key="prev"
           className={ `${baseClass}--prev` }
-          onClick={ ::this.handlePrevMonthClick } />
-        <span
+          onClick={ ::this.handlePrevMonthClick }
+          tabIndex="-1"
+          type="button"
+          aria-label={ localeUtils.getPreviousLabel(locale) } />
+        <button
           key="next"
           className={ `${baseClass}--next` }
-          onClick={ ::this.handleNextMonthClick } />
+          onClick={ ::this.handleNextMonthClick }
+          tabIndex="-1"
+          type="button"
+          aria-label={ localeUtils.getNextLabel(locale) } />
       </div>
     );
   }
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -7,6 +7,12 @@ const WEEKDAYS_SHORT = ["Su", "Mo", "Tu",
 const MONTHS = ["January", "February", "March", "April", "May", "June",
   "July", "August", "September", "October", "November", "December"];
 
+const LABELS = {
+  NEXT: "Next",
+  PREV: "Previous"
+};
+
+
 const Utils = {
 
   addMonths(d, months) {
@@ -127,6 +133,14 @@ const Utils = {
 
   getMonthsDiff(d1, d2) {
     return d2.getMonth() - d1.getMonth() + (12 * (d2.getFullYear() - d1.getFullYear()));
+  },
+
+  getNextLabel() {
+    return LABELS.NEXT;
+  },
+
+  getPreviousLabel() {
+    return LABELS.PREV;
   }
 
 };
--- a/test/Utils.js
+++ b/test/Utils.js
@@ -244,4 +244,15 @@ describe("Utils", () => {
     });
   });
 
+  describe("getNextLabel", () => {
+    it("returns Next", () => {
+      expect(Utils.getNextLabel()).to.equal("Next");
+    });
+  });
+
+  describe("getPreviousLabel", () => {
+    it("returns Previous", () => {
+      expect(Utils.getPreviousLabel()).to.equal("Previous");
+    });
+  });
 });


gpbl:
So the reason they are not buttons is that they are meant to be used with the mouse. With the focus on the calendar, use the arrow keys to move back and forward. When you tab from there you directly focus the day cell (if `onDayClick` or `onDayTouchTap` are enabled). That was my attempt, at least – you can try the intended behavior in the selectable day example. Do you think this break some accessibility rule?

basham:
`<button>` is the correct semantic. Rather than making it a `<span>` to avoid mouse interaction, just add `tabindex="-1"` attribute to it.

Also, perhaps if there's more than one month displayed, the value of `aria-label` is just `Previous` or `Next`, so you avoid the problems of misrepresenting or trying to over-accommodate whatever the month groups are.

```html
<div className="DayPicker-NavBar">
  <button
    tabindex="-1"
    aria-label="Previous"
    key="prev"
    className={ `${baseClass}--prev` }
    onClick={ ::this.handlePrevMonthClick }>
  </button>
  <button
    tabindex="-1"
    aria-label="Next"
    key="next"
    className={ `${baseClass}--next` }
    onClick={ ::this.handleNextMonthClick }>
  </button>
</div>
```

joeybaker:
What @basham has looks more correct to me, but it is a breaking change (people will have to add additional styling to override the `<button>` defaults). @gpbl if you're okay with that, I'm happy to make the change.

basham:
`<button>` should have `type="button"` attribute attached to it. Otherwise, it's assumed to be `type="submit"`, which is undesirable.

> `submit`: The button submits the form data to the server. This is the default if the attribute is not specified, or if the attribute is dynamically changed to an empty or invalid value.

> `button`: The button has no default behavior. It can have client-side scripts associated with the element's events, which are triggered when the events occur.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

Therefore, the updated code (with alphabetical attributes, since that's my particular code convention):

```html
<div className="DayPicker-NavBar">
  <button
    aria-label="Previous"
    className={ `${baseClass}--prev` }
    key="prev"
    onClick={ ::this.handlePrevMonthClick }
    tabindex="-1"
    type="button">
  </button>
  <button
    aria-label="Next"
    className={ `${baseClass}--next` }
    key="next"
    onClick={ ::this.handleNextMonthClick }
    tabindex="-1"
    type="button">
  </button>
</div>
```

gpbl:
Thanks people – yes if we are going to make breaking changes I'll cut a major version, so no problem.

@joeybaker We need to make changes to support localized `aria-label`s, and some are not so easy to implement:

- [Utils](https://github.com/gpbl/react-day-picker/blob/master/src/Utils.js) will need 2 new functions: `getNextMonthLabel` and `getPreviousMonthLabel`, and return the relative constants added to the top of the file. Those should be used in `DayPicker.js` to get the labels.
- For supporting localized strings, we should add the signature of those two functions to the [localeUtils](https://github.com/gpbl/react-day-picker/blob/master/src/DayPicker.js#L25-L30) prop

Now, the difficult one:

- There's the [LocaleUtils](https://github.com/gpbl/react-day-picker/blob/master/src/addons/LocaleUtils.js) addon, based on moment.js, that should also implement the functions – but it won't be able to localize the "Previous" and "Next" by using moment.js (I guess). Here there will be some design decisions to do: e.g. 
    - dev should be able to *Object.assign* to `LocaleUtils` their own custom `getNextMonthLabel`/`getPreviousMonthLabel` methods, and add the strings for the locales they want to support (need to open an issue about this, since it was already required)
    - include some locale data together with the package

I tried to avoid the `aria-label` attribute because of these problems. 😅 Maybe we can accept a compromise or a better solution.

joeybaker:
@gpbl what do you think of just adding props for `previousMonthLabel` and `nextMonthLabel`? That would push the problem of localization (which is kinda beyond scope of this element IMHO) out to the user.

basham:
@joeybaker There's similar offloading discussions happening for issue #35. Unsure if there's any strategy that could be shared there. But in general, putting more responsibility on the implementer and having a less complicated widget would be preferred to me.

joeybaker:
@basham cool. Looks like we're thinking along the same lines.

@gpbl I'm not sure how you want to handle the i18n problem, but I quickly did as you suggested and:

1. changed to `<button>`s
2. added methods to `Utils`

Pending the result of #35, this might be good to go.

basham:
Instead of `MONTH_LABELS`, I wonder if it's best to be more generic and call it `NAVIGATION_LABELS` or `NAVIGATION` or `LABELS`?

joeybaker:
Hmmm… good point. I like `LABELS`.

joeybaker:
Changed.

basham:
:+1:

src/Utils.js
```diff
@@ -7,6 +7,11 @@ const WEEKDAYS_SHORT = ["Su", "Mo", "Tu",
 const MONTHS = ["January", "February", "March", "April", "May", "June",
   "July", "August", "September", "October", "November", "December"];
 
+const LABELS = {
+  NEXT: "Next"
+  , PREV: "Previous"
```
gpbl: No comma-first please :-)

src/Utils.js
```diff
@@ -127,6 +132,14 @@ const Utils = {
 
   getMonthsDiff(d1, d2) {
     return d2.getMonth() - d1.getMonth() + (12 * (d2.getFullYear() - d1.getFullYear()));
+  },
+
+  getNextMonthLabel() {
+    return LABELS.NEXT;
+  },
+
+  getPreviousMonthLabel() {
```
gpbl: We could change the function names to `getPreviousLabel` and `getNextLabel` then...

src/DayPicker.js
```diff
@@ -104,14 +104,18 @@ class DayPicker extends Component {
...
       <div className="DayPicker-NavBar">
-        <span
+        <button
           key="prev"
           className={ `${baseClass}--prev` }
-          onClick={ ::this.handlePrevMonthClick } />
-        <span
+          onClick={ ::this.handlePrevMonthClick }
+          tabIndex="-1"
+          aria-label={ this.props.localeUtils.getPreviousMonthLabel() } />
```
gpbl: Sorry, I'd ask for another change. We need to pass the `locale` to the function for supporting localization as in the other `format*` functions:
 
```
const { localeUtils, locale } = this.props;
//snip 
aria-label={ localeUtils.getPreviousMonthLabel(locale) }
// snip
aria-label={ localeUtils.getNextMonthLabel(locale) }
```

thanks !

src/DayPicker.js
```diff
@@ -104,14 +104,18 @@ class DayPicker extends Component {
...
+        <button
```
basham: Add `type="button"`

src/DayPicker.js
```diff
@@ -104,14 +104,18 @@ class DayPicker extends Component {
     const baseClass = "DayPicker-NavButton DayPicker-NavButton";
     return (
       <div className="DayPicker-NavBar">
-        <span
+        <button
```
basham: Add `type="button"`

joeybaker:
@basham why add `type=button`? That's the default, no?

basham:
@joeybaker No, `type="submit"` is the default type. See my [comment above](#issuecomment-126415928).

joeybaker:
@gpbl changes made!

gpbl:
@joeybaker thanks! need to add a comma or the build will fail... sorry for stressing you :0

joeybaker:
My bad. Should be fixed now.

gpbl:
Thanks @joeybaker and @basham for the invaluable help! I need to organize a bit the release before merging.

ezufelt:
I recently started using React and found this component for use with my project. I am a blind developer, so am glad to see the interest in improving accessibility.

Any thoughts when this might get merged and a new version released? Until then I will need to use a forrked or patched version.

joeybaker:
Looks like this is no longer merge-able. @gpbl are you interested in me rebasing?

gpbl:
@joeybaker yes thanks! I think moving the `render` methods to the bottom of the file broke it 🐔 Could you please check if the CSS styles are still working after this change? I believe the change to `<button>` would require to update the CSS as well (run `cd site && npm install && npm start` and open localhost:3000)

gpbl:
@ezufelt there was some work to do with the release of node 4 and react 0.14. Now we can go on the [v2 release](https://github.com/gpbl/react-day-picker/milestones/v2.0.0) which will include this and other ARIA improvements.

ezufelt:
This sounds good.

My only comment on the PR from an accessibility perspective is wondering why tabindex = -1. I believe that it should be 0 (or not specified at all).

Although the calendar may be keyboard accessible when it receives focus, this doesn't work particularly well for a screen-reader user, who may be tabbing through a form including the date picker.

The challenge around the keyboard accessibility for screen-reader users is:

1. Discoverability - it is hard to determine the keys to use to navigate the calendar.
2. Interaction - most (all?) screen-readers will only allow navigation using keyboard commands like this if the "application" (role="application") mode is enabled. Typically this is reserved for true web applications and placed on the root element in the DOM, however, I have used it on widgets where the benefit of additional keyboard commands outweighed the drawbacks of modifying the default behaviour of the assistive technology. This is, in my opinion, a judgement call that should be made by the implementer of the component.

gpbl:
@ezufelt how would you expect the day picker work with the tab key? Should each press focus the next day?

ezufelt:
TL;DR

I don't think that each day should receive focus, likely only the selected day and any other non-day interactive elements.

I want to be really careful with my response, since it is how "I" expect it to work, and I am only one assistive technology user.

First, my general heuristic method when I use forms is not to tab through, but to read line by line down the page, or to jump from one form component to the next using the "f" key (I typically use the JAWS screen-reader on the web).

If I were to tab through a form I would likely not want to land on all 28 - 35 dates in the single month of a calendar (or up to 70 in 2 month view).

So I think that whether you tab or navigate the calendar by other means, there are few that would want to tab to each date. This still raises the challenge of finding / selecting a new date. Having appropriate table semantics (I see another issue about this) does help the navigation of the month view, since screen-readers typically give useful keystrokes for moving through a table / grid.

I would make all interactive elements (if they receive focus or not, either a button or anchor (or use role=button / role=link).

For keyboard only users without a screen-reader, or when the component is part of an ARIA application, the navigation by keystroke should work well. For those interacting with a screen-reader where the component is not part of an ARIA application, the ability to tab to a date, and to read by row, or navigate the grid, should give enough information to have a meaningful experience.

There are some other things I haven't covered above that are out of scope for this issue. All focusable elements should show an appropriate focus state as a visual affordance that they are focused. I think that the Month, when interactive (a callback has been passed through props, should also be a button / anchor.

joeybaker:
Okay, this is now rebased on master. However, I do agree with @ezufelt it's very strange to have controls and not make them "visible". I'd vote to just make them buttons.

davidgilbertson:
@ezufelt I'd be interested to hear your thoughts. We're using this date picker on our site, but when you tab into the 'date' field of our form, you first land in the input box where you can type day/year/month. The next tab stop is the button to show the calendar which I'm considering of less use to keyboard-only users.

Would that be clearer to you than navigating a calendar?

PR #42, opened 2015-07-29
Integrate react-a11y into tests

joeybaker:
Related to #33 This might be a good way to ensure future ARIA compatibility. It runs react-a11y within the tests. This is just a demo for discussion purposes. Thoughts?

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "babel": "^5.8.21",
     "babel-core": "^5.8.22",
     "babel-eslint": "^4.0.10",
+    "chai": "^3.3.0",
     "classlist-polyfill": "^1.0.1",
     "coveralls": "^2.11.2",
     "eslint": "^1.5.0",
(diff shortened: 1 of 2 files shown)


gpbl:
Very cool, thanks, I haven't thought to use `react-a11y` in the tests 😌 But why is the build passing? I was expecting some errors.

Is the hook on `console.log` required? I think `a11y(React, { throw: true });` should work out of the box.

joeybaker:
That's my fault! For some reason, what I had was succeeding on the first test run, but failing on subsequent runs with `test:watch`. I got it working and for got to re-check the first run. 

I've modified the code, and tests correctly fail now.

gpbl:
Tests have been updated to work with node 4 and react 0.14 (they are much faster now). Could you update the PR to work with master? E.g. we can now place `a11y(React, {throw: true})` at the top of the file.

joeybaker:
Rebased on master. I left the specific a11y test in place b/c otherwise _all_ the tests fail with the same error. I'm not sure that's useful output.

gpbl:
I'm closing this as ARIA has been added to eslint/airbnb config.

PR #57, opened 2015-09-01
Fix addMonths 31st day bug

EvanSimpson:
When adding months, if the `currentMonth` value has a day out of range of the target month, the date carries over into the next month. For example, if `currentMonth` is set to August 31st and `addMonths` is called with `months = 1`, the resultant date is October 1st, where the expected output month would be September, because September has no 31st day.

From the [Mozilla docs for `Date.prototype.setMonth()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setMonth) :
> If you do not specify the `dayValue` parameter, the value returned from the getDate() method is used.

So this issue can be avoided by ensuring that the day value argument in `setMonth` is always set to 1.

### DIFF
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -11,7 +11,7 @@ const Utils = {
 
   addMonths(d, months) {
     const newDate = this.clone(d);
-    newDate.setMonth(d.getMonth() + months);
+    newDate.setMonth(d.getMonth() + months, 1);
     return newDate;
   },
 


gpbl:
Thanks @EvanSimpson! I'm going first to add a test for this, do you have a case for reproduce the bug? 

It should not have happened because the `addMonths` function is always using `state.currentMonth`, and my assumption in the code is that a `month` is a date having its date `1`, hours, minutes and ms as `0`. So to fix the issue I'd write that function as:

```js
addMonths(d, months) {
    const newDate = this.startOfMonth(d);
    newDate.setMonth(d.getMonth() + months, 1);
    return newDate;
  }
```

What do you think? Or are you using that function somewhere else in your code?

gpbl:
Maybe I've got it: it seems the only case where i change `currentMonth` to a date not being the first day of the month is here, https://github.com/gpbl/react-day-picker/blob/master/src/DayPicker.js#L67 e.g. when updating the props to the component. That line should be written as

```js
this.setState({
    currentMonth: Utils.startOfMonth(nextProps.initialMonth)
});
```

Is this your case too?

EvanSimpson:
Yes, I believe that must be the case, because I only see the issue after updating `initialMonth` to be a newly selected date.

I think all of the proposed changes have the same effect, though if your intention is to make sure `currentMonth` only ever represents a month, then changing `componentWillReceiveProps` seems like the right way to go.

gpbl:
Alright than I replace this PR with #60. Thanks for reporting the issue!

gpbl:
Published as v1.0.5 🎉

EvanSimpson:
:confetti_ball: :calendar: :tada:
Awesome! Thanks for being so responsive.

PR #66, opened 2015-10-05
Added locale settings based on moment

msangui:


gpbl:
Hi @msangui thanks for your PR 👋 

Maybe the docs are not clear, but we support already a moment-based localization. See "Localization with moment.js" in the [tips page](http://www.gpbl.org/react-day-picker/#tips). 

I wish to keep this module independent from any i18n library, so if you want to use moment.js to translate the calendar content, you can use the [`LocaleUtils`](https://github.com/gpbl/react-day-picker/blob/master/src/addons/LocaleUtils.js) add-on.

msangui:
Thanks for clarifying! :D

PR #74, opened 2015-10-15
Let the event from next/previous month click to propagate.

kblcuk:
There seem to be no real need to stop event propagation in this case, and doing so might cause unwanted side-effects in certain cases.


gpbl:
Thanks!

gpbl:
Published as v1.0.10 !

PR #81, opened 2015-11-11
Update style.css to fix broken layout

JKillian:
This was overriding `.DayPicker-Weekday { display: table-cell }` and causing the layout to break


gpbl:
Ehy @JKillian thanks for the PR! We should however fix this problem restoring the old behavior as per #82.

JKillian:
:+1: I think your fix is better! Glad things got tweaked

PR #83, opened 2015-11-11
Restore the row for the weekdays header (fix #82)

### LINKED ISSUE #82: DayPicker-Weekdays should have a table-row as child
As pointed out by @nkbt, [this commit](https://github.com/gpbl/react-day-picker/commit/0164a38f651771c00d3b4949898937d2013c7ddd) broke some styles, so [the proposed solution](https://github.com/gpbl/react-day-picker/pull/81) by @JKillian is to remove `display: table-row` in  the CSS.

However it is preferred to have a `table-row` inside `DayPicker-Weekdays`, which is a `table-header-group`.

So we should revert that commit and restore the surrounding `div`. This `div` was however missing a className, thus the confusion that lead to this problem: I'd assign to this element the `DayPicker-Weekdays-row` class.

gpbl:
This PR address reverts [this commit](https://github.com/gpbl/react-day-picker/commit/0164a38f651771c00d3b4949898937d2013c7ddd) adding a `table-row` element around the `DayPicker-Weekday` table cells. #82

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -54,18 +54,19 @@
 
   .DayPicker-Weekdays {
     display: table-header-group;
-    > div {
-      display: table-row;
-    }
   }
 
-    .DayPicker-Weekday {
-      display: table-cell;
-      padding: .5rem;
-      font-size: .875em;
-      text-align: center;
-      color: #8b9898;
+    .DayPicker-WeekdaysRow {
+      display: table-row;
     }
+
+      .DayPicker-Weekday {
+        display: table-cell;
+        padding: .5rem;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -54,19 +54,20 @@
 
   .DayPicker-Weekdays {
     display: table-header-group;
-    > div {
```
nkbt: I see! CSS is OKish with nested selectors by silently dropping them =(. Loved that feature for years.

gpbl: lol... it's css, what could go wrong? :)

Issue #79, opened 2015-11-10
Bower Support


JKillian:
Although not ideal, I'm in a situation where installing the package via NPM doesn't meet my use case as we're not using browserify, webpack or anything like that. Is there a current usage story for using this via bower? If not, would you be interested in a contribution to make it more bower friendly?

gpbl:
Yes thanks! I have no idea how does bower work, but sure it would be nice to support it.

gpbl:
Woah I've made some crazy things with `bower register`, `npm publish` etc... 👯 now it should work everything as version 1.1.3. 

@JKillian Could you please test if it *actually* works with bower and eventually close the issue? I don't get why it doesn't appear in [their search](http://bower.io/search/?q=react-day-picker)...

JKillian:
@gpbl I see it in the search now - probably just a caching issue! I'll try it out now

JKillian:
Working like a charm in my project :+1:

JKillian:
Right now we're only exposing the DayPicker React Class as a global variable named `DayPicker`. Do you think instead we should export an object called `DayPicker` and then it could have the following fields:

* DayPicker
* LocalUtils
* MomentLocaleUtils
* Helpers
* DateUtils

or something similar to that?

gpbl:
I was wondering as well... we should however export only the "official" API. 

I don't know what are the common practices for a bower component, but I imagine:

```js
DayPicker
DayPicker.localeUtils
DayPicker.dateUtils
DayPicker.momentLocaleUtils
```

I think we can add some static properties to `DayPicker` when exporting it, so that the component is actually `<DayPicker />` and not `<DayPicker.DayPicker>` or, as I thought at first, `<DayPicker.Component />`. 

(I'm not so happy for this `*Utils`  convention , how would you call them?)

gpbl:
I have no interest in bower so I'd really appreciate a PR 🖖. (pay attention to that `momentLocaleUtils`, because it requires `moment` but I won't put moment as dependency. Maybe you need some "ignore" setting in the webpack config).

Deep thanks for your support, JKillian!

JKillian:
I think PR #93 should take care of things here. Let me know once you release and I'll test it out!

gpbl:
@JKillian I've published as 1.1.4 – it should work! Does it? 👀

JKillian:
Working well for me! :+1:

gpbl:
excellent! thanks for the help!

PR #84, opened 2015-11-11
Add support for publishing to bower

JKillian:
Here's a simple Bower build that'll expose a global `DayPicker` variable. Bower uses git tags to keep track of the current version, so you'll have to add a tag when you change the version in `package.json`.  (Here are the [bower docs](http://bower.io/docs/creating-packages/) for convenience.) Let me know if you have any questions!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+bower_components
 node_modules
 npm-debug.log
 /examples/built
--- a/README.md
+++ b/README.md
@@ -49,3 +49,21 @@ function MyComponent() {
   return <DayPicker initialMonth={ new Date(2016, 1) } modifiers={{ isSunday }} />
 }
 ```
+
+Or install via bower
+
+```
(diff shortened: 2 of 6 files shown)


JKillian:
Related to #79

gpbl:
Nice, thanks! But how does bower know we have a new bower-enabled package?

gpbl:
I've made some changes with #85

JKillian:
:+1: 

As far as bower, you just have to use [bower register](http://bower.io/docs/creating-packages/#register) once and add git tags to mark releases. (I'd assume you'd just do this whenever you update the version in package.json and get ready to push a new release to npm.)

JKillian:
By the way, really like this component! It's the best React one of its kind I could find out there

gpbl:
> By the way, really like this component! It's the best React one of its kind I could find out there

Thanks! Indeed I feel it deserves some more stars 😢

PR #93, opened 2015-11-18
Add utils modules as fields of bower export

JKillian:
Related to #79


JKillian:
Updated the PR to move things from `DayPickerBower` to `DayPicker`. Did you want `MomentLocaleUtils` included as well? I wasn't sure

gpbl:
I think we should keep it out, it is very simple and who is using bower, could just copy and paste it.

JKillian:
@gpbl :+1:  I think keeping it out is good as well

PR #96, opened 2015-11-20
Fix bug where endpoint months could not be shown

### LINKED ISSUE #95: Bug in fromMonth/toMonth and showMonth method
Say `toMonth` is set to December 2015 and someone tried to call `showMonth(new Date(2015,11))` in order to display December 2015.

`allowMonth` incorrectly returns false as `toMonth` and `d` are the same month. It should return true. I believe this should be a simple fix by changing `Helpers.getMonthsDiff(toMonth, d) >= 0` to `Helpers.getMonthsDiff(toMonth, d) > 0` (and doing the same for `fromMonth`).

JKillian:
Fixes #95

### DIFF
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -93,8 +93,8 @@ export default class DayPicker extends Component {
 
   allowMonth(d) {
     const { fromMonth, toMonth }  = this.props;
-    if ((fromMonth && Helpers.getMonthsDiff(fromMonth, d) <= 0) ||
-      (toMonth && Helpers.getMonthsDiff(toMonth, d) >= 0)) {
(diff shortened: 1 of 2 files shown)


gpbl:
Thanks, i've added a [test case](https://github.com/gpbl/react-day-picker/commit/250b642c70e2a31cd7293cd37149b1512ffd4ae6) for it!

JKillian:
Great! I think we may have added redundant test cases though - the tests I had in my PR are very similar to the tests you added. (No big deal, just wanted to let you know in case)

gpbl:
Oh sorry sorry! I didn't see it 🎣

PR #101, opened 2015-11-24
Fix incorrect import

megheaiulian:
When require('../src/DayPicker') returns a object. The class itself is on 'default' property.
It's the way ES6 modules work.

### DIFF
--- a/test/DayPicker.js
+++ b/test/DayPicker.js
@@ -14,7 +14,7 @@ ExecutionEnvironment.canUseDOM = true;
 
 const TestUtils = require("react-addons-test-utils");
 
-const DayPicker = require("../src/DayPicker");
+const DayPicker = require("../src/DayPicker").default;
 
 const keys = {
   LEFT: 37,


gpbl:
Good catch! Thanks. Yet this is *not* how ES6 modules [work](http://exploringjs.com/es6/ch_modules.html). I think it is a bug in babel 6.0.

megheaiulian:
Sorry, what i mean is that this is how require should work with ES6 modules.
`export default function(){}`  would be `export.default=function(){}`
You could switch `const DayPicker = require("../src/DayPicker").default;` to  a `import`.

gpbl:
👍 yes i've just investigated the thing a bit. 

With babel 6, commonJS behavior has changed: 
https://phabricator.babeljs.io/T2212
https://stackoverflow.com/questions/33505992

There's even [a plugin](https://www.npmjs.com/package/babel-plugin-add-module-exports) to restore the old behavior. 

The point is that I can't use `import` because tests in mocha wouldn't work. *Internally* in the tests I'm OK to use `.default`, but then we should also use it in [the main file](https://github.com/gpbl/react-day-picker/blob/master/DayPicker.js).

```js
module.exports = require("./lib/DayPicker").default;`
```

Thanks a lot for finding the culprit!

Issue #88, opened 2015-11-15
New captionElement prop


gpbl:
If specified, the value of this prop replaces the [caption element](https://github.com/gpbl/react-day-picker/blob/master/src/DayPicker.js#L325-L328) ([example](https://github.com/gpbl/react-day-picker/issues/52#issuecomment-156832760)).

adidahiya:
+1

PR #103, opened 2015-11-24
Add captionElement prop

gpbl:
This pull request adds a `captionElement` prop (#88) and a new `getMonths()` function to `LocaleUtils` that may be useful when using it.


src/DayPicker.js
```diff
@@ -2,6 +2,7 @@ import React, { Component, PropTypes } from "react";
 import Helpers from "./Helpers";
 import DateUtils from "./DateUtils";
 import LocaleUtils from "./LocaleUtils";
+// import Caption from "./Caption";
```
adidahiya: did you mean to leave this in commented like this?

gpbl: ah-ah good catch, thanks @adidahiya

PR #123, opened 2016-01-21
Spread attributes to container.

### LINKED ISSUE #122: Expose onBlur prop
`<DayPicker onBlur={} />`

I have a use-case for there where the DayPicker appears as a popover.

ATM, I show/hide the datepicker through a state (active when the input is focused; inactive on blur). However, clicking the DayPicker div changes the focus from the input to the `DayPicker`.

I have no means of checking if the user has focused away from the `DayPicker`.

- If there's blur, there's usually focus, but I'm not sure if there's any relevance.
- Or we could just use the spread operator so all other props are accepted (`<DayPicker onDayClick={...} {...other} />`).

I'm willing to send a PR, just need some approval first.

srph:
This would allow us to pass event listeners and all sort
of things. Closes #122.

Do not merge yet, needs documentation.

**edit**: oops, right before that &mdash; should we remove the explicit `style` props?


gpbl:
Thanks @srph, also for adding a test 🙌

Yes, I believe that `style`, `onKeyDown` and `tabIndex` should be treated as standard attributes. 

* `tabIndex` should have a default value as 0 (it is already like that), it would eventually set to `-1` when not needed
* `onKeyDown` requires special care as in wrote in [this comment](https://github.com/gpbl/react-day-picker/issues/122#issuecomment-173139313)

If you prefer, I can push these changes. For the docs, don't worry, I'll update them as the PR is merged.

srph:
I made some updates. I think I'm quite lost -- but I can revert stuff back to the original (before these new commits toda)).

If it looks good, I can squash the commits first, though.

gpbl:
Looks fine to me thanks! (Don't worry about the decreased coverage, i'll check it out)
Thinking twice I see the tabIndex will work properly as it is. It should be added only if months can be changed, but its value may came from an external prop.

gpbl:
Let me know if you need helps with the test... I know it's a bit too messy there, I should separate them in different files.

srph:
Hmm, anything missing?

By the way, how about the role props / attribute? Should we allow it to be overridable.

gpbl:
Nothing, it's perfect! 👍 Going to cut a new release soon 🚀

srph:
Alright, I hope I didn't break anything \*crosses fingers\*. Thanks for your work!

PR #128, opened 2016-02-05
Change documented return type of getFirstDayOfWeek to number

akirk:
In the documentation the return type of `getFirstDayOfWeek` is a `bool`. This should actually be a `number`. The [code that uses](https://github.com/gpbl/react-day-picker/blob/master/src/Helpers.js#L37) this value does a strict comparison with a number
```
if(week.length > 0 && day.getDay() === firstDayOfWeek) {
```

Also that value can indeed be a number different than 0 (Sunday) and 1 (Monday), for example in Hebrew it is commonly 6 (Saturday).


gpbl:
Thanks!

PR #131, opened 2016-02-19
Keyboard nav and aria improvements

limscoder:
#### What does this PR do?
keyboard nav:
 * navigate between weeks with up/down arrow keys when inner grid is focused
 * navigate between years with up/down arrow keys when out div is focused

aria:
 * added additional aria labels so screen reader reads date when focus changes (tested on OSX/chrome/voiceover and win8/IE11/JAWS)

#### Any background context you want to provide?
There are 5 commits. Started with some refactorring to simplify DayPicker, then added keyboard nav features and aria labels. Aria may still not be "perfect", but it's better and works with screen readers.

#### Where should the reviewer start?
DayPicker.js, test/DayPicker.js

#### How should this be manually tested?
Test keyboard nav in example app. Test with screen reader in example app (cmd-f5 to enable voiceover in OSX).

### DIFF
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -84,21 +84,13 @@ export default class DayPicker extends Component {
   }
 
   allowPreviousMonth() {
-    const { fromMonth } = this.props;
-    if (!fromMonth) {
-      return true;
-    }
-    const { currentMonth } = this.state;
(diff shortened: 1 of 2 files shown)


limscoder:
needs a little more review

PR #132, opened 2016-02-19
Keyboard nav and aria improvements

limscoder:
#### What does this PR do?
keyboard nav:
 * navigate between weeks with up/down arrow keys when inner grid is focused
 * navigate between years with up/down arrow keys when outer div is focused

aria:
 * added additional aria labels so screen reader reads date when focus changes (tested on OSX/chrome/voiceover and win8/IE11/JAWS)

#### Any background context you want to provide?
There are 5 commits. Started with some refactoring to simplify DayPicker, then added keyboard nav features and aria labels to date cells, so that it works better with screen readers.

#### Where should the reviewer start?
Look at each commit independently.

#### How should this be manually tested?
Test keyboard nav in example app.
Test with screen reader in example app (cmd-f5 to enable voiceover in OSX).

### DIFF
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -84,21 +84,13 @@ export default class DayPicker extends Component {
   }
 
   allowPreviousMonth() {
-    const { fromMonth } = this.props;
-    if (!fromMonth) {
-      return true;
-    }
-    const { currentMonth } = this.state;
-    return Helpers.getMonthsDiff(currentMonth, fromMonth) < 0;
+    const previousMonth = DateUtils.addMonths(this.state.currentMonth, -1)
+    return this.allowMonth(previousMonth);
   }
 
   allowNextMonth() {
-    const { toMonth, numberOfMonths } = this.props;
-    if (!toMonth) {
-      return true;
-    }
-    const { currentMonth } = this.state;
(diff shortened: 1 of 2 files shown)


gpbl:
Hi @limscoder thanks so much for this PR! Really appreciated. I'm on vacation now, I can give a look to it in the next days. 🌴

limscoder:
Any progress on reviewing this? Can I answer any questions?

src/DayPicker.js
```diff
@@ -429,10 +468,17 @@ export default class DayPicker extends Component {
         tabIndex = this.props.tabIndex;
       }
     }
+
+    const ariaLabel = this.props.localeUtils.formatDate ?
+      this.props.localeUtils.formatDate(day) : day.toDateString();
```
gpbl: The `formatDate` here is a new function to add to the [LocaleUtils](http://www.gpbl.org/react-day-picker/docs/LocaleUtils.html) right? I can take care of this (need to update docs, tests, etc. :smile:) 
We should add the `locale` props as second argument to the function, so implementers can use their own localization utils.

limscoder: yep, new function.

gpbl:
Thanks again @limscoder, those are great additions. They are going in the next major release, I'm working on it this week.

limscoder:
thx

PR #133, opened 2016-02-24
Add support to set minimum number of weeks returned by getWeeksArray()

rbrander:
- added numberOfWeeks prop to DayPicker
- added minWeeks to Helpers.getWeekArray()
- added unit test for minWeeks
- added documentation for new numberOfWeeks prop

@gpbl Please review

### DIFF
--- a/docs/APIProps.md
+++ b/docs/APIProps.md
@@ -41,6 +41,12 @@ The number of months to render, where `initialMonth` is the first month. Default
 
 ---
 
+#### numberOfWeeks (Number)
+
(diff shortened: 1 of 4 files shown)


gpbl:
Thanks @rbrander for the PR! I guess its reason is to avoid the calendar height to change when browsing between months. 

An easier solution is to set the container's height big enough to host 5 or 6 weeks. Having a prop setting an arbitrary number of weeks seems an overkill to me, and introduces potential bugs as it must consider different use cases.

PR #141, opened 2016-03-10
Add support for react 15.0.0-rc.1

GertjanReynaert:


gpbl:
Thanks Gertjan!

gpbl:
There are some problems with the touch-tap event plugin: https://github.com/zilverline/react-tap-event-plugin/issues/64, which points to https://github.com/facebook/react/issues/6221 :/

gpbl:
Released in v2

PR #144, opened 2016-03-18
handle the situation when no day is selected

NXTaar:
I made a little addition to documentation, about `modifiers` section
In your `selected` function when no day is selected (for example when datepicker is rendered on the page, and the user hasn't chose some date yet) the `selectedDay` is `null`, and this code will throw an error `Uncaught TypeError: Cannot read property 'getDate' of null`.

Users who will copy this chunk of code will get this error, so i added a little check for the preventing of such situation.


gpbl:
Thanks !

PR #146, opened 2016-03-22
Add default option to onDayKeyDown switch statement

jgkingston:
Thanks for this pull request!

Please explain the intent of your changes. If they involve a lot of code, you may want to discuss this first on our gitter room: https://gitter.im/gpbl/react-day-picker.

Add support for additional, custom keyDown events. For example passing a function to close the day picker if the user presses tab while focussed on a day.

### DIFF
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -302,9 +302,8 @@ export default class DayPicker extends Component {
       }
       break;
     default:
-        if (onDayKeyDown) {
-          onDayKeyDown(e);
-        }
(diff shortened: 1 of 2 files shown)


jgkingston:
My specific need for this is it make an accessible date input that a user can navigate into, through and away from using the keyboard and that also only shows the calendar when the associated text input has focus.

gpbl:
Hey @jgkingston, `onDayKeyDown` has been added to v2. As the version got some refactoring I won't merge this PR. Thanks anyway!

PR #147, opened 2016-03-30
Add reverseMonths property to render months in the reverse orde…

sonrtomas:
Desire Functionality: Display months in reverse order using the reverseMonths property

This pull request is intended to display the months on a reverse order when we are displaying multiple months in a column.

I'm attaching images as examples, (I'm displaying a calendar column from January of 2008 until today):

reverseMonths false default

reverseMonths true

On the first image, reverseMonths property is set to false (default) Jan 2008 appears at the top of the column. 

On the second image, reverseMonths property is set to true Mar 2016 appears at the top of the column.

### DIFF
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -43,7 +43,7 @@ export default class DayPicker extends Component {
 
     enableOutsideDays: PropTypes.bool,
     canChangeMonth: PropTypes.bool,
-    reverseMonthsRender: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


gpbl:
Very interesting thanks! Do you mind to change the prop name to  `reverseMonths`? Thanks!

sonrtomas:
Renamed prop name to reverseMonths

gpbl:
Thanks!

PR #157, opened 2016-04-15
Add version number to DayPicker class

JKillian:
This can be helpful if you need to figure out what version of a library is actually running in your app.

The downside is that as a dev, you have to remember to update this for each release, but I think it's worth it!


gpbl:
This is indeed a good idea! Thanks 👍

Issue #89, opened 2015-11-15
New dayElement prop


gpbl:
This prop should replace the current `renderDay` prop (that should become deprecated) and accept a React Element that will receive the `day` prop, e.g.

```js
function MyCustomDay({ day }) {
  return <div>{ day.toString() }</div>;
}

function MyComponent() {
  return <DayPicker dayElement={ <MyCustomDay /> } />
}
```

(this will supersede #50)

gpbl:
Closing – no need for this.

olegakbarov:
Actually i was waiting for this feature 🙄

gpbl:
@olegakbarov What is your use-case @olegakbarov?

olegakbarov:
I just need a custom element instead of `renderDay`.

gpbl:
@olegakbarov what you need the custom element for, if I may ask?

gpbl:
What about a `dayClass` prop? https://github.com/gpbl/react-day-picker/pull/172#issuecomment-222692854

lo1tuma:
I like this proposal. Currently we are using `renderDay` which depends on some state. In order to re-render, when the state has been changed, we need to create a new function which is bound to the new state and pass it to the `DayPicker` component. In order to avoid creating a new function each time when we render `DayPicker` it would be very helpful to define the day component class and additional props.

For Example:

```js
class Day extends React.Component {
    render() {
        const isSelectable = this.props.date > this.props.selectableRangeStart && this.props.date < this.props.selectableRangeEnd;
        return <div className={ isSelectable ? 'selectable' : ''}>{this.props.date.getDate()}</div>;
    }
}
```

```js
class DateRangePicker extends React.Component {
    render() {
         return <DayPicker dayClass={Day} dayProps={{ selectableRangeStart: this.state.selectableRangeStart, selectableRangeEnd: this.state.selectableRangeEnd }} />;
    }
}
```

gpbl:
With the question raised by @boatkorachal and solved with https://github.com/gpbl/react-day-picker/pull/179 my conclusion is that we should go with `dayElement`. So, @lo1tuma your case could be:

```jsx
<DayPicker 
    dayElement={ day => 
        <Day selectableRangeStart={this.state.selectableRangeStart} selectableRangeEnd={this.state.selectableRangeEnd} /> 
    } 
/>
```

Yet not sure about the signature of the dayElement function yet :) 

What do you think?

lo1tuma:
[...]

PR #159, opened 2016-04-18
Added support for all kinds of mouse, touch and key events on days an…

johannesd:
This adds all kinds of events to days and months that are supported by react as described here https://facebook.github.io/react/docs/events.html. It also supports handlers for the capture phase. 

Usage: In order to add a handler for mouseDown on days, use the prop "onDayMouseDown" on the component. For months, it would be "onCaptionMouseDown". For the capture phase, use "onDayMouseDownCapture", which is standard in react.

By calling e.preventDefaults() in a handler, you can stop react-day-picker to execute its own actions. E.g. in order to have an own action when pressing the cursor keys on a day and preventing the component from jumping to another day, just call e.preventDefaults() in your "onKeyDown" handler.

In order to keep the code short, I reduced the event handling to one single function "handleEvent" (instead of "handleDayKeyDown", "handleDayMouseEnter", etc.) that gets the event name as first parameter. Then, I iterate the props to find the ones starting with "onDay*", and register the corresponding handlers "on*" on the day element. The same happens with "onCaption*".

Only the type definitions add up quite a bit.

I don't have time to update the documentation at the moment...

### DIFF
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -45,7 +45,9 @@ export default class DayPicker extends Component {
 
     renderDay: PropTypes.func,
     weekdayComponent: PropTypes.func,
+    weekdayElement: PropTypes.element,
     navbarComponent: PropTypes.func,
+    navbarElement: PropTypes.element,
 
     captionElement: PropTypes.element,
 
@@ -330,7 +332,33 @@ export default class DayPicker extends Component {
       this.showPreviousMonth();
     }
   }
+  renderNavbar() {
+    const {
+      locale,
+      localeUtils,
+      canChangeMonth,
+      navbarComponent,
+      navbarElement,
+    ...attributes } = this.props;
 
+    if (!canChangeMonth) return null;
+    const props = {
+      className: 'DayPicker-NavBar',
+      nextMonth: this.getNextNavigableMonth(),
+      previousMonth: this.getPreviousNavigableMonth(),
+      showPreviousButton: this.allowPreviousMonth(),
+      showNextButton: this.allowNextMonth(),
+      onNextClick: this.showNextMonth,
+      onPreviousClick: this.showPreviousMonth,
+      dir: attributes.dir,
+      locale,
+      localeUtils,
+    };
+    if (navbarElement) {
+      return React.cloneElement(navbarElement, props);
+    }
+    return React.createElement(navbarComponent, props);
+  }
   renderDayInMonth(day, month) {
     let dayModifiers = [];
     if (DateUtils.isSameDay(day, new Date())) {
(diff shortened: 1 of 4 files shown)


gpbl:
Very nice thank you! Maybe we can avoid putting all those prop types, by generating them using an external function. Let see if I can find a better way :)

gpbl:
Also, we need to lint the file or the build won't run the tests :P 

I'm now wondering if this kind of abstraction for managing events is actually a good idea. 
Maybe rendering [a custom day element](https://github.com/gpbl/react-day-picker/issues/89) could help? We could attach a `data-` attribute to the elements containing day and modifiers... uhm...

johannesd:
Uff, sorry, I'll take a look at the failed checks tomorrow.

I don't know how event handlers on a custom day element would work along with react-day-picker's own handlers (i.e. key down)...

gpbl:
Don't worry, it is just the linter :) 
My main concerns are to keep the test coverage at 100% and to not complicate the `DayPicker.js` too much. It's already in need of some refactoring and I think this may be a good moment :) There's an upcoming major release so we don't have to worry *too much* about breaking changes. 

An option could be to implement an high-order component, which can pass down to the custom elements the current day and the modifiers. (this would mean that the `renderDay()` function should become a custom element itself)

gpbl:
@johannesd I definitely would like to improve the way we are passing down any event to captions and day cells, but we should find a cleaner way. 

From v2 we use a new `<Day />` component that may help to implement the right solution. 

I was thinking to a wrapper component that could help to abstract the changes you proposed with this PR:

An `EventWrapper` component may parse the props passed to the DayPicker and pass the event handler to the wrapped component:
```js
<EventWrapper prefix="thing" handlers={ this.props } />
   <Component />
</EventWrapper>
```

The `handlers` prop would parse `this.props` to find any prop beginning with `onThing*`. It would pass them down to `Component` as react events:

```js
const handlers = {
  onDayClick: () => { },
  onDayTouchStart: () => { },
}
<EventWrapper prefix="day" handlers={ handlers } />
   <Day />
</EventWrapper>
```

`Day` would receive `onTouchStart` and `onClick`.

The next step is making the wrapped component to pass custom arguments to the event handler, e.g. in the Day component they must receive the SyntheticEvent, the day and the day's modifiers. This is the code managing the events now:

```
function handleEvent(handler, day, modifiers) {
  if (!handler) {
    return undefined;
  }
  const dayModifiers = {};
  modifiers.forEach(modifier => { dayModifiers[modifier] = true; });
  return e => {
    e.persist();
    handler(e, day, dayModifiers);
  };
}

// snip
render() {
  const eventHandlers = {
    onClick: handleEvent(props.onClick, props.day, props.modifiers),
    onKeyDown: handleEvent(props.onKeyDown, props.day, props.modifiers),
    onMouseEnter: handleEvent(props.onMouseEnter, props.day, props.modifiers),
    onMouseLeave: handleEvent(props.onMouseLeave, props.day, props.modifiers),
    onTouchEnd: handleEvent(props.onTouchEnd, props.day, props.modifiers),
    onTouchStart: handleEvent(props.onTouchStart, props.day, props.modifiers),
  }
  return <div {...eventHandlers}</div>
}
```

we could abstract it as you made in this PR in an Helper function.

PR #172, opened 2016-05-31
renderWeekDay property

stanislav-ermakov-roi:
Adds new renderWeekDay property which should contain a function that renders weekday headers


gpbl:
Thanks @stanislav-ermakov-roi, nice idea! 👍 

I wonder however if it wouldn't be better to pass a prop with a `ReactComponent` class, instead of a function. Such component class would have the following props signature:

```jsx
<Weekday weekday={ number } localeUtils={ object } locale={ string } />
```

then, when implementing the day picker, the developer would pass it via the new `weekdayClass` prop:

```jsx
import DayPicker from 'react-day-picker';
import Weekday from './Weekday';

<DayPicker weekdayClass={ Weekday } />
```

In the DayPicker source code, we should write instead:

```diff
<div key={i} className="DayPicker-Weekday">
-   {renderWeekDay(i, locale, localeUtils)}
+   {React.createElement(this.props.weekdayClass, { weekday: i, locale, localeUtils })}
</div>
```

I understand you wanted to keep the similarity to `renderDay` prop :) However I was thinking to update that prop as well, e.g. passing a class: `dayClass={ Day }`. A similar idea was tracked in #89, using an element instead of a class.

What do you think?

stanislav-ermakov-roi:
@gpbl, I think your idea is good and is inline with architecture of 2nd version of component. I also need similar functionality for Navbar, since I need custom labels for next and previous buttons. So another pull request is coming very soon. I can update this one with approach you suggested, since I would like this to be merged very soon, if it's possible.

Regarding dayClass. Actually I would name this dayComponent, since class can be confused with className. Suggested idea could also be useful, but what if I just want to change cell content and not write all container props again? May be it's possible to update Day class to be able to extend it easily without the need to rewrite all the logic it contains already.

gpbl:
@stanislav-ermakov-roi awesome! Then, better discuss the implementation details first, as I feel bad asking to review a PR :) Would you mind to join the [gitter room](https://gitter.im/gpbl/react-day-picker)?

gpbl:
>  Actually I would name this dayComponent, since class can be confused with className. 

I see, yes, it should work... must be clear we are passing a component class (`Day`), not an element (`<Day />`). 

> but what if I just want to change cell content and not write all container props again

IMHO the component should not replace the container (e.g. that with `DayPicker-day` class), just the content. So is just about eventually renaming the prop. 

I think as in my diff above, even `weekdayComponent` should just used for the content, The container with class name `DayPicker-Weekday` should stay there.

stanislav-ermakov-roi:
Ok. I see your point. Going to rework this pull request.

gpbl:
@stanislav-ermakov-roi nevermind, while chatting comes out it's better the opposite about the container thing :)

gpbl:
Awesome thanks! Let me see why build fails then I'll merge :)

gpbl:
The test was just bad written, my bad. Fixed!

PR #175, opened 2016-06-06
IE11 for of loop fix #174

### LINKED ISSUE #174: IE11 symbol is undefined error
I updated to 2.1.0 today and stumbled upon a `"Symbol" is undefined` error in IE 11.
I think the problem lies in the new `for of`loop in [Helpers.js](https://github.com/gpbl/react-day-picker/blob/6adf3dc6a937c98a600626d4ca58a92b27e5a446/src/Helpers.js#L26).
Babel transpiles that into a Symbol iterator which doesn't exist in IE.

davidspiess:
As discussed in ticket #174 this pull request replaces the for of loop with a a reduce function, as babel compiles it to a Symbol, which has not support in IE.

### DIFF
--- a/src/Helpers.js
+++ b/src/Helpers.js
@@ -20,17 +20,14 @@ export function getDaysInMonth(d) {
   return resultDate.getDate();
 }
 
-export function getModifiersForDay(d, modifierFunctions) {
-  const modifiers = [];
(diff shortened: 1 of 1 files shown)


gpbl:
Thanks!

PR #176, opened 2016-06-06
Added fixedWeeks flag

fcsonline:
#### Problem

If your calendar is in a container with other elements around and you move between months, the container resizes to the number of weeks shown and this can create an annoying effect.

#### Solution

**Added fixedWeeks flag**

With this new flag, the calendar will show always the same number of
weeks. Exactly 6, because is the maximum number of week that a month can
have.

You can see here an example with this flag enabled:

6weeks

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -11,6 +11,7 @@
 | [captionElement](APIProps.md#captionelement) | | `Element` |
 | [disabledDays](APIProps.md#disableddays) | | `Bool` |
 | [enableOutsideDays](APIProps.md#enableoutsidedays) | false | `Bool` |
+| [enableFixedAmountOfWeeks](APIProps.md#enableFixedAmountOfWeeks) | false | `Bool` |
 | [fromMonth](APIProps.md#frommonth) | | `Date` |
 | [initialMonth](APIProps.md#initialmonth) | Current month | `Date` |
 | [locale](APIProps.md#locale) | en | `String` |
--- a/docs/APIProps.md
+++ b/docs/APIProps.md
@@ -47,6 +47,12 @@ Display the days outside the current month. Default is `false`.
 
 ---
 
+#### enableFixedAmountOfWeeks `Bool`
+
+Display a fixed number of weeks. All months have a number of weeks between 4 and 6. With this flag, the calendar will show exactly 6 weeks (Examples for 4 and 6 weeks: Febrary 2015, July 2016). Default is `false`.
+
+---
+
 #### canChangeMonth `Bool`
 
 Enable the navigation between months. Default is `true`.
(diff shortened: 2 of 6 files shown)


gpbl:
👋  @fcsonline 

I guess this is a dup of https://github.com/gpbl/react-day-picker/pull/133 (see [comment](https://github.com/gpbl/react-day-picker/pull/133#issuecomment-190910506). Maybe I don't get it: what is the problem with setting the container's min-height?

fcsonline:
The problem is that for months with 4 or 5 weeks you'll see a weird empty space. With this you can ensure that the calendar will show data for them.

gpbl:
However, for months having just four weeks, you would see two weeks more from the next month, which is also weird. I think such graphic glitches should not be handled by this component. How are other implementation handling this issue?

fcsonline:
This flag is disabled by default. I mean, you will see those extra weeks only when you have this flag enabled. I won't be weird because is what people will expect. The problem is that those "graphic glitches" can only be handled by this component. An external wrapper can not do any trick to avoid blank spaces.

Was https://github.com/gpbl/react-day-picker/pull/133 trying to solve the same problem?

gpbl:
Thanks for replying! Yes, in the OS X date picker I see this behavior implemented as well – so I changed my mind :) I never payed attention to this.

I'd call this prop `fixedWeeks` instead, or something else shorter than `enableFixedAmountOfWeeks`. (yes there's a `enableOutsideDays`, but that was a bad choice, may become deprecated soon). What do you think?

fcsonline:
Great! I'm going to change the flag name and fix those eslint errors. :clap:

gpbl:
Awesome thanks

fcsonline:
👏  Thanks!

fcsonline:
Could you release it, please?

gpbl:
I want first to update the docs and the examples :) While trying this prop, I wonder if setting `fixedWeeks` should automatically set `enableOutsideDays` to `true`. It doesn't make much sense to enable 6 weeks without displaying the outside days, does it?

fcsonline:
I was thinking about that during development. Maybe you want to avoid the graphic glitches but don't want to see outside days. Weird case indeed. I don't have a strong opinion about it.

gpbl:
@fcsonline So my position here is to enable them. Outside days can always be disabled by setting the prop to `false`.

fcsonline:
So you mean to enable it if is `undefined`, right? 👌

gpbl:
Uhm nope the default value for the prop is `false` :( We would mess the code just for an edge case that doesn't make sense anyway :) Release coming soon!

fcsonline:
Cool!

gpbl:
Published as v2.2.0! [example](http://www.gpbl.org/react-day-picker/examples/?fixedWeeks)

fcsonline:
Thanks!

PR #177, opened 2016-06-07
Add navbarElement prop

boatkorachal:
The use case for me is when I want to pass extra props to the navbar. I am not sure there is another way i don't figure out.
Thanks

### DIFF
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -45,6 +45,7 @@ export default class DayPicker extends Component {
     renderDay: PropTypes.func,
     weekdayComponent: PropTypes.func,
     navbarComponent: PropTypes.func,
+    navbarElement: PropTypes.element,
 
     captionElement: PropTypes.element,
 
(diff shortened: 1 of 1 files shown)


gpbl:
Hi @boatkorachal,

there's already the `navbarComponent` prop: http://www.gpbl.org/react-day-picker/examples/?customComponents

Thanks anyway!

boatkorachal:
Hi @gpbl,

my use case is I want to pass all birthdays, which is store outside the DayPicker component, to render the sum of unseen birthdays at the top of previous and next arrow. But I have no idea how to do it with navbarComponent prop.

Thanks!

gpbl:
I see :) I think you have here two options:

- absolute positioning of the elements containing the sum, put in the day picker container. A bit hacky but sure the easiest way
- i think you can use `navbarComponent` as well (this requires you to rewrite the whole navbar, check the source): you pass a react component class which get the birthdays as prop, as you do with the DayPicker implementation

boatkorachal:
I'd like to go with the option 2 but I still don't understand how to do it. Could you show me some example? 
What I understand is if I fix some code to pass all DayPicker props to navbarComponent then it is possible. 
Thanks again for your time :)

gpbl:
Yeah I see your problem @boatkorachal :) Let me think about it.

gpbl:
So you are right, it's not easily doable :( 

I think we should have added a `navbarElement` instead of a `navbarComponent` prop, so to use a React element (not a class) to render the navbar. Basically, a `navbarElement` prop would work the same as the `navbarComponent`, rendered with `cloneElement` instead of `createElement`. The API could look like:

```js
<DayPicker navbarElement={ ({...navbarProps}) => <MyNavBar {.. navbarProps} birthdaysCount={200} /> } />
```

To maintain consistency, we should do the same for `weekdayComponent` vs. `weekdayElement`. @boatkorachal would you mind to update your PR? (actually is more a rewrite :)

boatkorachal:
okie! I sent a new PR #179  (i don't know how to change the target PR branch). and sorry if i did anything wrong this is the first time i send PR haha ><

gpbl:
Closing as #179 has been merged already.

alexkuz:
@gpbl hi, I just found out that components are deprecated and I'm confused :) Why is that, again?
In your example, `({...navbarProps}) => <MyNavBar {...navbarProps} birthdaysCount={200} />` is a component, not an element, so I really don't see why elements are preferable.

gpbl:
@alexkuz yes instead of components we want elements, so it is easier to pass down props to them and we don't need to create a function for them.

> `({...navbarProps}) => <MyNavBar {...navbarProps} birthdaysCount={200} />` is a component, not an element

That function is actually returning an element (not a component), anyway there I was just thinking about a possible API. The final way to pass elements is:

```js
navbarElement={ <MyNavBar birthdaysCount={200} /> }
```

alexkuz:
> That function is actually returning an element (not a component)

No, I mean the function itself is a (stateless) component.

It's just I had the same issue in my component and decided to use components, so I wondered what was your motivation. Thanks for explanation!

PR #179, opened 2016-06-09
navbarElement and weekdayElement instead of component

boatkorachal:
from #177

### DIFF
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -45,6 +45,7 @@ export default class DayPicker extends Component {
 
     renderDay: PropTypes.func,
     weekdayComponent: PropTypes.func,
+    weekdayElement: PropTypes.element,
     navbarComponent: PropTypes.func,
     navbarElement: PropTypes.element,
 
@@ -336,12 +337,11 @@ export default class DayPicker extends Component {
       locale,
       localeUtils,
       canChangeMonth,
-      onDayClick,
       navbarComponent,
       navbarElement,
     ...attributes } = this.props;
 
-    if (!canChangeMonth) return null
+    if (!canChangeMonth) return null;
     const props = {
       className: 'DayPicker-NavBar',
       nextMonth: this.getNextNavigableMonth(),
@@ -353,11 +353,11 @@ export default class DayPicker extends Component {
       dir: attributes.dir,
       locale,
       localeUtils,
-    }
+    };
     if (navbarElement) {
-      return React.cloneElement(navbarElement, props)
+      return React.cloneElement(navbarElement, props);
     }
-    return React.createElement(navbarComponent, props)
+    return React.createElement(navbarComponent, props);
   }
   renderDayInMonth(day, month) {
     let dayModifiers = [];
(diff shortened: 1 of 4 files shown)


gpbl:
Hi @boatkorachal thanks for the help and congrats for your first PR 🎊 

To maintain backward compatibility, we _should not_ remove these `*Component` props. My idea was  actually to _add_ two new props. We could eventually deprecate them. Sorry for not being clear enough: I'll add some comments to your PR.

gpbl:
(You should rebase your branch with master to fix these conflicts)

src/DayPicker.js
```diff
@@ -43,8 +43,8 @@ export default class DayPicker extends Component {
     onCaptionClick: PropTypes.func,
 
     renderDay: PropTypes.func,
-    weekdayComponent: PropTypes.func,
-    navbarComponent: PropTypes.func,
```
gpbl: Keep those lat two lines. I'll add a PropType to deprecate them.

src/DayPicker.js
```diff
@@ -63,8 +63,8 @@ export default class DayPicker extends Component {
     canChangeMonth: true,
     reverseMonths: false,
     renderDay: day => day.getDate(),
-    weekdayComponent: Weekday,
-    navbarComponent: Navbar,
```
gpbl: idem

src/DayPicker.js
```diff
@@ -328,7 +328,29 @@ export default class DayPicker extends Component {
...
+      previousMonth: this.getPreviousNavigableMonth(),
+      showPreviousButton: this.allowPreviousMonth(),
+      showNextButton: this.allowNextMonth(),
+      onNextClick: this.showNextMonth,
+      onPreviousClick: this.showPreviousMonth,
+      dir: attributes.dir,
+      locale,
+      localeUtils,
+    };
+    return React.cloneElement(navbarElement, props);
```
gpbl: Here we should write instead

```
if (navbarComponent) {
  return React.createElement(navbarComponent, props);
}
return React.cloneElement(navbarElement, props);
```

boatkorachal:
ewww i just did the rebase and force push. hope that i haven't done anything wrong :)

gpbl:
Yay thanks! Could you even test that the rendered component is actually the navbarElement? 🙏 
https://coveralls.io/builds/6541993/source?filename=src%2FDayPicker.js#L357

boatkorachal:
Isn't it checked at proptypes? or should I import react-addons-test-utils and change to something like this

``` javascript
if ( TestUtils.isElement(navbarElement) ) {
  return React.cloneElement(navbarElement, props)
}
```

gpbl:
The test coverage has been decreased, this is because we didn't include a test to check if passing a `navbarElement` did actually render the element. See https://coveralls.io/builds/6541993/source?filename=src%2FDayPicker.js#L357 

Actually, even `weekdayElement` hasn't be checked: I wonder we are actually rendering it :) 
I can't work on this now, so that's why I'm asking :)

boatkorachal:
I have just update tests. Is it ok? (I have to admit that this is the first time i write the tests too haha)

boatkorachal:
oh sorry i forgot to run lint. i'm gonna fix it now.

gpbl:
Looks good! Thanks a lot!

PR #197, opened 2016-07-19
Add support for paged navigation. (fixes #196)

### LINKED ISSUE #196: Add support for navigating with `numberOfMonths`
I'm using this component with `numberOfMonths = 4` and I need the navigation between months to be more like changing the page of months. E.g:

The visible months can be [1, 2, 3, 4] or [5, 6, 7, 8] or [9, 10, 11, 12]
So the visible months can never be [3, 4, 5, 6] for example.

I'm not setting the `fromMonth` and `toMonth` props so I can never run into the case where some of the months are not allowed.
But I understand that it can problematic.

One solution would be to only render the allowed months. E.g:
If the `fromMonth = 2` and `toMonth= 3` then we render [2, 3]
If the `fromMonth = 2` and `toMonth= 5` then we render [2, 3, 4], [5]

An other solution would be to say that the only change to the current behavior is that `showNextMonth()` and `showPreviousMonth()` will change the `currentMonth` by `numberOfMonths` instead of `1`.

I would be happy to create a PR

zaygraveyard:


gpbl:
Thanks!

PR #205, opened 2016-09-05
Add isFutureDay DateUtils function

McFreely:
While using the picker, I felt the need to exclude days in the future, instead of days in the past. There is nothing ground-breaking here.

I've also added tests and documentation.

### DIFF
--- a/docs/DateUtils.md
+++ b/docs/DateUtils.md
@@ -25,6 +25,10 @@ Return `true` if the two dates `d1` and `d2` are the same day, e.g. ignoring the
 
 Returns `true` if `d` is in the past, e.g. is yesterday or any day before yesterday.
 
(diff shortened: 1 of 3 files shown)


gpbl:
Thanks! Would you mind to fix the lint issues so the Travis build can pass?
https://travis-ci.org/gpbl/react-day-picker/builds/157622516 🙏🏽

McFreely:
Yep, it's done.
But i'm at a loss about the coverage decreasing...

gpbl:
Ah I have no idea eithe 😄 I'll check it. Thanks again

gpbl:
I'll cut a release within the next weeks, please have some patience ...

McFreely:
Cheers, thanks for your reactivity! Have a good day!

PR #211, opened 2016-09-16
Add calendar overlay example

frandsaw:
The example of calendar overlay.
A calendar shows on input's focus.
The calendar hides on input's blur caused by outer click, pressing tab, etc.
It doesn't hide by clicking on the calendar itself.


gpbl:
Thanks @frandsaw for the example!

gpbl:
Brilliant solution @frandsaw, I've published it on http://www.gpbl.org/react-day-picker/examples/?overlay. I've updated a bit the code in #212 .

PR #216, opened 2016-09-28
Build the output files as a UMD module

pguimera:
As the title says, I've updated webpack's output to build the file as a [UMD module](https://github.com/umdjs/umd) to improve the library's compatibility when using the built file directly.


.gitignore
```diff
@@ -6,3 +6,4 @@ npm-debug.log
 _book
 lib
 _site
+.idea
```
gpbl: If you don't mind I'd skip this change, as it's not related to the source code :)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-day-picker",
-  "version": "2.4.1",
+  "version": "2.4.2",
```
gpbl: I'd prefer to cut a minor release (hence `2.5.0`)

.gitignore
pguimera: Fixed 👍

package.json
pguimera: Updated

pguimera:
I've cleaned the history a bit whilst applying the suggested changes. Also, in regards to the https://unpkg.com _integration_ there's additional steps to be made (based from this changes) which are specified in the `SUGGESTED WORKFLOW` section of said page.

gpbl:
Thanks a lot! I'll work on the unpkg thing before releasing on npm.

gpbl:
Hey @pguimera,

I've published v2.5.0. It works now from unpkg as well 🙌

pguimera:
Awesome news @gpbl, thanks for that.

I can now use this amazing library in my project without having to do weird tricks in my tooling 😄

Issue #207, opened 2016-09-08
Weekday names should respect the first day of week


emwee:
I want my calendar to let week starts on Monday. The API is pretty much straight-forward, but I can't get it to work.

My code:

```js
const localeUtils = {
	formatDay: (d, locale='en') => formatDate(d, 'LLLL'),
	formatMonthTitle: (d, locale) => moment.months()[d.getMonth()],
	formatWeekdayShort: (i, locale) =>  moment.weekdaysShort()[i],
	formatWeekdayLong: (i, locale) => moment.weekdays()[i],
	getFirstDayOfWeek: (locale) => 0,
}

const Datepicker = props =>
	<DayPicker
		localeUtils={localeUtils}
	/>
```

It doesn't matter if `getFirstDayOfWeek` returns 0 or 1, the first weekday is always Sunday.

I am using version 2.4.1.

gpbl:
Hi @emwee! 
Yeah that's not clear from the examples: if you want to set the first day on monday, also your weekdays names should start on monday as well. So I believe the following should work (I can't check right now):

```diff
const localeUtils = {
    formatDay: (d, locale='en') => formatDate(d, 'LLLL'),
    formatMonthTitle: (d, locale) => moment.months()[d.getMonth()],
-    formatWeekdayShort: (i, locale) =>  moment.weekdaysShort()[i],
-    formatWeekdayLong: (i, locale) => moment.weekdays()[i],
+    formatWeekdayShort: (i, locale) =>  moment.weekdaysShort()[i+1],
+    formatWeekdayLong: (i, locale) => moment.weekdays()[i+1],
    getFirstDayOfWeek: (locale) => 0,
}
```

Please let me know if it works so I can update the docs! Thanks.

gpbl:
(Sorry my previous comment included a code sample that didn't work, the trick is making the weeks name array starting from monday)

gpbl:
I'm filing this as bug as I agree it is an unexpected behavior  :)

emwee:
Alright, fixed it the way you suggested for now. Thanks! :-)

````

const getWeekdayByIndex = (weekdays, i) =>
	weekdays[(i + 1) < weekdays.length ? i + 1 : 0]

export const localeUtils = {
	formatDay: (d, locale='en') => formatDate(d, 'LLLL'),
	formatMonthTitle: (d, locale) => moment.months()[d.getMonth()],
[...]

PR #220, opened 2016-10-10
Ensure weekday labels respect the locale’s `firstDayOfWeek`

makenosound:
I noticed today that the `firstDayOfWeek` locale setting isn’t respected by the weekday headings. 

This change passes through that prop to the `<Weekdays>` component and then appropriately normalises it to a `0-6` integer using the modulo operator so that they do match up to the dates below.

### DIFF
--- a/src/Month.js
+++ b/src/Month.js
@@ -29,6 +29,7 @@ export default function Month({
     <div className={ className }>
       {React.cloneElement(captionElement, captionProps)}
       <Weekdays
+        firstDayOfWeek={ firstDayOfWeek }
         locale={ locale }
         localeUtils={ localeUtils }
         weekdayComponent={ weekdayComponent }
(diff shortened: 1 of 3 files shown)


gpbl:
Thanks!  Is it possible for you to add a test to reproduce this? I believe it is a fix for #207 – I need to investigate this bug better, but it looks it is a potentially breaking change.

makenosound:
@gpbl Yep, will add a test when I get a chance!

I don't think it should be a breaking change though, since it's just passing through some additional data.

makenosound:
@gpbl I’ve added a test for that now, and yep, it should fix up #207.

gpbl:
Thanks @makenosound for your help!

gpbl:
Published as v3 👍

makenosound:
No worries, @gpbl. Thanks for the great little lib.

PR #276, opened 2017-03-07
Fix aria roles on weekday and month elements

oigewan:
Fixes: #277 

Per https://www.w3.org/TR/wai-aria/roles#columnheader, the `columnheader` role should go on the equivalent of a `<th>` element and should be owned by something with the `row` role.  

Also, per https://www.w3.org/TR/wai-aria/roles#rowgroup `rowgroup` elements should be owned by something with a `grid` role, so I moved that up as well.

However, this does pose a bit of an issue for people using the `weekdayElement` prop.  Since you're not outputting the result of that in a wrapping div, people passing in an element on that prop will need to be sure to output that role themselves.

Another possible change is for the library to output the wrapping div with the classname and role.  I think that would be a breaking change (resulting change in HTML would probably throw off some people's layout CSS), so I'm not attempting that here.

cc: @gpbl

### DIFF
--- a/src/Month.js
+++ b/src/Month.js
@@ -31,7 +31,7 @@ export default function Month({
   };
   const weeks = getWeekArray(month, firstDayOfWeek, fixedWeeks);
   return (
-    <div className={ classNames.month }>
+    <div className={ classNames.month } role="grid">
       {React.cloneElement(captionElement, captionProps)}
       <Weekdays
         classNames={ classNames }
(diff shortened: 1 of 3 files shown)


gpbl:
Awesome thanks for your contribution!

> However, this does pose a bit of an issue for people using the `weekdayElement` prop

Sure, however this is left to the developer.

gpbl:
Published as v5.2.0 🎉

PR #279, opened 2017-03-08
Allow 'today' modifier to override default

maxdubrinsky:
The product I am working on allows users to specify a timezone other than that of their browsers, and in doing so might cause what the day picker component thinks `today` is to be different from what the user expects. This change simply allows me to override the `today` modifier with a custom function that could use other references, such as `moment-timezone` objects.

This was run with `test`, `cover`, and `lint`, but if there are other modifications necessary, please let me know.

Thanks, and I welcome your feedback!

### DIFF
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -421,11 +421,11 @@ export default class DayPicker extends Component {
     return React.cloneElement(navbarElement, props);
   }
   renderDayInMonth(day, month) {
-    const { today } = this.props.classNames;
     const propModifiers = Helpers.getModifiersFromProps(this.props);
     const dayModifiers = Helpers.getModifiersForDay(day, propModifiers);
-    if (DateUtils.isSameDay(day, new Date()) && !propModifiers.hasOwnProperty(today)) {
-      dayModifiers.push(today);
+    if (DateUtils.isSameDay(day, new Date()) &&
+        !Object.prototype.hasOwnProperty.call(propModifiers, this.props.classNames.today)) {
+      dayModifiers.push(this.props.classNames.today);
     }
     if (day.getMonth() !== month.getMonth()) {
       dayModifiers.push(this.props.classNames.outside);
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -421,17 +421,15 @@ export default class DayPicker extends Component {
     return React.cloneElement(navbarElement, props);
   }
   renderDayInMonth(day, month) {
-    let dayModifiers = [];
-    if (DateUtils.isSameDay(day, new Date())) {
-      dayModifiers.push(this.props.classNames.today);
+    const { today } = this.props.classNames;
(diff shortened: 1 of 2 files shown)


src/DayPicker.js
```diff
@@ -421,17 +421,15 @@ export default class DayPicker extends Component {
     return React.cloneElement(navbarElement, props);
   }
   renderDayInMonth(day, month) {
-    let dayModifiers = [];
-    if (DateUtils.isSameDay(day, new Date())) {
-      dayModifiers.push(this.props.classNames.today);
+    const { today } = this.props.classNames;
```
gpbl: @maxdubrinsky could you change the name of this var to `todayModifier`? So it's easier to understand what it is... thanks!

maxdubrinsky: Sure thing.

src/DayPicker.js
```diff
@@ -421,17 +421,15 @@ export default class DayPicker extends Component {
...
+    const propModifiers = Helpers.getModifiersFromProps(this.props);
+    const dayModifiers = Helpers.getModifiersForDay(day, propModifiers);
+    if (DateUtils.isSameDay(day, new Date()) && !propModifiers.hasOwnProperty(today)) {
```
gpbl: Please don't use `hasOwnProperty` this way or the lint will complain:
```diff
- propModifiers.hasOwnProperty(today)
+ Object.prototype.hasOwnProperty.call(propModifiers, today)
```
thanks!

gpbl:
Good idea, thanks – do you feel to add a test as well? 👍

gpbl:
> This was run with test, cover, and lint, but if there are other modifications necessary, please let me know.

Weird it didn't catch the lint, did you run

```
npm run lint
```

maxdubrinsky:
I did, but apparently I added that after I ran `lint` the first few times. I can throw a couple tests in, though.

src/DayPicker.js
maxdubrinsky: Gotcha. Just about to ask how you guys wanted that handled syntacticly. I would normally use `in`, but there are a couple of reasons not to for performance reasons.

gpbl: even if uglier I'd just use `Object.prototype.hasOwnProperty` :)

gpbl:
@maxdubrinsky are you working on this during the next... hours 😊? I'd like to release an update soon 🚀  
Otherwise I can deal with it.

maxdubrinsky:
Sorry for the delay, I will be able to push an update in the next hour or so.

gpbl:
No problem, thanks for your effort ❤️

gpbl:
Awesome thanks!

maxdubrinsky:
Thanks for being so prompt! Saved me a bunch of time and headache

gpbl:
Thanks, published as v5.2.0 🎉

PR #284, opened 2017-03-10
add example of how to disable past dates

tizol:
- add example of how to disable past dates


gpbl:
I think it would be better to change the example itself – instead of disabling weekends, disabling the past days.

tizol:
@gpbl  Sounds good 👍 !
I will change the example to disabling past days.

ChuksFestus:
its still not there tho, can somehow show me how its done?

PR #287, opened 2017-03-16
Cleanup: remove dead code

julen:
Neither `showNextMonth` nor `showPreviousMonth` handlers receive any callbacks ever.


gpbl:
Thanks for the PR! I'm closing this, as you can see from the tests, the callback is actually called [here](https://github.com/gpbl/react-day-picker/blob/master/src/DayPicker.js#L222-L224): it is needed for handling focus.

julen:
Ooops, I overlooked the output of grep, apologies!

PR #289, opened 2017-04-05
Fix RangeAdvanced example

diegobfernandez:
I tried using the RangeAdvanced example and realized that when the first day is selected and then an older date is clicked another day selection is need for it to get to a valid state and show to mouse enter "effects" correctly.

I changed it a little bit to fix it and the final result imho is a little more easy to understand, so I would like to share it with everybody.

I tried to keep it the same but, if it is of yours interest I have a version with an onRangeSelected prop.


gpbl:
Interesting, thanks ! Will check it out better later 👍🏽

gpbl:
Thanks @diegobfernandez! Much better this way.

PR #292, opened 2017-04-11
Fixed advanced modifiers example

mjomble:
Seems like it should be "first" instead of "blahblah", based on the modifier functions.


gpbl:
😁 thanks!!

PR #296, opened 2017-04-14
Add modifiers-styles

susielu:
Thanks Giampaolo for making such a great date picker component. 

I also recently ran into the case where I wanted to add inline styles. I found that for all cases I could change the styling by importing the css file like you have in your examples. The one exception was for the day modifiers. 

I wanted to propose this as a solution to the inline styles. You would pass in an object of modifiersStyles and when you go into each day, it would go into the modifiersStyles object to see if there was a corresponding style object and apply it to the day, since you've already done the test for which modifiers should be applied :)

It's working well in my local example when I built it, so feel free to use this for inline styles for days if you find it helpful.

Cheers,
Susie

### DIFF
--- a/src/Day.js
+++ b/src/Day.js
@@ -15,6 +15,7 @@ function handleEvent(handler, day, modifiers) {
 }
 export default function Day({
   classNames,
+  modifiersStyles,
   day,
   tabIndex,
   empty,
@@ -30,7 +31,6 @@ export default function Day({
   ariaDisabled,
   ariaSelected,
   children,
-  modifiersStyles,
 }) {
   let className = classNames.day;
   if (classNames !== defaultClassNames) {
@@ -41,20 +41,25 @@ export default function Day({
       .map(modifier => ` ${className}--${modifier}`)
       .join('');
   }
-  if (empty) {
-    return <div role="gridcell" aria-disabled className={className} />;
-  }
 
-  const styles = Object.assign(
-    {},
-    ...Object.keys(modifiers).map(modifier => modifiersStyles[modifier])
-  );
+  const style =
+    modifiersStyles &&
+    Object.assign(
+      {},
+      ...Object.keys(modifiers).map(modifier => modifiersStyles[modifier])
+    );
+
+  if (empty) {
+    return (
+      <div role="gridcell" aria-disabled className={className} style={style} />
+    );
+  }
 
   return (
     <div
       className={className}
       tabIndex={tabIndex}
-      style={styles}
+      style={style}
       role="gridcell"
       aria-label={ariaLabel}
       aria-disabled={ariaDisabled.toString()}
--- a/src/Day.js
+++ b/src/Day.js
@@ -30,6 +30,7 @@ export default function Day({
   ariaDisabled,
   ariaSelected,
   children,
+  modifiersStyles,
 }) {
   let className = classNames.day;
   if (classNames !== defaultClassNames) {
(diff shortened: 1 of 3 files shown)


gpbl:
Thanks for taking care of this issue! I don't understand how do you get running inline styles – could you provide an example?

susielu:
Sure, building off of the advanced modifiers example: http://react-day-picker.js.org/examples/?advancedModifiers

If you added this to the component: 

```
render() {
    const modifiers = {
      even: day => day.getDate() % 2 === 0,
      odd: day => day.getDate() % 2 !== 0,
      first: day => day.getDate() === 1,
    };
    return (
      <div>
        <span>
          <DayPicker
            modifiers={modifiers}
            onDayMouseEnter={this.handleDayMouseEnter}
            onDayClick={this.handleDayClick}
            modifiersStyles={{
               even: {
                   background: "green",
                   fontWeight: "bold
               }
           }}
          />
        </span>
        <p>
          {this.state.text}
        </p>
      </div>
    );
  }
```

Then when it goes through each day, if the modifiers have evaluated "even" for that day, it will also get the corresponding style and apply it inline to that day as well. Does that make sense? I can also push a full example if it doesn't. 

Thanks.

gpbl:
Sorry I misunderstood you – I was thinking you were applying inline styles to the whole component 😄 
Yes i think this could be a first approach to inline styles. At least, to add modifiers one wouldn't need to change the css.

I have some comments to your changes first.

Could you even add some tests to https://github.com/gpbl/react-day-picker/blob/master/test/daypicker/modifiers.js?

src/DayPicker.js
```diff
@@ -122,6 +122,7 @@ export default class DayPicker extends Component {
     canChangeMonth: true,
     reverseMonths: false,
     pagedNavigation: false,
+    modifiersStyles: {},
```
gpbl: `modifiersStyles` should have an `undefined` default value. This way we can avoid calculations when building the `style` object in the Day cell.

src/Day.js
```diff
@@ -43,10 +44,17 @@ export default function Day({
   if (empty) {
     return <div role="gridcell" aria-disabled className={className} />;
```
gpbl: The empty cell may also have a modifier's `style`. Thus, better put this lines after creating the style objects and add the `style` prop to it.

src/Day.js
```diff
@@ -30,6 +30,7 @@ export default function Day({
   ariaDisabled,
   ariaSelected,
   children,
+  modifiersStyles,
```
gpbl: please put this after `classNames` arg, thanks!

src/Day.js
```diff
@@ -43,10 +44,17 @@ export default function Day({
   if (empty) {
     return <div role="gridcell" aria-disabled className={className} />;
   }
+
+  const styles = Object.assign(
+    {},
+    ...Object.keys(modifiers).map(modifier => modifiersStyles[modifier])
+  );
+
```
gpbl: * Use `style` singular, not `styles` plural.
* Since `modifiersStyles` could be `undefined` (see my other comment), better write:

```
let style;
if (modifierStyles) {
// snip
```

just to reduce resource usage :)

susielu:
Alright, made the changes you suggested and added a test.

gpbl:
Awesome thanks 👍🏽

gpbl:
@susielu your PR has been included in `v5.3.0`. Thanks for your help improving this component!

susielu:
👏👏🦄

PR #309, opened 2017-04-26
expose modifier utility function to consumer

cwmoo740:
Great library, by the way. 

I'm working on a project that needs to validate dates based on these modifiers (user can either type into a text box, or use the react-day-picker). It would be nice if I could import a function that lets me validate if a react-day-picker modifier applies to a given date.

This is one proposal here - it doesn't add very much code, just shrinks the Helpers.js and splits up getModifiersForDay into two functions.


gpbl:
This is great, thanks! I was thinking to something similar 👍

gpbl:
Releasing this later :)

gpbl:
(I meant "few hours" as here in Peru, where it means days later 😆)

Actually I was in the middle of moving to CircleCI, and I've had to add the docs ([see here](http://react-day-picker.js.org/ModifiersUtils.html)).

The utilities are published in v5.4.0, please note that the file is called `ModifiersUtils` (with _Modifiers_, plural) instead of `ModifierUtils` (singular).

Please report back if something is not working as expected, thanks!

PR #311, opened 2017-04-27
ModifierUtils dayMatchesModifier make modifier argument optional

cwmoo740:
modifiers are often undefined or not passed in as a component prop.


gpbl:
thanks!

DimaKacer:
Hi guys! How about this issue now. It seems to me, this issue still live.

PR #318, opened 2017-05-01
ModifiersUtils export from DayPicker.js

cwmoo740:
I noticed that ModifiersUtils is not exported from the main file. I think I forgot to add that into my last PR.


gpbl:
Woah my bad for not checking this first! Thanks!

gpbl:
Published as v5.4.1

PR #339, opened 2017-05-10
Update utils.js

seymourisdead:
modifiersUtils export

Hi.

I noticed you don't add modifiersUtils export, so i add it.


gpbl:
Thanks!

gpbl:
Published as v5.5.1

PR #364, opened 2017-05-25
Add componentWillReceiveProps

youreme:
Hi,

When DayPickerInput gets nextProps, it doesn't change its states.
How about adding componentWillReceiveProps lifecycles?

After it mounted,
if it doesn't trigger any change events, the value state is different from value property.


gpbl:
Thanks! I need to add a test before merging this :)

gpbl:
Superseded by #365

PR #365, opened 2017-05-25
Update Input state when receiving new value from props

### LINKED ISSUE #363: When DayPickerInput get new props, it doesn't change its state
Hi,

When DayPickerInput gets nextProps, it doesn't change its states.
How about adding componentWillReceiveProps lifecycles?

After it mounted, 
if it doesn't trigger any change events, the value state is different from value property.

DayPickerInput

```js
componentWillReceiveProps(nextProps) {
  this.state = getStateFromProps(nextProps);
}
```

gpbl:
Closes #363


infomiho:
@gpbl When will this be merged? :smile:

Issue #366, opened 2017-05-25
Show the disabled Nav Buttons when fromMonth or toMonth props is specified


aniruddhashevle:
Basically if we add `fromMonth` or `toMonth` props then the nav buttons (which are used to navigate next or previous months) are not displayed for the months that are defined in fromMonth or toMonth props.

**This is working properly.**

However, it'd be better if there is a choice to **show those Nav Buttons even for the months that are defined in fromMonth or toMonth props** and those should be **disabled** for the respective months.

For this I've made the changes accordingly and **I'm sending the pull request.**

gpbl:
This is a good idea, I'd hide them via CSS (e.g. by setting an `--interactionDisabled` suffix to the button). It would be a breaking change, tough :(

ktusharprdxn:
@gpbl @aniruddhashevle I thik we should not remove the navigation html structure if it is disabled, so that user can set arrow image for disabled. or setting the class will also work.

gpbl:
Thanks @aniruddhashevle for your contribution! It has been published in v6.0.0 ([changelog](https://github.com/gpbl/react-day-picker/blob/master/CHANGELOG.md)) 🎉

PR #367, opened 2017-05-25
Add ability to make a choice for showing/hidding disabled Nav Buttons

aniruddhashevle:
I've added a prop named **showDisabledNavButton** which expects **boolean value(default value is false)** to make a choice **to show/hide disabled Nav Buttons** when fromMonth or toMonth props is defined.

If you wanna **show those disabled buttons define showDisabledNavButton prop with true value**, otherwise it's default value is false which hide the disabled Nav Buttons.

This is related to issue #366


gpbl:
Hi @aniruddhashevle thanks for your PR, but we won't add a prop just for this. The feature you want may be fixed in another way. Please follow the discussion in #366 thanks!

PR #373, opened 2017-05-31
Made the navButtons disabled when fromMonth or toMonth props are defined

aniruddhashevle:
I've just **shown prev and next navButtons disabled** when fromMonth or toMonth props are defined.

I've shown the **navigation html structure as it is** and added the class(disabled styling) and set onClick event null(disabled navigation functionality) if fromMonth or toMonth props are defined.


aniruddhashevle:
I've just set const previousButton as below.

    <span
      role="button"
      aria-label={labels.previousMonth}
      key="previous"
      className={
        showPreviousButton
          ? classNames.navButtonPrev
          : `${classNames.navButtonPrev} DayPicker-NavButton-disabled`
      }
      onClick={showPreviousButton ? () => previousClickHandler() : null}
    />

And on contrary, I've set for const nextButton as well.

aniruddhashevle:
The test cases fails due to the following:

 251 passing (3s)
  2 failing

  1) <Navbar /> should not render the previous button:
     AssertionError: expected { Object (root, unrendered, ...) } to have a length of 0 but got 1
      at Context.<anonymous> (test/Navbar.js:68:64)

  2) <Navbar /> should not render the next button:
     AssertionError: expected { Object (root, unrendered, ...) } to have a length of 0 but got 1
      at Context.<anonymous> (test/Navbar.js:72:64)

aniruddhashevle:
@gpbl: I'm getting the following error while testing the changes.

251 passing (4s)
  2 failing

  1) <Navbar /> should not render the previous button:
     AssertionError: expected { Object (root, unrendered, ...) } to have a length of 0 but got 1
      at Context.<anonymous> (test/Navbar.js:68:64)

  2) <Navbar /> should not render the next button:
     AssertionError: expected { Object (root, unrendered, ...) } to have a length of 0 but got 1
      at Context.<anonymous> (test/Navbar.js:72:64)

aniruddhashevle:
Related to #367

ktusharprdxn:
I think it will work without prop as per your previous PR.

prupeshprdxn:
@gpbl I required this changes in production. Can you please review this and merge asap. or atleast you can give me a workaround for now.

src/Navbar.js
```diff
@@ -15,25 +15,33 @@ export default function Navbar({
...
       role="button"
       aria-label={labels.nextMonth}
       key="right"
-      className={classNames.navButtonNext}
-      onClick={() => nextClickHandler()}
-    />;
+      className={
+        showNextButton
+          ? classNames.navButtonNext
+          : `${classNames.navButtonNext} DayPicker-NavButton-disabled`
```
gpbl: Here and in the change above, please use `${classNames.navButtonInteractionDisabled}` and set it to `DayPicker-NavButton--interactionDisabled` in `classNames.js`.

src/Navbar.js
```diff
@@ -15,25 +15,33 @@ export default function Navbar({
...
+      }
+      onClick={showNextButton ? () => nextClickHandler() : null}
```
gpbl: Please use `undefined` instead of `null` (here and in the change above)

src/style.css
```diff
@@ -46,6 +46,7 @@
...
     }
 
+    .DayPicker-NavButton-disabled { opacity: .5; }
```
gpbl: Please use `.DayPicker-NavButton--interactionDisabled` as selector name and set `display: none` to replicate current the behavior (as default, the navigation button must be stay hidden)

gpbl:
@aniruddhashevle would you mind to fix the tests? Instead of `should not render the previous button` it `should disable interaction with the previous button`, thus have the `--disableInteraction` class and the `onClick` event handler should not work. Thanks!

aniruddhashevle:
@gpbl : Yes. I'll do it by tomorrow. Thanks.

aniruddhashevle:
@gpbl: I've added the changes, please review: https://github.com/gpbl/react-day-picker/pull/398

PR #401, opened 2017-06-14
Accessibility: Days should have a negative tabIndex by default

### LINKED ISSUE #400: Users are able to focus days in other months
When a user first tabs to a calendar with `enableOtherDays` set, focus initially lands on the first day in another month.

https://jsfiddle.net/hhjrc0qp/

oigewan:
Fixes #400

**Bug:**
To properly control tab behavior, only one of the days should have a tabIndex > -1.  However, when a user first tabs to a calendar with `enableOtherDays` set, focus initially lands on the first day in another month.

**Root Cause:**
Since the Month component only explicitly sets a tabIndex for days in the current month and the Day component defaults the tabIndex to 0, days in other months can be tabbed to.

**Changes:**
- Always pass a tabIndex to the Day component.  Default it to -1.
- Remove the `tabIndex|| 0` when outputting the day, and add a defaultProp which sets the tabIndex to -1.
- Add a test.

cc: @gpbl

### DIFF
--- a/src/Day.js
+++ b/src/Day.js
@@ -40,6 +40,10 @@ export default class Day extends Component {
   };
 
   static defaultProps = {
+    tabIndex: -1,
+  };
+
+  static defaultProps = {
     modifiers: {},
     empty: false,
   };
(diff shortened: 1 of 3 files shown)


gpbl:
Thanks for the PR and the detailed summary ✨ 
Got to fix that weird error with coveralls to pass the build on CircleCI first 😱

PR #404, opened 2017-06-14
IE11: Update babel config to use the runtime

### LINKED ISSUE #403: Bug: Fails to load in IE11
I was testing my app in IE11, and as soon as I attempted to move between days using the keyboard, my console was full of errors.  It seems that your babel settings aren't polyfilling anything, and your output code includes things like `Object.assign` and `Array.from` which don't exist in IE11.

oigewan:
**Bug:**
Fixes #403 

As mentioned in the issue, this library breaks in IE11 due to missing polyfills.  You could just document the polyfill requirements for this library, but since IE 11 is still [over 2% of browser usage](http://caniuse.com/usage-table) and this is a library, I think it'd be a good idea to set up babel to use polyfills and helpers.

I also had issues testing changes to the library.  When I ran `npm link` to test my changes, I kept getting errors saying that webpack could not find the "react" library.  I looked and saw that React is listed as a peerDependency.  However, since React is imported by the library's code and is required in order for this library to compile properly, I'm thinking it should be a dependency.

**Notes:**
[Babel documentation](http://babeljs.io/docs/plugins/transform-runtime/) on the plugins I'm using and dependencies I added.

**Changes:**
- Add the transform-runtime plugin configured to use the helpers and polyfills.
- Add a dependency on the babel-runtime.
- Moved React from a peerDependency to a dependency.

cc: @gpbl

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,9 +63,6 @@
     "url": "https://github.com/gpbl/react-day-picker/issues"
   },
   "homepage": "https://react-day-picker.js.org",
-  "peerDependencies": {
-    "react": "~0.13.x || ~0.14.x || ^15.0.0"
-  },
   "devDependencies": {
     "@types/react": "^15.0.27",
     "autoprefixer": "^7.1.1",
@@ -74,6 +71,7 @@
     "babel-eslint": "7.2.3",
     "babel-loader": "7.0.0",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.5",
+    "babel-plugin-transform-runtime": "^6.23.0",
     "babel-preset-es2015": "6.24.1",
     "babel-preset-react": "6.24.1",
     "babel-preset-stage-1": "6.24.1",
@@ -113,7 +111,9 @@
     "webpack": "^2.6.1"
   },
   "dependencies": {
-    "prop-types": "^15.5.10"
+    "babel-runtime": "^6.23.0",
+    "prop-types": "^15.5.10",
+    "react": "~0.13.x || ~0.14.x || ^15.0.0"
   },
   "precommit": "lint-staged",
   "lint-staged": {
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,12 @@
 {
   "presets": ["react", "es2015", "stage-1"],
+  "plugins": [
+    ["transform-runtime", {
+      "helpers": true,
+      "polyfill": true,
+      "regenerator": false
+    }]
+  ],
   "env": {
     "production": {
       "plugins": ["transform-react-remove-prop-types"]


package.json
```diff
@@ -63,9 +63,6 @@
     "url": "https://github.com/gpbl/react-day-picker/issues"
   },
   "homepage": "https://react-day-picker.js.org",
-  "peerDependencies": {
-    "react": "~0.13.x || ~0.14.x || ^15.0.0"
```
gpbl: Any reason why you are removing react from `peerDependencies`?

package.json
```diff
@@ -113,7 +111,9 @@
     "webpack": "^2.6.1"
   },
   "dependencies": {
-    "prop-types": "^15.5.10"
+    "babel-runtime": "^6.23.0",
```
gpbl: Do you know the impact this is having to the final bundle size? I'd rather remove the stuff not supported by IE11.

package.json
oigewan: Mentioned in the PR description, but I'll copy/paste here:
> I also had issues testing changes to the library. When I ran npm link to test my changes, I kept getting errors saying that webpack could not find the "react" library. I looked and saw that React is listed as a peerDependency. However, since React is imported by the library's code and is required in order for this library to compile properly, I'm thinking it should be a dependency.

gpbl: Sorry I missed that part 😄 

React [is included](https://github.com/oigewan/react-day-picker/blob/4f6ce4463f60ab3072bc0d5f142a67cb93819f11/package.json#L102) as `devDependencies`, which should be installed when developing the module. So definitely react must stay as `peeDependency` here.

package.json
oigewan: The bundle size did go up a noticeable amount.  I did some investigation on what would need to be done.  The `Array.from` references are actually introduced by a babel plugin, so it's affects aren't quite easy to see.

I'll close this PR and open another one as soon as I'm capable of testing my changes.

gpbl:
@oigewan [These lines](https://github.com/gpbl/react-day-picker/blob/master/src/Day.js#L84-L91) in `Day.js` could be use the [object-assign](https://github.com/sindresorhus/object-assign) module (to add as dependency). 

`Array.isArray`is actually [supported in IE11](https://docs.microsoft.com/en-us/scripting/javascript/reference/array-isarray-function-javascript#requirements) so it should work.

I'm still wondering where does this `Array.from` come from 🤔

gpbl:
@oigewan ah the `Array.from` comes from spreading arrays, e.g. [here](https://github.com/gpbl/react-day-picker/blob/master/src/DayPicker.js#L275). We can also fix that without adding other dependencies...

oigewan:
Yup.  New PR: https://github.com/gpbl/react-day-picker/pull/406

PR #406, opened 2017-06-15
Update code to ensure IE11 compatibility

### LINKED ISSUE #403: Bug: Fails to load in IE11

oigewan:
**Bug:**
Fixes #403 

As mentioned in the issue, this library breaks in IE11 due to missing polyfills.  You could just document the polyfill requirements for this library, but since IE 11 is still [over 2% of browser usage](http://caniuse.com/usage-table) and this is a library, it'd be a good idea to remove code that's incompatible with IE11.

Unfortunately, some babel plugins also add code that's incompatible by default.  The main culprit here is the use of the spread operator to clone a NodeList to an array.  Babel uses a helper for the spread operator, and that helper uses `Array.from`.  Therefore, using the spread operator to clone an array isn't something that can be done here any more.

**Notes:**
[Babel documentation](http://babeljs.io/docs/plugins/transform-runtime/) on the plugins I'm using and dependencies I added.

**Changes:**
- Install the `object-assign` ponyfill and use it instead of `Object.assign`
- Replace all usages of the `[...dayNodes].indexOf(...)` code with something equivalent, but more IE friendly.

cc: @gpbl

### DIFF
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -272,7 +272,7 @@ export default class DayPicker extends Component {
 
   focusPreviousDay(dayNode) {
     const dayNodes = Helpers.getDayNodes(this.dayPicker, this.props.classNames);
-    const dayNodeIndex = [...dayNodes].indexOf(dayNode);
+    const dayNodeIndex = Helpers.nodeListToArray(dayNodes).indexOf(dayNode);
 
     if (dayNodeIndex === 0) {
       this.showPreviousMonth(() => this.focusLastDayOfMonth());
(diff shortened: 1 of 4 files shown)


src/DayPicker.js
```diff
@@ -12,6 +12,10 @@ import classNames from './classNames';
 import keys from './keys';
 import PropTypes, { ModifierPropType } from './PropTypes';
 
+function nodeListToArray(nodeList) {
```
gpbl: Would you mind to move this to `Helpers.js`? Thanks!

oigewan: Done

oigewan:
@gpbl Issue addressed.  Did a rebase and force pushed to the branch

gpbl:
perfect 👌🏽

oigewan:
Thanks for reviewing and merging @gpbl.  Any idea on when you're gonna do your next release?

PR #414, opened 2017-06-17
Update DayPickerInputAPI.md

zorahrel:
Custom components must have an instance. To manage focus is used `ref` prop, but as wrote [on React documentation about refs](https://facebook.github.io/react/docs/refs-and-the-dom.html#refs-and-functional-components) in order to use `ref` you should use a Class component.
So documentation should not tell that the component can be of type `function`.


gpbl:
Sure, my bad! Thanks for the PR 👍

PR #416, opened 2017-06-19
Single day range + better reset behaviour

olance:
A small "fix", which I think makes the range example a bit more usable and intuitive:
- Single-day ranges can be selected
- Any click within the current range will reset the range
- Any click outside the current range will still start a new range


gpbl:
Awesome thanks!

PR #433, opened 2017-06-26
Update utils-locale.md

yagisuke:
Modify the sample source.
Because, sample source fails to get the 'year'.

react-day-picker is very cool.
thanks!


gpbl:
Thanks 👍

Issue #446, opened 2017-07-05
Day.shouldComponentUpdate is buggy, slow, and probably unnecessary


oigewan:
The `<Day />` component has a buggy `shouldComponentUpdate` method.  As it stands, the component only checks to see if modifiers have been changed.  However, if ANY other prop changes (children, empty, etc) then this will not re-render.  I've put together a quick example here: https://jsfiddle.net/afL5mqcL/  As you can see, changing the value of the `enableOutsideDays` prop does nothing since the `<Day />` will not re-render.

I thought about putting together a PR that improves this situation, but after more thought, I don't even thing the `shouldComponentUpdate` method is necessary.  `<Day />` lives pretty low in the render tree (only the customizable `renderDay` renders as a child) and the current implementation does a sort which is O(n * log(n)) followed by a join which will create one or more strings in memory on both the modifiers objects.  All this to avoid outputting a div and running the renderDay function.  However, when things do need to update, you've done the string joins twice as well as additional, unnecessary sorts.

All in all, I think that we should just remove this check entirely as it doesn't seem to be providing the value that you're looking for.

gpbl:
Thanks for the feedback! 

The `shouldComponentUpdate` has been added there as [re rendering](https://github.com/gpbl/react-day-picker/issues/389#issuecomment-307650746) is slow especially when displaying multiple months. Usually a day has 0-2 modifiers so checking that objects shouldn't be as expensive (the `render` function does some logic).

Instead of removing SCU I'd add a shallow compare for `children`, `empty` and  even `ariaLabels` and `modifierStyles`. It would be nice, however, have a deeper analysis of the performance impact for both solutions, being not a fan of SCU either :)

oigewan:
I've created another PR that does full, proper checking of prop equality before determining whether or not to render the day.  As you can see, it's a bit more involved.  However, given the API and the user expectation that changing a prop will cause the Calendar to update, I'm not sure there's another implementation that's better....

gpbl:
Published as `v6.1.0` 😊

PR #452, opened 2017-07-07
Bugfix: Better Day.shouldComponentUpdate implementation and exporting…

oigewan:
**Bugs:**
Fixes https://github.com/gpbl/react-day-picker/issues/446

The `Day.shouldComponentUpdate` method was not allowing days to update when anything other than modifiers changed.  This can lead to buggy behavior if many public props of DayPicker change.

Also, the NavbarPropTypes and WeekdayPropTypes were being exported incorrectly starting here: https://github.com/gpbl/react-day-picker/commit/0b5a46f91faf5179c945e97a3fd273982143f237

**Changes:**
- Export propTypes directly off of components inside of `DayPicker.js` where appropriate.
- For most props in `Day`, do a shallow compare to determine the return value of `shouldComponentUpdate`.  However, `modifiers` and `modifiersStyles` need their own shallow compare as well, so make an exception for them.

cc: @gpbl


gpbl:
Thanks for the help! 🌈

Issue #457, opened 2017-07-13
On-the-fly Application of Date Validity


arjun-menon:
Note: *This is not an issue, but a question regarding a particularly tricky situation and how to handle it.* In my use case scenario, some dates are valid, and some dates are invalid. Unfortunately the existing [date disabling](http://react-day-picker.js.org/examples/disabling.html) mechanism is too simple to handle my scenario. I feel `react-day-picker` might benefit from a more sophisticated date-disabling mechanism. I explain my use case scenario below:

The server has a list of dates that are valid, but it's a very large list that contains the valid dates for several years.

To make things efficient, I am thinking that the client side (web app), should ask the server for a small number of valid dates, dynamically, *as necessary*.

For example, if my `initialDate` for `react-day-picker` is in July 2017, I will ask the server first, which dates are valid in June, July, and August of 2017. Then if the user navigates to February 2016, I would send another request to the server, asking: give me the valid dates of January, February, and March 2016. (The reason I ask for valid dates for the month before/after is so that, the UI is fast/responsive when the user goes to the next/previous month).

In addition, I also want `react-day-picker` to skip months in which there are no valid dates. So that if August 2017 has no valid dates, then, it shows September after July. Similarly if a whole year doesn't have valid dates, it skips that whole year. Basically I'd like it to skip months, until finding a month containing valid dates.

In order to implement checking for valid dates *on-the-fly* (as the user switches between months), it would be helpful to be able to pass in a function (via props) that looks something like `getValidDates(year, month, callback)`. Basically, `react-day-picker` would call this function several times (as necessary) to get valid dates.

As the user navigates through the calendar, `react-day-picker` would call the function, and perhaps display a spinning loading symbol/indicator (that could be passed via prop?), and wait until `callback` is invoked with the list of valid dates before rendering the calendar.

gpbl:
👋 there! 

Not sure I'm following - what is stopping you to implement your requirements? I believe the API gives you all what you need.

There's a `onMonthChange` prop that you can use to fetch data from the server when the calendar displays another month.

You can use the `showMonth` method to change the month and eventually skip the one without valid dates.

Good luck 👍🏽

arjun-menon:
[...]

PR #458, opened 2017-07-14
Allow users to intercept and block showMonth

arjun-menon:
I've added a new prop of type function, called `shouldMonthChange`. This function can be used to intercept and block `showMonth`. If `shouldMonthChange` returns `false`, then `showMonth` does nothing.

In addition, a third parameter `force` has been added to `showMonth` which when truthy would result in `shouldMonthChange` being ignored.

Also, the arguments passed to `showMonth` are propagated to `shouldMonthChange`, so that if it needs to, it can call `showMonth` itself.

The use case scenario that lead me to make this pull request is described in #457.

### DIFF
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -23,6 +23,7 @@ export default class DayPicker extends Component {
     numberOfMonths: PropTypes.number,
     fromMonth: PropTypes.instanceOf(Date),
     toMonth: PropTypes.instanceOf(Date),
+    shouldMonthChange: PropTypes.func,
     canChangeMonth: PropTypes.bool,
     reverseMonths: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


gpbl:
This `shouldMonthChange` prop seems to clash with `canChangeMonth` :) I was thinking about a `onBeforeMonthChange` prop. This prop should receive a function returning a promise. When the promise is resolved, the month will change.

```jsx
function handleMonthChange() {
   return new Promise.resolve();
}
<DayPicker onBeforeMonthChange={handleMonthChange} />
```

arjun-menon:
That sounds good. I'll rewrite the PR based on that.

gpbl:
This issue is too old and there's nothing actionable to do for now – closing. Please open a new issue if required, thanks.

Issue #462, opened 2017-07-19
Feature Request: Allow node consumers to remove propTypes for production builds


oigewan:
I can see that you're using the transform-react-remove-prop-types babel plugin for the webpack builds, but not for the `build:node` task.  Fortunately, recent versions of the plugin allow you to wrap propType assignments in a `process.env.NODE_ENV !== 'production'` ternary.  Docs: https://www.npmjs.com/package/babel-plugin-transform-react-remove-prop-types#mode

I'd like to start using that in the babel build for this component to allow node consumers to remove propTypes for their production build.  This would require node consumers to inject the value of `process.env.NODE_ENV` via webpack's [Environment Plugin](https://webpack.js.org/plugins/environment-plugin/) or rollup's [Replace Plugin](https://github.com/rollup/rollup-plugin-replace), but most consumers will probably be doing that any way.

PR #463, opened 2017-07-19
Use babel-plugin-transform-react-remove-prop-types for build:node task

oigewan:
**Feature:**
Fixes https://github.com/gpbl/react-day-picker/issues/462

This will wrap all assignments to the propTypes property of a React component class in a `process.env.NODE_ENV !== "production"`, which will allow npm module consumers to remove prop types for their production build.  This will allow for smaller deployables for consumers.

**Known Issue:**
This will require users to inject the value for `process.env.NODE_ENV`.  Consumers can do this via webpack's [Environment Plugin](https://webpack.js.org/plugins/environment-plugin/) or rollup's [Replace Plugin](https://github.com/rollup/rollup-plugin-replace).

**Changes:**
Add `transform-react-remove-prop-types` to the `.babelrc` file with the mode set to "wrap".  Output will end up looking like this:
```js
exports.default = Caption;
Caption.propTypes = process.env.NODE_ENV !== "production" ? {
  date: _propTypes2.default.instanceOf(Date),
  months: _propTypes2.default.arrayOf(_propTypes2.default.string),
  locale: _propTypes2.default.string,
  localeUtils: _propTypes2.default.object,
  onClick: _propTypes2.default.func,
  classNames: _propTypes2.default.shape({
    caption: _propTypes2.default.string.isRequired
  }).isRequired
} : {};
```

cc: @gpbl


gpbl:
thanks!

PR #471, opened 2017-07-27
Navbar - fix for disabled interaction in RTL

edoshor:
The RTL support in Navbar lacks symmetry when dealing with disabled interaction with the previous and next buttons.

The fix swaps the values of the showPreviousButton and showNextButton props on RTL.


gpbl:
Thanks! sorry for the delay ✌🏽

Issue #483, opened 2017-08-09
Add flow typings


TrySound:
I'm flow user and would be great to have typings in this repo like for typescript. Do you want me to make a PR?

gpbl:
That would be awesome 👍🏽 go for it! If you find some troubles please write back here.

gpbl:
@TrySound should we give up with flow types? I don't see much interested for this as for TypeScript typings.

TrySound:
@gpbl Nope, I'll take care of it for major release.

gpbl:
Sounds great! Thanks!

arjun-menon:
Would the types be added here, or to the [https://github.com/flowtype/flow-typed](flow-typed) project?

gpbl:
@arjun-menon the idea is to add them here

gpbl:
Makes flow type still sense? 

f we want still to add flow-types then yes, better in https://github.com/gpbl/react-day-picker/issues/flow-typed.

PR #484, opened 2017-08-09
Add flow types definition

TrySound:
Ref https://github.com/gpbl/react-day-picker/issues/483


TrySound:
@gpbl Is it good for you?

adidahiya:
sigh. this makes me sad because the syntax is almost identical to the typescript definitions that already exist in this project and now we're stuck maintaining both. not sure what can be done about it though...

TrySound:
Syntax is similar but react libs are quite different.

adidahiya:
> react libs are quite different

what do you mean? do you mean that react's flow typings are quite different from `@types/react`? how so?

TrySound:
Lot of stuff. In flow Synthetic[Mouse|Keyboard|Input...]Event events are global and doesn't have target generics. And another React stuff has different names. Also typescript namespaces with synthetic exports are incompatible with flow and will be broken with native modules some day.

TrySound:
@gpbl do you have a plan to publish this?

stunaz:
@gpbl what is holding you from publishing this?

TrySound:
We need to upgrade them to flow55 and rebase pr. I don't have a time right now.

gpbl:
We moved to typescript: https://github.com/gpbl/react-day-picker/pull/944 no plans to add flow anymore... thanks!

PR #485, opened 2017-08-17
Fix typed definition files

lpcarignan:
I added `DayPickerInput` type definition (and its properties) in `index.d.ts`. I also moved `index.d.ts` to root of project as this seem to make things work in Typescript project.

I've explicitly defined variables `DayPicker` and `DayInputPicker` in `module.export` variable. This helped Typescript to find the components.

I've commented out the code in `DayPickerInput.js` as it seemed to conflict in Typescript project when I'm importing both controls (DayPicker & DayPickerInput). I would remove this file but not sure if it impacts Javascript projects.

I had to move `tsconfig.json` to the root of the project as `dtslint` calls this file when it starts. I tried changing the location of index.d.ts in tsconfig.json but dtslint wasn't cooperating.

I ran `yarn test` and it passed. I ran `yarn dtslint` too and I had no errors.


DayPicker.js
```diff
@@ -2,14 +2,16 @@
 /* eslint-env node */
 
 var DayPicker = require('./lib/src/DayPicker');
+var DayPickerInput = require('./lib/src/DayPickerInput');
```
TrySound: This will force users to keep large moment js in their bundle. It's bad.

lpcarignan: I share your concern @TrySound 

It could be 'my bad' so here's what I'm doing in my sample [Typescript project](https://github.com/lpcarignan/DayPickerInputTypescript).

I compile Typescript and use Browserify to put everything in a single `index.js` file that I load in the browser. Browserify doesn't include the code in `DayPickerInput.js`. It only includes code from `DayPicker.js`. I can't explain why. I'm not familiar enough with how:

- Typescript transcompile `import { DayPicker, DayPickerInput, DayModifiers } from "react-day-picker";`  to Javascript
- What the `require("react-day-picker")` Javascript statement actually does behind the scenes
- Why only `DayPicker.js` is only loaded by `require("react-day-picker")`

If you hava some insights on this, feel free to share them with me and I can try them in the sample Typescript project.

TrySound:
@gpbl I see the solution in removing moment is in favor of simple set/get pair for formatting and parsing.

DayPicker.js
adidahiya: Please don't change the shape of the module export like this -- it's a breaking change. If you want to import `DayPickerInput` you'll need to do it as `import { DayPickerInput } from "react-day-picker/DayPickerInput";` and define the types like this:

```
// index.d.ts stays the same
```

```
// DayPickerInput.d.ts
export class DayPickerInput ... {
    ...
}
```

I do think moving the `types/` folder contents to the root will be necessary for this, but you could make that happen only in the distribution if @gpbl doesn't want those at the root of the source repo.

gpbl:
I believe @adidahiya got the point, thanks for the feedback 💟  maybe we've solved the issue!
No problem moving the type definition files to the root, if this is the only way to do it – however I wonder why this is necessary.

@lpcarignan could you please 
- [ ] revert the changes in `DayPicker.js`, `DayPickerInput.js` and `index.d.ts`
- [ ] add `DayPickerInput.d.ts` as suggested by @adidahiya 
- [ ] add `DayPickerInput.d.ts` to the `files` and `types` properties in `package.json`
- [ ] use 2-spaces instead of 4 when formatting the code 🙆 

thanks!

lpcarignan:
I've done the changes that you ask @gpbl.  

Still learning GitHub but I don't believe I need to do a new pull request.

I'll let you guys review them and hopefully, it makes it to the next release.

Thank you for all of your help

index.d.ts
```diff
@@ -1,32 +1,31 @@
 // TypeScript Version: 2.2
 import * as React from 'react';
 
-declare namespace DayPicker {
```
adidahiya: in general this is a good change because it retains backwards compatibility for the top-level symbol exports.

_however_, you need to make the export include the `DayPicker` class, and the only way to do that is with `export = DayPicker` and declaration merging. so you're going to need to keep this `namespace` as it is. I recommend putting the `DayPicker` class in new file and importing it as `import { DayPicker } from "./DayPicker"`. when you import it here, it will merge with this namespace and you can `export = DayPicker` at the bottom. the diff in this file should be fairly small (not 100s of lines).

DayPicker.js
```diff
@@ -9,7 +9,7 @@ var Weekday = require('./lib/src/Weekday').default;
 var Navbar = require('./lib/src/Navbar').default;
 var PropTypes = require('./lib/src/PropTypes');
 
-module.exports = DayPicker.default || DayPicker;
+module.exports.DayPicker = DayPicker.default || DayPicker;
```
lpcarignan: If I go back to `module.exports = DayPicker.default || DayPicker`, my sample Typescript application doesn't load the DayPicker anymore.

I get this React error message:
`
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in. Check the render method of...
`

I reverted to your code and I tried the following to solve the React error from above:

- Set the keyword `default` in `export default class DayPicker ...` in index.d.ts
- Tried `import { default as DayPicker}` from "react-day-picker";` in main.tsx of sample application 

Both of them did not work and I get the same error.

From my understanding, `module.exports = DayPicker.default || DayPicker` is like assigning the default value to object `exports`.

Would you have an idea on how to import this so it works in Typescript?

tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+      "es6",
+      "dom"
+    ],
+    "noImplicitAny": true,
+    "noImplicitThis": true,
+    "strictNullChecks": true,
+    "noEmit": true,
+    "forceConsistentCasingInFileNames": true
+  },
+  "files": [
```
lpcarignan: Done

index.d.ts
lpcarignan: I'm terribly sorry @adidahiya but I do not understand your recommendation on this point.

If I move `DayPicker` (and its DayPickerProps) to its own `DayPicker.d.ts`, I cannot import it in `index.d.ts`. It would have to be the other way around as `DayPickerProps` needs the definitions in index.d.ts.

Would it be possible to help me better understand the change that you would like me to do?

I did put back the `declare namespace DayPicker` with the `export = DayPicker;` at the end of the file. I tested it out in my sample Typescript app and it works. Coming from a C#/Java background, I would comment that it is a bit odd to see `DayPicker.DayPicker` but I'm relatively new to Javascript/Typescript so I'm open to leaving it back as it is.

lpcarignan: @adidahiya Was I clear enough explaining the problem I am encountering following your recommendation? If time permits, I would be more than happy to have your input so I can move forward with this.

adidahiya: I made a new PR to provide the same funtionality as this one: https://github.com/gpbl/react-day-picker/pull/487

My declaration merging idea didn't work out, but you should be able to use those types and import like this:

```ts
import * as DayPicker from "react-day-picker";
import * as DayPickerInput from "react-day-picker/DayPickerInput";
```

gpbl:
Closing this as it has been replaced by #487

Issue #477, opened 2017-08-06
Allow a custom component as overlay wrapper


gpbl:
Add a new `overlayComponent` prop defaulting to `div` to the `DayPickerInput` component. This could be used to [fix alignments](https://github.com/gpbl/react-day-picker/issues/536) or add custom content to the overlay.

PR #486, opened 2017-08-22
Add a overlayWrapper prop to DayPicker for greater customization.

wldcordeiro:
This PR tries to address #477 by adding a `overlayWrapper` prop that allows you to customize the component wrapping the `DayPicker` within `DayPickerInput`


src/DayPicker.js
```diff
@@ -507,25 +511,26 @@ export default class DayPicker extends Component {
...
-          onBlur={this.props.onBlur}
-        >
-          {this.renderNavbar()}
-          {this.renderMonths()}
-        </div>
-      </div>
+        {this.renderNavbar()}
+        {this.renderMonths()}
+      </div>,
+      containerProps.children ? containerProps.children : null
```
gpbl: 🤔 what if the calendar should be placed after the `containerProps.children`?

gpbl:
Hi, thanks for this PR 🙏

I was thinking to add a new prop to `DayPickerInput`, not implementing it into `DayPicker`. 

Indeed, wrapping `DayPicker` inside a component can be already done without using the `containerProps`' `children`:

```jsx
<MyWrapper>
   <DayPicker />
</MyWrapper>
```

What we need is a different wrapper for the overlay, [here](https://github.com/gpbl/react-day-picker/blob/efd1d6f0150c2d52c2139dedc378963d43808811/src/DayPickerInput.js#L267-L277) where we are rendering the DayPicker in `DayPickerInput.js`:

https://github.com/gpbl/react-day-picker/blob/efd1d6f0150c2d52c2139dedc378963d43808811/src/DayPickerInput.js#L267-L277

For example, I was thinking the prop to be a function and working this way:

```jsx
<DayPickerInput
  wrapperComponent={ ({children, ...props)} => 
     <div {...props}>
        <p>This is placed after the calendar</p>
        <div>{ children }</div> /* this will render the calendar */
        <p>This is placed after the calendar</p>
     </div>
  }
/>
```

Maybe it's better call the prop `overlayWrapper` instead?

wldcordeiro:
Updated the PR, the prop is now called `overlayWrapper` and it's on the `DayPickerInput`. I've added a basic test along with a documentation entry for it.

gpbl:
Hey, thanks for this – I've fixed some parts in #563. Sorry for the delay...

gpbl:
Thanks @wldcordeiro for your help! I’ve published the new prop in the v7.

Issue #494, opened 2017-09-22
DayPickerInput Update From New Props


kradical:
Hello,

First of all thank you for maintaining this project! It has saved me a lot of time :).

I am currently using it in an app that uses redux-form everything works awesome except when the props get updated the logic of `DayPickerInput.componentWillReceiveProps` relies on state rather than props.

I will open up a PR and link it here. I believe I have a good fix.

🐛 If you are reporting a bug, show us how to reproduce it here:
https://codesandbox.io/s/v10o1xnm95
* type in `01/01/2007`
* click the header to change the month to February
* type in `01/01/2009`

No update of the DayPickerInput.

gpbl:
Closing this as #495 has been merged. Thanks @kradical for your contribution 💯

gpbl:
The fix has been published as in v6.2.0.

kradical:
I think it is in 6.1.1, Inspecting the code for tag 6.1.1 it seems to be. And I updated to 6.1.1 and could remove my hacky workaround! Cheers!

PR #495, opened 2017-09-22
DayPickerInput check for prop change with previous props rather than previous state

kradical:
Hello!

I filed this issue: https://github.com/gpbl/react-day-picker/issues/494.

I have tested this with dayPickerProps and without dayPickerProps.

### DIFF
--- a/src/DayPickerInput.js
+++ b/src/DayPickerInput.js
@@ -78,28 +78,28 @@ export default class DayPickerInput extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    const { dayPickerProps: { month }, value } = this.props;
-    const {
-      dayPickerProps: { month: nextMonth },
-      value: nextValue,
-    } = nextProps;
+    const { dayPickerProps, value } = this.props;
+    const { dayPickerProps: nextDayPickerProps, value: nextValue } = nextProps;
 
     const hasDifferentValue = nextValue !== value;
 
-    const monthExistsOnBoth = nextMonth && month;
+    const willUpdateDayPickerMonth =
+      nextDayPickerProps.month && dayPickerProps.month;
 
-    const newMonthProp = nextMonth && !month;
+    const newMonthProp = nextDayPickerProps.month && !dayPickerProps.month;
 
     const shouldDisplayAnotherMonth =
       newMonthProp ||
-      (monthExistsOnBoth &&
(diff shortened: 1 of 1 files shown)


src/DayPickerInput.js
```diff
@@ -78,21 +78,28 @@ export default class DayPickerInput extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    const { month, value } = this.state;
-    const hasDifferentValue = nextProps.value !== value;
+    const { dayPickerProps: { month }, value } = this.props;
```
gpbl: To make the code easier to understand, I usually avoid destructuring an object at the second level: could you please here just use:

```js
const { dayPickerProps, value } = this.props;
```

and update the code below? Thanks!

src/DayPickerInput.js
```diff
@@ -78,21 +78,28 @@ export default class DayPickerInput extends React.Component {
...
+    const {
+      dayPickerProps: { month: nextMonth },
+      value: nextValue,
+    } = nextProps;
+
+    const hasDifferentValue = nextValue !== value;
+
+    const monthExistsOnBoth = nextMonth && month;
```
gpbl: So this would become:

```js
const monthExistsOnBoth = nextDayPickerProps.month && dayPickerProps.month;
```

I'd change the variable name from `monthsExistsOnBoth` to `willUpdateDayPickerMonth` (I don't understand what `both` are or why a month should _exist_ or not :))

src/DayPickerInput.js
```diff
@@ -78,21 +78,28 @@ export default class DayPickerInput extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    const { month, value } = this.state;
-    const hasDifferentValue = nextProps.value !== value;
+    const { dayPickerProps: { month }, value } = this.props;
+    const {
+      dayPickerProps: { month: nextMonth },
+      value: nextValue,
+    } = nextProps;
```
gpbl: Here also:

```js
dayPickerProps: nextDayPickerProps,
```

I know it's a bit more verbose, but it would really help!

src/DayPickerInput.js
kradical: Will do!

src/DayPickerInput.js
kradical: very true, will update tomorrow!

gpbl:
Thanks!

PR #498, opened 2017-09-26
Add React 16 as peer dependency

brycehill:
I didn't see an open issue/PR regarding React 16. All the tests passed and I didn't see any errors when using react-day-picker with React 16 in my app.


gpbl:
thanks!

PR #501, opened 2017-09-30
Upgrade tests for React 16

gpbl:


gpbl:
Some tests are failing, possibly because of this https://github.com/airbnb/enzyme/issues/1201

TrySound:
@gpbl Seems like these changes or at least warning fix wasn't published in 6.2.0. Don't see it in node_modules.

gpbl:
@TrySound thanks for the heads up! Please use v6.2.1. 

I forgot to push the built files for v6.2.0, so you were actually installing v6.1.1 🥇

PR #504, opened 2017-10-10
Fix modifiers when using cssmodules

tume:
Had problems using DayPickerInput with CSS modules so was just copying what was done here https://github.com/gpbl/react-day-picker/pull/450/ for daypicker.js


gpbl:
Thanks a lot! Could you please run `yarn lint -- --fix` and add a test case? 🙏

gpbl:
Replaced by #517

ethanve:
@gpbl I'm still having issues with this in v7.1.6 -- what's the status here?

Issue #508, opened 2017-10-14
Move input related props to `inputProps`


gpbl:
In DayPickerInput, we are spreading input props from the component, e.g.:

```jsx
// Will render 
// <input name="birthday" autocomplete={false} placeholder="Enter a date" />
<DayPickerInput 
    name="birthday"
    autocomplete={false}
    placeholder="Enter a date"
/>
```

We are also using props that are not compatible with `input`, or that may override `input` props:

```jsx
<DayPickerInput 
    format="LL" /* format will not be passed down to <input> */
    name="birthday"
    autocomplete={false}
    placeholder="Enter a date"
/>
```

This may create issues as React 16 now supports custom dom attributes should pass https://reactjs.org/blog/2017/09/26/react-v16.0.html#support-for-custom-dom-attributes

So to fix it, we should add a `inputProps` prop similar to what `dayPickerProps` does:

**Before**

```jsx
<DayPickerInput 
    name="birthday"
    autocomplete={false}
    placeholder="Enter a date"
/>
```

**After**

```jsx
<DayPickerInput 
    inputProps={{
        name: 'birthday',
        autocomplete: false,
        placeholder: 'Enter a date'
    }}
/>
```

abiduzz420:
Can I take up this issue ? @gpbl

gpbl:
@abiduzz420 that would be awesome! 

To keep this in the next minor release, please don’t change the current behaviour - just spread `props.inputProps` to the `input` element. In the next major release, we can remove the current behavior.

seeliang:
hi @gpbl, I could not get the autoComplete set work unless 
```javascript 
Input,
            _extends(
              {
                ref: function ref(el) {
                  return (_this6.input = el);
                },
                placeholder: this.props.placeholder
                autocomplete: this.props.autocomplete // add this
              },
```

could you help when you have time? many thanks

vik7607:
I want DayPickerInput should be like text field not inputwithBox  How can i achieve input with text field and drop down for DayPickerInput

PR #518, opened 2017-10-15
Replace moment specifier with format/parse functions

TrySound:
This PR will remove moment dependency in favour of parse and format helpers which can be handled with library chosen in a team.


gpbl:
Thanks for this PR! 

I was thinking to provide a default format, e.g. `format` could return the day as `YYYY-MM-DD`, and `parse` would convert it to a `Date` object. `String(new Date())` is a bit too generic as it returns also the time.

I was reasoning about writing a small utility that could recognize `D`, `DD`, `M, `MM`, `YY` and `YYYY` as format strings, so that `format` would accept also strings like `DD-MM-YY` and `parse` would convert them accordingly. The idea is to cover the most common cases without writing too much code.

gpbl:
@TrySound also I believe `parse` and `format` should receive the `locale` as second argument, which may come from the `dayPickerProps`.

```jsx
const parse = (string, locale) => { return 'parsedDate' };
const format = (date, locale) => { return 'string' };

<DayPickerInput 
  dayPickerProps={{ locale: 'ru' }} 
  format={format} 
  parse={parse}
/>
```

This would help developers to implement their own i18n.

TrySound:
Maybe better to pass date objects directly and use internal format and parser for simple specifer as defaults.  So you don't need to expose these defaults to let users understand date.

gpbl:
> Maybe better to pass date objects directly and use internal format and parser for simple specifer as defaults. So you don't need to expose these defaults to let users understand date.

Not sure I follow – what do you mean?

TrySound:
`<DateInput value={new Date(1992, 12, 20)} onChange={d => console.log(d.toISOString())} />`
This will allow to work with date directly without parsing received date

gpbl:
OK, this is another change we can consider, and it makes sense – as it is somewhat related to https://github.com/gpbl/react-day-picker/issues/508 ("Move input related props to `inputProps`").

Until now, the `value` prop means "value" as in input fields, i.e. is expected to be a string. Once we close #508, we can do this. However, I'd open another PR for this change – on the master branch where moment is still a required dependency – to be sure we update the tests just for that change.

gpbl:
I also think we should open a `v7` branch and merge these PR there. 

There are still [some fixes to do](https://github.com/gpbl/react-day-picker/milestone/11) on v6 that I would like to publish before other breaking changes.

PeterKottas:
Hi guys, where do we stand with this? Do we have a date in mind for v7? I rly think getting rid of this moment dep is high priority

gpbl:
@PeterKottas me too :) The problem is that there's a pending minor release, I'm working on it right now. 

I'd like to release this minor version where you can _optionally opt out_ from moment. I.e., it defaults to moment, but if you specify `format` and `parse` attribute, it will not. I'm not sure if it is possible, however.

PeterKottas:
Oh that sounds cool. You should definitely be able to do that. Just require the module, then check if it's define before you use it and fallback to some detetime format. I think you already have logic for that there. It would however need to remain a peer dependency which scares some people away. But it it's only for now, I'd go for it.

PR #523, opened 2017-10-20
Updated display style property for .DayPicker-Footer

saenglert:
The today button isn't showing up for me and I was able to track the problem down to said line in the stylesheet. 

On another page: It would be useful to be able to switch between the today button behaving like a footer and a (working) caption.


gpbl:
Woah thanks!!

Issue #475, opened 2017-08-02
[IE/EDGE] Incorrect sequence of focusing elements


enheit:
Hello, I've found incorrect sequence of focused elements using `TAB` key.

Unfortunately, I could not open [CodeSandbox](https://codesandbox.io/) in IE, therefore I use Edge as alternative.

Look at CodesandBox [issue example](https://codesandbox.io/s/oQWmOBmPj)
**Note: Issue is reproducing only in IE & Edge browsers**

Browsers to reproduce issue:
- IE 11 (Version - 11.1480.14393.0, Update version: 11.0.44) (Win 10)
- Edge (Version - 38.14393.1066.0) (Win 10)

Actual result:
*On `TAB` key press, focus moves to first input instead of next. (Jumps between first and second ).*
- The `DayPickerInput` [v.6.1.0](https://github.com/gpbl/react-day-picker/releases/tag/v6.1.0) lose sequence of focusable elements
![incorrect-gif](https://user-images.githubusercontent.com/8645216/28879966-3d6647b2-77ac-11e7-8113-0e259a66b4cf.gif)

Expected result:
- The next `DayPickerInput` should be focused (GIF animation from Crhome browser)
![correct-focusing](https://user-images.githubusercontent.com/8645216/28879968-3f787ea8-77ac-11e7-91f7-21bfc3871c56.gif)

gpbl:
Thanks for reporting this and the clear description! 

Any chance you can take a look to the source? (I'm traveling) To hide and display the overlay on focus/blur events, we are using a timeout. My guess is that it does not work correctly in IE.

gpbl:
This should have been fixed with https://github.com/gpbl/react-day-picker/pull/525 released in v7, please report back if not - thanks!

PR #525, opened 2017-10-23
Fixed lost input focus

martinmosko:
### Issue:
1. Open Firefox (its not an issue in chrome) https://codesandbox.io/s/XDAE3x0W8
2. Click on day picker input (it shows calendar)
3. Click on navigation bar element (next, prev or nav title elements)
4. Main day picker input loses own focus and calendar stays open
5. Now when you click anywhere outside of calendar. Calendar is still open but is should close.

### Solution:
Looks like the problem in FF is just timing of `this.input.focus(); ` call. This fix is solving the issue.


gpbl:
Ah! Awesome. Maybe this fixes also https://github.com/gpbl/react-day-picker/issues/475

Issue #524, opened 2017-10-20
Add Modifiers type export back


azhangstrata:
DayModifiers was imported into DayPicker.d.ts but not exported. Can we add it back as it was in 6.1.x versions?

gpbl:
Yes! Would you open a PR? 🙏

PR #526, opened 2017-10-23
Add modifiers back to type

azhangstrata:
According to this issue https://github.com/gpbl/react-day-picker/issues/524, adding the DayModifiers and Modifiers back to type file.


gpbl:
thanks a lot!

Issue #521, opened 2017-10-17
Don't override DayPickerInput selectedDays prop 


jm-maniego:
Useful for date range picker

Example: https://codesandbox.io/s/1v2r6p75j4

PR #531, opened 2017-10-26
Don't overrides DayPickerInput selectedDays prop

hydrognomik:
This PR for #521. `dayPickerProps` moved after `selectedDays` so DayPickerInput's `selectedDays` won't be overridden.

### DIFF
--- a/test/daypickerinput/rendering.js
+++ b/test/daypickerinput/rendering.js
@@ -174,21 +174,5 @@ describe('DayPickerInput', () => {
       expect(spy).toHaveBeenCalledTimes(3);
       spy.mockRestore();
     });
-
-    it('should use `selectedDays` from DayPickerInput props', () => {
-      const wrapper = mount(
-        <DayPickerInput
-          selectedDays={new Date(2017, 1, 8)}
-          dayPickerProps={{
-            month: new Date(2017, 1),
-            selectedDays: new Date(2017, 1, 9),
-          }}
-        />
-      );
-      wrapper.instance().showDayPicker();
-      wrapper.update();
-      expect(wrapper.find('.DayPicker-Day--selected')).toHaveLength(1);
-      expect(wrapper.find('.DayPicker-Day--selected').at(0)).toHaveText('8');
-    });
   });
 });
--- a/test/daypickerinput/rendering.js
+++ b/test/daypickerinput/rendering.js
@@ -174,5 +174,21 @@ describe('DayPickerInput', () => {
       expect(spy).toHaveBeenCalledTimes(3);
       spy.mockRestore();
     });
+
+    it('should use `selectedDays` from DayPickerInput props', () => {
+      const wrapper = mount(
+        <DayPickerInput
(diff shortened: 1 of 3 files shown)


gpbl:
Thanks for the PR! Wondering why tests are failing here 🤔

hydrognomik:
It's because DayPickerInput has `HIDE_TIMEOUT`, but test checks `--selected` modifier immediately. Should I add `setTimeout` [here](https://github.com/gpbl/react-day-picker/blob/master/test/daypickerinput/events.js#L263)?

jm-maniego:
I suggest supporting the old version, and just override it when it's passed

something like this maybe..
```javascript
var selectedDay = this.props.dayPickerProps.selectedDays;
if (!selectedDay && this.state.value) {
      const m = moment(this.state.value, this.props.format, true);
      if (m.isValid()) {
        selectedDay = m.toDate();
      }
    }
```

test/daypickerinput/rendering.js
```diff
@@ -174,5 +174,21 @@ describe('DayPickerInput', () => {
       expect(spy).toHaveBeenCalledTimes(3);
       spy.mockRestore();
     });
+
+    it('should use `selectedDays` from DayPickerInput props', () => {
+      const wrapper = mount(
+        <DayPickerInput
+          selectedDays={new Date(2017, 1, 8)}
```
gpbl: I'm not sure why you are introducing a new `selectedDays` prop here. The idea of this update is that it doesn't matter which day is the user clicking in the overlay: the selected days will be those coming from the `dayPickerProps` if present.

test/daypickerinput/rendering.js
```diff
@@ -174,5 +174,21 @@ describe('DayPickerInput', () => {
...
+          dayPickerProps={{
+            month: new Date(2017, 1),
+            selectedDays: new Date(2017, 1, 9),
+          }}
+        />
+      );
+      wrapper.instance().showDayPicker();
+      wrapper.update();
+      expect(wrapper.find('.DayPicker-Day--selected')).toHaveLength(1);
+      expect(wrapper.find('.DayPicker-Day--selected').at(0)).toHaveText('8');
```
gpbl: So this test is OK, it should have picked the `selectedDays` from `dayPickerProps`. No need to set `selectedDays` in the `<DayPickerInput />`.

The tests we are missing is: 
* if `selectedDays` is passed to `dayPickerProps`, clicking a day in the day picker will still use the days from `dayPickerProps`
* if `selectedDays` is passed to `dayPickerProps`, typing a new valid day in the day picker will still use the days from `dayPickerProps`

hydrognomik: Oh, ok, I'll remove that extra prop, but what to do with behavior that checks by [this](https://github.com/gpbl/react-day-picker/blob/master/test/daypickerinput/events.js#L248) test?

gpbl:
Awesome thanks a lot !

PR #532, opened 2017-10-29
install guide added to repo readme

abiduzz420:
I was looking to troubleshoot the application on local dev server and I was not able to locate install guide in the main README file. Only when I opened docs directory, I was able to locate install guide. I suggest that install guide should be placed in main README file of the repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,17 @@
   </a>
 </p>
 
+## Run the website locally
+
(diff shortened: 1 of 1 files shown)


gpbl:
Thanks for the suggestion, however the docs aren’t meant for this. If someone would like to run the docs locally he/she should refer to the docs directory README. Running the docs locally is for contributors who wants to update them.

PR #497, opened 2017-09-25
Added `renderWeek` prop to render a custom element for week box

jenshandersson:
I needed a way to render a custom element for week boxes on the left.
Following the `renderDay` implementation I created a `renderWeek` prop.
It defaults to the current behavior, rendering the weekNumber.

PR #546, opened 2017-11-14
add month parameter to renderWeek

smesgr:
In addition to the pull request #497 I require an additional parameter to distinguish weeks outside of the current month (enableOutsideDays & fixedWeeks set to true) and handle those weeks differently.


gpbl:
Good one, thanks for the idea!

smesgr:
thanks for the fast merge. Can't wait for the next release :)

Issue #557, opened 2017-11-22
Days of the week header has odd underline of dots. Shouldn't be any underline.


bytor99999:
We are upgrading, let's just say from this project almost 3 years ago. The latest version seems to have dots underneath the Header of the calendar drop down. Right under "Su" "Mo" etc, and I need to get rid of them, but I did not find anything in the CSS for underlines.

hydrognomik:
Weekdays placed inside `abbr` tag, so they have that dots. Try [this](https://stackoverflow.com/questions/26276556/remove-underlining-caused-by-abbr-tag).

gpbl:
We got this question often, maybe we should remove the underline from the CSS style...

bytor99999:
The link did not solve our problem. I added this to our datePicker.jsx file which is used throughout our application that encapsulate react-day-picker component and I still get the dotted lines under the day of the week.

`  _getStyles() {
    return {
      containerStyle: {
        display: 'table'
      },
      textField: {
        background: 'transparent',
        minWidth: '80px',
        position: 'relative'
      },
      wrapStyle: {
        position: 'relative'
      },
      overlay: {
        width: '100%',
        height: '100%',
        position: 'fixed',
        top: 0,
        right: 0,
        zIndex: 10
      },
      overlay2: {
        width: '100%',
        position: 'fixed',
        top: 0,
        right: 0,
        zIndex: 996,
        height: this.state.topMenuHeight + 10
      },
      abbr: {
        borderBottom: "none !important",
        cursor: "inherit !important",
        textDecoration: "none !important"
      }
    };
  }`

gpbl:
@bytor99999 that underline is a default of the browser’s CSS for the abbr tag, which is used by this component for the months abbreviations You have to override it in your style.

PR #560, opened 2017-11-23
Removes underline from weekdays

hydrognomik:
PR for #557. It removes dotted underline from weekdays.


gpbl:
Yuppie thanks!

gpbl:
Published in v7! Thanks

PR #575, opened 2017-11-29
Typescript type definiton doesn't provide the third parameter in renderWeek. Could you please add this. Thanks

smesgr:
React Type definition doesn't provide the third parameter added in #546. Please add this parameter. Thanks


gpbl:
Great find! thanks!

gpbl:
Published as v7.0.3

smesgr:
thanks!

PR #576, opened 2017-12-01
include renderMonths typing

mantaskaveckas:
Hey, we're using your awesome library along with TypeScript, but we were missing a typing for renderMonths function/method/whatever. Looking forward to your response. :v:

### DIFF
--- a/types/DayPicker.d.ts
+++ b/types/DayPicker.d.ts
@@ -10,7 +10,7 @@ import {
   DayPickerProps,
 } from './props';
 
-export default class DayPicker extends React.Component<DayPickerProps, any> {
+export default class DayPicker<T> extends React.Component<DayPickerProps, any> {
   static dayPicker: HTMLDivElement;
   static VERSION: string;
   static LocaleUtils: LocaleUtils;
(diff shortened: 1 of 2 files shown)


gpbl:
Hi @mantaskaveckas! Actually that method should not be public – what do you need it for?

mantaskaveckas:
@gpbl we're trying to put months into our own container (we're integrating slider into datepicker), and this works fine using renderMonths. As far as I know, there is no other public way to provide custom container?

gpbl:
Well that’s an interesting use case... do you have a screenshot of what your are trying to do? Maybe there’s another way to do this.

I’m not keen to expose that method since it uses some internals, and if we expose it we need to count it for breaking changes which are an hassle when releasing updates 😊

mantaskaveckas:
Here's what we're trying to do:

Use Case

gpbl:
Understood! This is a kind of animation I'd like to have as well 😍 

What if we could support it natively? Did you already implemented it? Is it requiring some external dependency, like react-transition-group?

mantaskaveckas:
@gpbl yep, we've already implemented it. We're using a small gallery slider (siema) port for React (react-siema) to achieve this. I think we could support it natively as well. :)

PR #585, opened 2017-12-08
Add possibility to disable roll to prev/next month if outside day is clicked

smesgr:
Some of my users have difficulties to understand the outside click feature.
If only one month is visible it seems to be ok, but if more then one month is rendered they find it confusing.

Thus I would like to disable this feature. 
With new property name I'm unhappy, but can't come up with a more precise phrase.

### DIFF
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -66,7 +66,7 @@ export default class DayPicker extends Component {
 
     // Customization
     showOutsideDays: PropTypes.bool,
-    rollOnOutsideDays: PropTypes.bool,
+    enableOutsideDaysClick: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


gpbl:
Thanks for you contribution! Yep we could call that prop `enableOutsideDaysClick`.

smesgr:
yeah suggestion sounds better then mine.  :+1:

ebrillhart:
Chiming in to say I'd love to have this merged - have a project where we'd like users to be able to see/select full weeks even if they fall between months, but don't want the month to change on click. Is there a timeline for merging and releasing this functionality?

gpbl:

PR #588, opened 2017-12-17
Add support for clicking TAB on input focus

### LINKED ISSUE #587: Clicking tab on input hides the overlay
When focusing on input, clicking tab should, IMO ,move focus to the "prev month" button, to provide better accessibility.

matanwerbner:
This fixes Issue #587 (that i opened), adds a keydown handler (tab doesn't get detected on keyup)
which sets clickedInside to true. then, onBlur, overlay remains open, which solves the problem.

### DIFF
--- a/src/DayPickerInput.js
+++ b/src/DayPickerInput.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 
 import DayPicker from './DayPicker';
 import { getModifiersForDay } from './ModifiersUtils';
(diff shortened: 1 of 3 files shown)


gpbl:
Awesome, thanks @matanwerbner !

gpbl:
After trying it doesn't work as expected (the field stays focused even pressing `TAB`). I think the `TAB` behavior is correct the way is implemented.

Issue #584, opened 2017-12-07
React dayPickerInput formatting to MM/DD/YYYY prevents entering the date in the input text box 


kblodget:
I have the following code sand box which is a fork of the range with 2 inputs example., https://codesandbox.io/s/w6vnlox29l I have it working the way I want but I want to input dates as `MM/DD/YYYY` instead of the default format. When I add the following attributes:

```jsx
      <DayPickerInput
            value={from}
            format="M/D/YYYY"
            formatDate={formatDate}
            parseDate={parseDate} 
            placeholder="MM/DD/YYYY"
            onDayChange={this.handleFromChange}
            dayPickerProps={{
              selectedDays: [from, { from, to }],
              modifiers,
              disabledDays: {
                after: new Date(),
              },
              numberOfMonths: 2,
              toMonth: new Date(),
            }}
          />
```

I can no longer correctly enter the date in the input field the date. When I remove the format, formatDate and parseDate attributes the date picker input functions the way I expected. I have read the documentation but think I may be missing something simple. How can I use the input text box and date picker with dates formatted as `MM/DD/YYYY`?

SpaceK33z:
I am not 100% sure this is the same bug, but I have this reproducable example: https://codesandbox.io/s/23lvr5861r
When you focus the `input` field and press backspace somewhere, for example, after the "2017" part, you'll see that "2017" is changed to "0201". This bug was introduced in 7.0.0; in 6.2.1 this worked like expected.

gpbl:
@SpaceK33z yes same bug, thanks for reporting your example!

jbarco:
I've noticed that the `formatDate` provided is using `moment` in "forgiving mode" (https://momentjs.com/guides/#/parsing/forgiving-mode/). Every time the input changes, the value is parsed and `moment` tries its best to make a valid date out of it, overriding the input value:

Say we have `10/10/2012` and the format is `DD/MM/YYYY`. If I delete the last digit on the year, the input value would be `10/10/ 201`. `Moment`  parses the string and, thinking the date has `201` as a year, it tries to create a date `DD/MM/YYYY` by formatting the year `201` with 4 digits: `0201`. The entire value is parsed as  `10/10/0201` and displayed in the input.

In strict mode, moment would return `undefined` as the string `10/10/201` does not match  `DD/MM/YYYY`.

One solution to fix the problem could be to set up moment to run on strict mode, but it would be hard to distinguish between empty values in the input and errors (wrong format).

[...]

PR #590, opened 2017-12-20
Set moment in strict mode

jbarco:
This will cause moment to return `undefined` when parsing a date that doesn't match the format.

 `parseDate` value will not override the input value if it doesn't match the format provided.

For: #584


gpbl:
Thanks a lot @jbarco !

Issue #579, opened 2017-12-03
Switch between years and months on DatePickerInput


bartpeeters:
When modifying the "Switch between years and months" example to use a DatePickerInput with dayPickerProps, the month and year selects do not properly open.

If you set showOverlay to true the selects do open, but as soon as you select a day and then reopen it the selects do no longer open. Is this a css issue?

Codesandbox: https://codesandbox.io/s/l5991q6x9z

gpbl:
🤔 It should because this line:

https://github.com/gpbl/react-day-picker/blob/master/src/DayPickerInput.js#L299-L302

I have no idea how to prevent this. Maybe intercepting some of the blur/click events? Need some time to investigate it.

bartpeeters:
Thanks for the quick response.

The blur event you linked is indeed causing this behavior, commenting out the lines you linked fixes the problem.

I don't really get why these lines are necessary, it seems to work nicer when they are commented out: no flicker every time you click inside the DayPicker.

Another way to fix this may be using event.relatedTarget to check if one of the custom elements originated the blur event. But I don't know how you could differentiate the custom elements from all the other elements in the DayPicker.

gpbl:
> I don't really get why these lines are necessary, it seems to work nicer when they are commented out: no flicker every time you click inside the DayPicker.

Those line were added by @martinmosko in https://github.com/gpbl/react-day-picker/pull/525/files
Maybe we could find another solution for the bug. 

I'm also not happy with the flickering :)

martinmosko:
@gpbl @bartpeeters 

well use case is part of PR description. I tried to solve this issue so probably right solution will be to work with other timeout usages to set them correctly together (some how). 

**Issue:**

Open Firefox (its not an issue in chrome) https://codesandbox.io/s/XDAE3x0W8
Click on day picker input (it shows calendar)
Click on navigation bar element (next, prev or nav title elements)
Main day picker input loses own focus and calendar stays open
Now when you click anywhere outside of calendar. Calendar is still open but is should close.
**Solution:**

Looks like the problem in FF is just timing of this.input.focus(); call. This fix is solving the issue.

gpbl:
thanks for dropping by @martinmosko ! no idea yet how we could fix that firefox bug in another way. 

Sure i hope to not introduce other timeouts since they make things unpredictable 😌

bartpeeters:
[...]

PR #598, opened 2018-01-03
Alternate solution for timeout blur

bartpeeters:
Follow up from https://github.com/gpbl/react-day-picker/issues/579#issuecomment-353798385, should replace #525.

---

Issue:
1. Open Firefox (its not an issue in chrome) https://codesandbox.io/s/XDAE3x0W8
2. Click on day picker input (it shows calendar)
3. Click on navigation bar element (next, prev or nav title elements)
4. Main day picker input loses own focus and calendar stays open
5. Now when you click anywhere outside of calendar. Calendar is still open but is should close.

The fix using a setTimeout onBlur to call this.input.focus() caused new issues: 
- A flicker in Chrome everytime you click inside the daypicker
- When using a custom captionElement containing a <select> it would not open because the timeout focus cancelled it

This alternate solution is to remove the timeout and instead add an onBlur on the overlay doing the same thing as the onBlur of the input.


gpbl:
Thanks @bartpeeters for your PR. I want to try it first, sure we have to remove that timeout thing since it doesn't well. 

I don't understand why the new `span` element should fire the `onBlur` event [here](https://github.com/gpbl/react-day-picker/pull/598/files#diff-e2a925e665d836a09575f79095191746R415). It shouldn't ever get the focus. Have you find the reason behind it?

bartpeeters:
@gpbl tabIndex on the daypicker ([here](https://github.com/gpbl/react-day-picker/blob/8c69e20419bfc3a601de7ee267c2436707f3ed62/src/DayPicker.js#L553)) makes it so the span is focusable.

src/DayPickerInput.js
```diff
@@ -417,22 +412,24 @@ export default class DayPickerInput extends React.Component {
...
-        classNames={classNames}
-        month={this.state.month}
-        selectedDay={selectedDay}
-        input={this.input}
-      >
-        <DayPicker
-          ref={el => (this.daypicker = el)}
-          onTodayButtonClick={onTodayButtonClick}
-          {...dayPickerProps}
+      <span onBlur={this.handleInputBlur}>
```
gpbl: My suspect is that this line does nothing. We should add an unit test to it.

test/daypickerinput/events.js
```diff
@@ -52,18 +52,6 @@ describe('DayPickerInput', () => {
         wrapper.find('input').simulate('blur');
         expect(onBlur).toHaveBeenCalledTimes(1);
       });
-      it('should focus the input if blur after clicking the overlay', done => {
```
gpbl: This test must stay here :) Why does your change make it fail?

bartpeeters:
@gpbl Thanks for the feedback.
I implemented another solution, using the onFocus event of the overlay to always make sure the DaypickerInput input field is focused. 

But do we want this? What if we have a custom overlay that includes an input? Now this input is unusable because it can't be focused. We can check if the nodeName of the focus event target is an input, but I don't really know if that is the way to go.

I also replaced the clickInside with the timeout by using a ref on the Overlay and checking if the relatedTarget of blur event is inside the overlay. I'm not sure that this is better than using a timeout.

About the bubbling of the onFocus and onBlur on the span which contains the overlay, the React Synthetic onFocus and onBlur events do bubble, unlike the normal blur and focus events.

gpbl:
@bartpeeters thanks for the updates. I've not checked your changes yet, however I'm proposing a different solution:

- reset this line to the original behavior, which was working well except on Firefox:

```diff
if (this.clickedInside) { 
-    this.blurTimeout = setTimeout(() => this.input.focus(), 0); 
+   this.input.focus(); 
}
```

- on Firefox, we just disable this to resolve https://github.com/gpbl/react-day-picker/issues/579:

```diff
- if (this.clickedInside) { 
+ if (!isFirefox && this.clickedInside) { 
```

- for the case when the calendar has other focusable elements inside, such as selects or input fields (as in https://github.com/gpbl/react-day-picker/issues/599), we allow the implementer to disable this focus with a new DayPickerInput prop: `keepFocus` that can be set to `false`.

bartpeeters:
@gpbl I like the idea of a new prop `keepFocus`.

The solution you propose has the same effect as these changes except they dont use `isFirefox`, so maybe you can take a look at the code. I'm also up to add the `keepFocus` prop and some tests if you approve of these changes.

mattfysh:
@gpbl how can I help push this forward? would really appreciate being able to set `keepFocus` to `false`

bartpeeters:
@mattfysh I just added the `keepFocus` prop and added api documentation for it. I think now we just need to wait till @gpbl has some time to review these changes.

docs/src/pages/api/DayPickerInput.js
```diff
@@ -246,6 +247,16 @@ function MyDayPickerInput(props) {
       <p>
         The value of the <code>input</code> field.
       </p>
+      <h3>
+        <Anchor id="keepFocus" />
+        keeFocus <code>boolean = true</code>
```
ghaynes1292: keepFocus

bartpeeters: Thanks!

rdiwelm:
Hey guys! How soon can we expect a new version release with the new `keepFocus` prop?

gpbl:
@rdiwelm soon ! I have been pretty busy the last month, hoping to catch up all this weekend 🙏🏽 sorry for the lack of communication.

roginfarrer:
I also have a feature depending on this release!

rdiwelm:
Hey guys! Sorry to be asking again. Is there any ETA on this? We're kinda blocked too. :/

kradical:
@roginfarrer @rdiwelm I was blocked by this too so we forked this dependency until this change gets merged. I understand that is not always a viable option.

gpbl:

rdiwelm:
Awesome! Thanks a lot, guys! :)

PR #612, opened 2018-01-22
Use dayPickerProps controlled month before value.

kradical:
Hello @gpbl !

First thank you for your time commitment to this project, it is very useful!

This change is a re-ordering of how the current month is calculated from props. I believe it should use the controlled month from `dayPickerProps.month` before it uses the current `value`.

My specific use case is a 4-month calendar using the DayPickerInput date range selection with a from/to input. In my case I want the first month in the 4 month range to show the start of the range but it will default to using the current value. Which on the `to` DayPickerInput will possibly be a different month.

Otherwise I don't think there is a way to use `DayPickerInput` as a controlled component (w/ redux-form for us) and control the month through props.

Let me know if there is anything else I can provide to make this easier!

### DIFF
--- a/src/DayPickerInput.js
+++ b/src/DayPickerInput.js
@@ -179,9 +179,6 @@ export default class DayPickerInput extends React.Component {
       } else {
         day = props.parseDate(props.value, format, dayPickerProps.locale);
       }
-      if (day) {
-        month = day;
-      }
     }
 
     // Use DayPicker's controlled month. Then try the current `value`. Finally default to today.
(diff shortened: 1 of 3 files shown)


kradical:
I'm looking into this a bit more and it looks like I may be able to "control" the component through `props.dayPickerProps.selectedDays` and forgo the use of the `value` prop I will update this comment and close the PR/Issue if I can make it work for my use case!

gpbl:
Thanks @kradical I'd review this soon :) Just catching up with the older stuff .

gpbl:

kradical:
awesome! Thank you!

PR #616, opened 2018-01-27
Get rid of the redundant nullability modifiers of the isDayAfter and …

vkrol:
…isDayBefore method parameters

These methods cannot be called with undefined or null because of https://github.com/gpbl/react-day-picker/blob/master/src/DateUtils.js#L9


gpbl:
Thanks!

Issue #586, opened 2017-12-08
Typescript import DayPickerInput


kirjai:
Thanks for providing good typescript types for this library ❤️

However, when trying to import `DayPickerInput` in a Typescript project, this style of import fails
`import DayPickerInput from 'react-day-picker/DayPickerInput'` with the following error message:

```
Try `npm install @types/react-day-picker/DayPickerInput` if it exists or add a new declaration (.d.ts) file containing `declare module 'react-day-picker/DayPickerInput';`
```

simply importing from `react-day-picker` doesn't work as well, as `DayPicker` component is exported as default.

Am i missing something obvious here, or there's no way of importing `DayPickerInput` in a typescript project?

gpbl:
I have no idea! The typescript stuff is from contributors: maybe you could take a look to the source to see how to fix this?

reinartg:
I am using this component in a typescript project. Have you tried importing it the typescript way:
`import * as DayPickerInput from 'react-day-picker/DayPickerInput';`

kirjai:
@reinartg yeah, i get an error saying that `react-day-picker/DayPickerInput` has no declaration.

```
Could not find a declaration file for module 'react-day-picker/DayPickerInput'.
```

Does that approach work for you?

96467840:
i had same problem. i solved it 
1. import DayPickerInput from 'react-day-picker/DayPickerInput';
and we see compilation error

[at-loader] Checking finished with 1 errors
[at-loader] ./ClientApp/components/fields/InputText.tsx:12:28 
    TS7016: Could not find a declaration file for module 'react-day-picker/DayPickerInput'. 'D:/work/my/github/webclient/node_modules/react-day-picker/DayPickerInput.js' implicitly has an 'any' type.
  Try `npm install @types/react-day-picker/DayPickerInput` if it exists or add a new declaration (.d.ts) file containing `declare module 'react-day-picker/DayPickerInput';` 

2. make 1.d.ts file in project and put into 
declare module 'react-day-picker/DayPickerInput';

enjoy

96467840:
d.ts file
`declare module 'react-day-picker/DayPickerInput' {

    import * as React from 'react';
    import { ClassNames } from 'node_modules/react-day-picker/types/common';
    import { DayPickerInputProps } from 'node_modules/react-day-picker/types/props';
    import DayPicker from 'node_modules/react-day-picker/types/DayPicker';

[...]

PR #618, opened 2018-01-31
Export more types from index.d.ts

### LINKED ISSUE #617: Export DayPickerProps (and more?) from @types/react-day-picker
It appears `DayPickerProps` is not explicitly exported from `@types/react-day-picker`. You can reach into it by importing a submodule:

```tsx
import { DayPickerProps } from "@types/react-day-picker/props"
```

But this is hacky and produces a lint error when `tslint`'s [`no-submodule-imports`](https://palantir.github.io/tslint/rules/no-submodule-imports/) rule is enabled. The workaround is to disable the rule for the import statement in question:

```tsx
// tslint:disable-next-line:no-submodule-imports
import { DayPickerProps } from "@types/react-day-picker/props";
```

This is a **`P3`** since we have a workaround, but it would be nice to have all types exported from `index.d.ts`, rather than exporting only `DayPicker`.

**EDIT:**
I may have the package name wrong (I haven't tried this myself - I'm following along in another PR as a reviewer). Are the types just exported from `react-day-picker` directly, or are they published under `@types/react-day-picker`? Either way, my request stands!

adidahiya:
Fixes #617

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -2,3 +2,6 @@
 
 import DayPicker from './DayPicker';
 export default DayPicker;
+
+export * from './common';
+export * from './props';


types/index.d.ts
```diff
@@ -2,3 +2,6 @@
 
 import DayPicker from './DayPicker';
 export default DayPicker;
+
```
i8wu: if you add `export * from './DayPickerInput';` it'll also solve https://github.com/gpbl/react-day-picker/issues/586

adidahiya: that is semantically incorrect. DayPickerInput is not exported from the root of the package: https://github.com/gpbl/react-day-picker/blob/master/DayPicker.js

i8wu: Hm, what would be the correct way then? Kinda new to typescript.

adidahiya: since the JS export is from a submodule, you can import in typescript like this:

```ts
import { DayPickerInput } from "react-day-picker/DayPickerInput";
```

that's why [`types/DayPickerInput.d.ts`](https://github.com/gpbl/react-day-picker/blob/master/types/DayPickerInput.d.ts) exists.

i8wu: Importing like that doesn't work, my compiler complains about no declaration file found.

adidahiya: hm, ok, can you file an issue for that? I might be able to look into it later

i8wu: Should I open up a new issue or is https://github.com/gpbl/react-day-picker/issues/586 good enough? Thank you!

alexis-thual:
Can someone merge this please? :)

gpbl:
Hey thanks for the PR! Are we safe with this? I'm not a TS developer, people were fighting for this kind of issues the last time... should I go and merge?

alexis-thual:
I am not a TS expert but this PR seems quite fair to me; you can go ahead and merge I guess :)

adidahiya:
This is safe; I contributed to the original types on DefinitelyTyped and in this repo. #586 is a separate issue and can be addressed in another PR.

gpbl:
Thanks @adidahiya as always!

gpbl:

theigor:
Hi guys - just got latest and getting this error:
```
ERROR in [at-loader] ./node_modules/react-day-picker/types/props.d.ts:177:27
    TS7006: Parameter 'DayModifiers' implicitly has an 'any' type.
```

where line 177 is
```ts
  onDayChange?(day: Date, DayModifiers): void;
```

so this seems like a legitimate issue. Did you mean to do 
```ts
  onDayChange?(day: Date, modifiers:DayModifiers): void;
```

theigor:
By the way, you can avoid things like this in the future by switching your tsconfig to 
```json
{
  "compilerOptions": {
    "noImplicitAny": true,
    ...
  },
  ...
}
```

Issue #634, opened 2018-02-23
Remove object-assign dependency


gpbl:
See https://github.com/gpbl/react-day-picker/pull/632

TrySound:
@gpbl Do you want me to send different PR?

gpbl:
Awesome :)

PR #635, opened 2018-02-23
Remove object-assign package in favor of spread operator

TrySound:
Ref #634 #632


gpbl:

PR #636, opened 2018-02-23
Add cross-env for save environment variables

TrySound:
This is better if somebody will contribute from windows


gpbl:

PR #637, opened 2018-02-24
Replace env with production to fix umd crashing in runtime

TrySound:
Ref #632 

Currently umd bundle just doesn't work.

- added `Input` component everywhere
- replace `process.env.NODE_ENV` with production to prevent crashing in umd bundle because prop types package does not exists.
- simplified all entry points in favor of src/index.js

I think this can be published as minor.

before
  daypicker.js 102.1K
  daypicker.min.js 39.27K

after
  daypicker.js 95.68K
  daypicker.min.js 39.27K

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,6 +1,8 @@
 /* eslint global-require: 0, import/no-extraneous-dependencies: 0 */
 /* eslint-env node */
 
+const webpack = require('webpack');
+
 const isProduction = process.env.NODE_ENV === 'production';
 
 module.exports = {
@@ -29,6 +31,11 @@ module.exports = {
       },
     ],
   },
+  plugins: [
+    new webpack.DefinePlugin({
+      'process.env.NODE_ENV': JSON.stringify('production')
+    })
+  ],
   externals: {
     react: {
       root: 'React',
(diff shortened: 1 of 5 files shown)


TrySound:
@gpbl Are you okay with this PR?

TrySound:
Ping? Currently umd is broken.

gpbl:
Yeah this should work now. I guess all these complications were made (by me) because I didn't want to include the moment.js package.

webpack.config.js
```diff
@@ -29,6 +31,11 @@ module.exports = {
       },
     ],
   },
+  plugins: [
+    new webpack.DefinePlugin({
+      'process.env.NODE_ENV': JSON.stringify('production')
```
gpbl: I think in development this plugins should not be there. Dunno gotta merge it and see how it works.

TrySound: UMD should not require build step, since it can be used in browsers. So `NODE_ENV` should be either `production` or `development`. Currently webpack adds process polyfill which sets `NODE_ENV` as `undefined`, so it's not `production`. But you excluded `prop-types` from umd and `prop-types` definitions are fails in runtime.

gpbl:

PR #641, opened 2018-02-25
Prefer react namespace over default export

TrySound:
One day react will land esm and all stuff will updated. We can do it now and don't worry later.


TrySound:
Also rollup will work only with one type of import.

TrySound:
What do you think about this one?

Issue #631, opened 2018-02-22
Possible to expose prop to delay date parsing/formatting to onBlur?


roginfarrer:
Sandbox: https://codesandbox.io/s/zkn6oqnq4l

I'm having some trouble using `dateFns` for the `formatDate` and `parseDate`. (I recognize that this library only supports Moment currently, but bare with me.) When using a custom `format` and `dateFns.format` or `dateFns.parse`, the input is impossible to edit. See a similar issue with Moment here: #584. As far as I know, `dateFns` does not offer a "forgiving" mode.

I think this issue could be addressed by exposing an option to postpone the execution of the parse/format functions until the `onBlur` event. That way the user has free range to edit the input, then the string can be formatted once the user is finished.

I'm open to other solutions, but I thought this could work.

gpbl:
This is a legit improvement (beside the issues you are having which I'll read later :))

roginfarrer:
@gpbl I put together a fork that accomplishes what I laid out above. Doesn't flip it as a prop yet, but I wanted to get some thoughts on it.

[See diff](https://github.com/gpbl/react-day-picker/compare/master...roginfarrer:master#diff-e2a925e665d836a09575f79095191746)

The big change is that `inputValue` is maintained in state separate from value. There is now also a check to make sure that the user's input is parseable. If it is, great, proceed and update the state. If not, revert to the previous `inputValue`.

gpbl:
I don't think all these is needed:

> I'm having some trouble using dateFns for the formatDate and parseDate.

See this example: 
https://codesandbox.io/s/r44xwwj35o

Hope it helps!

PR #643, opened 2018-02-28
`parseOnBlur` to delay parsing to input's onBlur event

roginfarrer:
This addresses the issue raised in #631.

When this prop is true, it escapes the parsing and input changes in the `onChange` handler, and instead executes them inside `onBlur`.

To do this, I had to create a new state value `inputValue` to be handles separately from the prop `value`, since they will not always be in sync. This shouldn't affect the component whether the `parseOnBlur` prop is true or false.


roginfarrer:
Hey @gpbl, do you have any thoughts on this PR?

gpbl:
i need to check it out yet :) 

At first I’d prefer to not add new props. Is this a side  case that can be worked around without changing the source?

💭 (this input component is taking up too much time. I’d prefer to put it in another package. I never liked input with date pickers eheh)

roginfarrer:
Hi @gpbl, have you looked at the suggested changes? I'll update the PR soon to be up to date with master.

I'm not sure how this could be accomplished without adding a prop, since the current implementation of `this.props.parseDate` is pretty integral to the functioning of the Input. Currently, the `parseDate` is executed in the input's `onChange` handler. This arose as an issue #584, the strict mode of Moment's parse function threw errors with custom `this.props.format`. The solution there was to change Moment's `parse` function to 'forgiving' mode.

However, there is no such mode for `date-fns` or likely any other date library's parse function. I'm proposing here the option to therefore execute the `this.props.parseDate` function in the input's `onBlur` handler. This will allow the user to write whatever they want in the input, then the `parse`, then the parse function can handle whether it's a Date-able string, or revert it to the original value.

Unless we permanently move the `parseDate` function to the `onBlur` handler, I'm not sure how we could accomplish this without a prop?

gpbl:
https://github.com/gpbl/react-day-picker/issues/631#issuecomment-386888145

PR #646, opened 2018-03-05
Fix error with focusing outside nodes 

### LINKED ISSUE #630: Cannot read property 'focus' of undefined in DayPicker
Steps to reproduce:

1. Open https://codesandbox.io/s/23vz0vrn7r
2. DayPicker will display December, 2018
3. Click on January 5, 2019 (with is also displayed because of showOutsideDays is true).
4. Press left cursor key.

Result: An arrow is thrown: "Cannot read property 'focus' of undefined"
Expected result: the focus moves from January 5 to January 4.

daypicker-focus

gpbl:
Fix #630


gpbl:

PR #647, opened 2018-03-05
Call `onDayChange` with `undefined` when day is not valid

### LINKED ISSUE #614: onDayChange not always called from DayPickerInput
The onDayChange function is not called where you manually enter **invalid** text into the text box of the DayPickerInput.

See https://codesandbox.io/s/v6m53q7o47

Select a date - alert is shown

Enter X as the date picker text - no alert

gpbl:
Fix #614


gpbl:

PR #653, opened 2018-03-06
Fixed 'any' type on onDayChange second argument

bungu:
Fixed error TS7006: Parameter 'DayModifiers' implicitly has an 'any' type.


gpbl:
thanks!!

PR #658, opened 2018-03-12
Revert "Revert "Fix umd build""

TrySound:
Reverts gpbl/react-day-picker#656

### DIFF
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -574,7 +574,3 @@ export default class DayPicker extends Component {
     );
   }
 }
-
-DayPicker.DateUtils = DateUtils;
-DayPicker.LocaleUtils = LocaleUtils;
-DayPicker.ModifiersUtils = ModifiersUtils;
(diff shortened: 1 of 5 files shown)


gpbl:
Still broken, going to find the fix. You can see `DateUtils` is not exported when running [the website](https://github.com/gpbl/react-day-picker/blob/master/docs/README.md) and trying to select a range 
 in this example: http://localhost:8000/examples/selected-range.

src/DayPicker.js
```diff
@@ -574,7 +574,3 @@ export default class DayPicker extends Component {
     );
   }
 }
-
-DayPicker.DateUtils = DateUtils;
-DayPicker.LocaleUtils = LocaleUtils;
-DayPicker.ModifiersUtils = ModifiersUtils;
```
gpbl: Removing these lines seems to cause the issue. Are those that were causing the UMD bug?

TrySound: I'm digging right now.

gpbl:
Nope it doesn't work either. Look I'm spending too much of my free time for fixing a problem nobody complains, let's keep v7 as it is.

TrySound:
Okay. v8 shouldn't have this problem. Its exports are transparent.

gpbl:
Let's hope that 🙏🏽

PR #660, opened 2018-03-13
Make sure utilities are exported again

gpbl:
Ref #658 cc @TrySound


TrySound:
Do you expect more patch/minor releases? We probably can get rid from all legacy entry points and default exports.

gpbl:
Yeah better put order in the exports in v8 (rollup).

TrySound:
@gpbl I mean can we start working on v8?

gpbl:
I think this PR can go in a patch. But does it fix the UMD export?

TrySound:
Okay. Let's merge this then.

Issue #633, opened 2018-02-23
Remove bower support


gpbl:
* [ ] remove bower.json

PR #661, opened 2018-03-13
Remove bower

TrySound:
Close https://github.com/gpbl/react-day-picker/issues/633


gpbl:
Why this? 🤔

TrySound:
I thought you wanted to remove bower. So lib won't be necessary to commit. Loading umd can be solved with http://unpkg.com/

gpbl:
So this is related to bower or to roll up? 

Sorry it’s a while I don’t change the config for a lib so i may miss something here. How do people access to the compiled js module and CSS?

TrySound:
```js
import 'react-day-picker/lib/styles.css'
```

TrySound:
Compiled code and styled It will be published on npm. Bundlers will use them.

gpbl:
Maybe i should check this PR once I’m pn my computer but isn’t this removing the lib folder? 🤨

gpbl:
Oh so OK when building we still create `lib`! 👍🏽

gpbl:
Could we please remove also bower.json?

gpbl:
Thanks!!

gpbl:
(@TrySound I'm merging the v8 related PRs to the `v8` branch.)

PR #662, opened 2018-03-13
Convert to named exports and export DayPickerInput

TrySound:
- removed root entry points
- exported `DayPickerInput` from `src/index.js`
- replaced aliases with `react-day-picker` link in docs


docs/package.json
```diff
@@ -23,7 +23,7 @@
     "moment": "2.21.0",
     "prismjs": "1.11.0",
     "prop-types": "15.6.1",
-    "react-day-picker": "../",
+    "react-day-picker": "link:../",
```
gpbl: ooh !! ✅

PR #678, opened 2018-03-23
Combine UMD configs with rollup multiconfig and add size snapshot

TrySound:
This will reduce cost of running node and loading rollup twice.


gpbl:
Some more explanation please? :)

TrySound:
I have a plan to distribute also cjs and esm bundles instead of many files. This will also reduce the cost of bundling vendors.

TrySound:
Added [rollup-plugin-size-snapshot](https://github.com/TrySound/rollup-plugin-size-snapshot). I created this plugin for improving bundles of frontend projects.

TrySound:
Ant thoughts?

gpbl:
@TrySound thanks this work has been landed into #944. Would love a review there, if you still have time :)

Issue #342, opened 2017-05-12
Adopt date-fns to format and manipulate dates


gpbl:
When I first wrote this component, moment.js was the only good library to work with `Date` objects – but being its size so big I preferred to keep this package without depending on it (one of the reasons I made this component was that the available date picker were always requiring moment.js). But working with Date objects in javascript is not easy at all, and maintaining my own Dates manipulation library requires some effort.

Now we have [date-fns](https://date-fns.org): it is modular like lodash so I expect it that adopting it would not increase so much the size of this module. Adopting date-fns would have many benefits:

- finally fix long standing [issues with timezones](https://github.com/gpbl/react-day-picker/issues/130) for some edge cases (need https://github.com/date-fns/date-fns/issues/376 closed)
- we won't need to change the date to 12AM to workaround such issues (need https://github.com/date-fns/date-fns/issues/376 closed)
- react-day-picker would have built-in support for i18n (wait for date-fns v2)
- we can remove the required moment.js support from the input component
- we would push js community away from moment.js and adopt this better date library

Similar discussion: https://github.com/airbnb/react-dates/issues/208

r-tock:
@gpbl can you make this time library agnositic, we migrated away from moment to js-joda.github.io and it has splendid support. If somehow react-day-picker can be modular to support any time library, we would appreciate it.

gpbl:
@r-tock thanks for your feedback.

date-fns seems modular enough so it shouldn't add much weight to the final bundle. I'll do my tests to see how it works first, but it seems i can just replace the original utils with some of theirs.

gpbl:
So far there are two missing pieces from date-fns: 

- the ability to parse a date from any format (https://github.com/date-fns/date-fns/issues/219). Not sure if the `parse` function in v2 will work with different locales as well.
- getting the first day of the week for different locales. [this issue](https://github.com/date-fns/date-fns/issues/493) seems it's not possible yet.

leshakoss:
@gpbl hello, thank you for trying date-fns :-)

- `parse` would work with different locales — as long as the locale is adjusted for `parse` and v2.
- locale-specific first day of the week will be possible in v2: https://github.com/date-fns/date-fns/blob/v2/src/startOfWeek/test.js#L20-L32

We will release an alpha very soon. Stable version will be released after all locales are converted for v2.

Is there something I can do to help your library adopt date-fns?

gpbl:
[...]

PR #687, opened 2018-04-05
Add example using date-fns to format and parse dates

brycehill:
I added a small example for how to use the new `formatDate` and `parseDate` props with [date-fns](https://date-fns.org/).

I think this might be able to close https://github.com/gpbl/react-day-picker/issues/342.

### DIFF
--- a/docs/src/code-samples/examples/input-date-fns.js
+++ b/docs/src/code-samples/examples/input-date-fns.js
@@ -1,0 +1,19 @@
+import React from 'react';
+
+import DayPickerInput from 'react-day-picker/DayPickerInput';
+import 'react-day-picker/lib/style.css';
+
+import format from 'date-fns/format';
+import parse from 'date-fns/parse';
+
+export default function Example() {
+  const FORMAT = 'M/D/YYYY';
+  return (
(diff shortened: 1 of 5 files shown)


gpbl:
Awesome, thanks! Yes we can close that issue...

docs/src/code-samples/examples/input-date-fns.js
```diff
@@ -0,0 +1,19 @@
...
+import 'react-day-picker/lib/style.css';
+
+import format from 'date-fns/format';
+import parse from 'date-fns/parse';
+
+export default function Example() {
+  const FORMAT = 'M/D/YYYY';
+  return (
+    <DayPickerInput
+      formatDate={format}
```
gpbl: However, date-fns `format` and `parse`  does not follow the `parseDate` and `formatDate` function signature.

This should be the right way to use the `formatDate` attribute

```jsx
formatDate = { (day, format, locale) => format(day, format, { locale }) }
```

Too bad `formatDate` uses the locale as third argument! 🤦‍♂️  Well, we could update them in a next release...

PR #517, opened 2017-10-15
Fix modifiers when using classNames

gpbl:
Replace https://github.com/gpbl/react-day-picker/pull/504

Issue #689, opened 2018-04-07
disabledDays not working using classNames


ivansurov:
Hi,
DayPicker DisabledDays not worked with ClassNames.
Example: https://codesandbox.io/s/vnyz3mo0x5

hannescalibrate:
Created PR: https://github.com/gpbl/react-day-picker/pull/741

PR #690, opened 2018-04-07
Fix modifiers when using classNames

ivansurov:
If using classNames modifiers is not correct working.

For example contains key "src-css-daypicker__disabled--21D2D" key instead of "disabled"

Issue: https://github.com/gpbl/react-day-picker/issues/689

### DIFF
--- a/src/DayPickerInput.js
+++ b/src/DayPickerInput.js
@@ -374,6 +374,16 @@ export default class DayPickerInput extends React.Component {
       formatDate,
       format,
     } = this.props;
+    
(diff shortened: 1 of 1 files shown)


gpbl:
Thanks! This should close #517 as well. Could you please lint the source and add a test? http://react-day-picker.js.org/contributing thanks!

ethanve:
Looks like this was never merged and I'm still seeing issues with CSS modules. Would you like me to open a PR for a fix? @gpbl

Issue #686, opened 2018-04-05
Add example for selecting week on day click


beckerei:
I'd like to highlight the week of the currently selected day.
As far as I can tell from the docs there is no good way to do so.

I'd happy to contribute by either adding this as an option (+ highlight month), or adding week and month selection to date ranges.

What do you think?

gpbl:
What do you mean for "good way"? I'm sure, it's possible to do this using a modifier matching the days being in the same week of the selected day. Maybe it is a bit verbose, an example would be nice. 

Would you try it in [codesandbox](https://codesandbox.io/s/XDAE3x0W8)? We can discuss here its implementation

beckerei:
Sorry for not giving feedback (, I went on spontaneous vacation for a couple of days).

I hacked a small example, which is currently in a closed source project, but here is how it looks like. I used the modifiers as you suggested.

![highlightweek](https://user-images.githubusercontent.com/3099581/38834878-02f0db18-41ca-11e8-910c-12ca9b479a02.gif)

But I realized that the "SELECTING DAYS"-Section does not use a special interface designed for the purpose of selecting ranges.
My initial thougt was to add a prop like `selectWeek={true|false}` and the DatePicker will automatically highlight the week (taking `firstDayOfWeek` into account). Not sure what `onChange` could return. Either an array of all days within the selected week, or an object with from and to as keys.

I'll try to extract what I've done at work into a little codesandbox.

gpbl:
It makes sense actually to add more ways to interact with weeks. Clearer APIs and i guess better performant... How do you think it should work?

gpbl:
@beckerei I had a second thought about this: if we add support for week events, like `onWeekClick`, we should likely add all the other events and maybe add modifiers for weeks. So, at least this is not a feature requested by many, I'd leave the API as they are.

beckerei:
I‘ll implement what we need and discuss with our ux/ui Designers.
Maybe that‘s something that would fit in the example section 👍

beckerei:
Here is an implementation similar to the one we now use at work.

https://codesandbox.io/s/8y510myyyl

It only takes the selected day as reference, and won't give back the entire selected week. But this should be easy to implement at least when working with a layer on top of react-day-picker.

If this is of interest I can extend the example to work with multiple selected Days, and can open a PR for the example section.

gpbl:
@beckerei it is interesting for sure, thanks for sharing. 

[...]

PR #719, opened 2018-05-12
Add example for selecting an entire week

beckerei:
I finally had the time to extract https://github.com/gpbl/react-day-picker/issues/686 into the example section.

The implementation slightly differs from the codebox example I previously build.

week


gpbl:
Awesome, thanks a lot!
It's published: http://react-day-picker.js.org/examples/selected-week 

I've updated a bit the code https://github.com/gpbl/react-day-picker/commit/56ea646d69a1ca97933bc5a1471f9dbc440dab1f

Issue #725, opened 2018-05-24
Calendar day hover effect is flickering


sv3k:
When you move the mouse cursor in gaps between days — hover background effect is flickering:

![Demo](https://image.ibb.co/dQYUvo/demo.gif)

This issue is reproducible on any DayPicker with hover effects, including [demo website](http://react-day-picker.js.org).

Browser: Chrome 66
OS: maxOS 10.13

gpbl:
Hi, check out this: https://github.com/gpbl/react-day-picker/issues/610

Issue #610, opened 2018-01-16
Flickering when moving mouse diagonally between days


antiwebbite:
Hey Guys,

I was wondering if anyone has an idea on how to fix the issue where the mouse and hover states flicker when moving the mouse diagonally between days?

Video: https://screencast-o-matic.com/watch/cFVDqOoDev

I've noticed that even if you force the cursor to pointer on DayPicker-Body, the issue still occurs when the cursor crosses those diagonal edges.

I've repro'd on the official examples as well.

Mac OS X 10.13.2
Chrome 63.0.3239.132

Thanks!

gpbl:
Did you make it? Maybe are the rounded corners? I'm closing this if you find the cause feel free to share thanks!

antiwebbite:
You are correct that changing the border-radius on the hover state from 50% to 0 makes the issue not occur.

I don't know how to fix it for use with a circular hover state, but for anyone where it doesn't matter, changing the border-radius to 3px on this element fixed it for me:

```
.DayPicker:not(.DayPicker--interactionDisabled) .DayPicker-Day:not(.DayPicker-Day--disabled):not(.DayPicker-Day--selected):not(.DayPicker-Day--outside):hover {
    border-radius: 3px;
}
```

sv3k:
Maybe it's worth to fix this issue in default react-day-picker styles.

gpbl:
@sv3k yeah would you mind to open a PR (provided that it works) thanks!

PR #726, opened 2018-05-24
Update default css styles to avoid flickering

sv3k:
Fix for #610, #725 — previously mouse cursor could flicker because of changing `border-radius` property on hover.


gpbl:
Awesome! thanks!

PR #730, opened 2018-05-30
Speed improvements in DayPicker if month is switched

smesgr:
Hi,

some of my users where unsatisifed with the speed of the day picker input. Especially with switching month in the overlay. I did some digging and could see the calendar is re-rendered twice. For some users with older systems and an IE11 this can cause sluggish behavior of up to 1,2seconds (~400ms per render)

The provided pull request try to avoid two setStates which causes the re-render.
The first part in (now) componentWillReceiveProps shouldn't have any negative impact. Removing the first re-render

The second part is a bit tricky. The problem is day picker input and day picker both have a month state. While in most cases the values are the same, there are some corner cases. Afaik the day picker is directly controlled by the parent component if month prop is given. Thus storing the value to state can be skipped. The parent - in this case the day picker input class - is responsible to pass new values via props. I tested several scenarios and it worked fine for me in both controlled and uncontrolled mode, but of course this is a API change if somebody passed a month prop, but don't do anything onMonthChange-callback. This part removes the second re-render.

The speed of the sluggish IE11 was improved from 1.2s to 400ms. Improvements are for all browsers, but afaik not noticeable outside the IE-world. (~10-70ms) 

Best Regards


gpbl:
Hey thanks for looking into this. If the problems is a useless re-render, shouldn't we make `shouldComponentUpdate` return `false` instead?

smesgr:
Afaik this could be done, but I'm afraid of side effects. The situation of the shouldComponentUpdate.

calls | prev | next 
Variant A | state day X, props day X | state day Y, props day X
Variant B | state day Y, props day X | state day Y, props day Y

I'm unsure if any of those variants should yield false in all cases available

gpbl:
@smesgr I'm short on time to investigate this 😒. My main perplexity in your changes is using the deprecated `componentWillReceiveProps`.

smesgr:
We have used the previous pull request for ~1 year now without complaints. :)

Because I would like to update the component to latest version. I have updated the Pull Request using this pattern:
https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization

BTW: I think i fixed an theoretically issue in getCurrentMonthFromProps along the way. If else-if-path was taken "this.props" was used instead of given props parameter.

gpbl:
Hey @smesgr thank for the follow up. I've been rewriting the component for the upcoming release (see #942) so this doesn't apply anymore. I'll be happy to have a feedback about performances if you will upgrade to v8, once its out. Thanks!

PR #753, opened 2018-07-11
Set style value in functional way in Day component

JuhQ:
Noticed that style setting wasn't as functional as it could've been. Now it should be a bit more pure and functional, with no side-effects.


gpbl:
Thanks, but I find the original way more readable!

JuhQ:
Roger that

gpbl:
Anyway thanks for the idea and for looking into these details 🔬

PR #754, opened 2018-07-11
Fix navbar eslint issue

JuhQ:
eslint was giving this issue:

```
8:3  error  defaultProps should be placed after propTypes  react/sort-comp
```

This pr should fix that


gpbl:
Thanks! I've noticed this in other project as well... did they update the rule in a minor version? 🤔

PR #762, opened 2018-07-29
Add top-level type definition for DayPickerInput

kryops:
This allows importing from `'react-day-picker/DayPickerInput'` from TypeScript.
Fixes https://github.com/gpbl/react-day-picker/issues/586

I chose to keep the existing type definitions for `DayPickerInput` in the `types` folder and only reference them from the new top-level definition file. Alternatively, the existing definition file could be moved up to the top level as well, but this would be a breaking change for TypeScript consumers.

The new file also takes care of exporting the `DayPickerInput` correctly as default export (see https://github.com/gpbl/react-day-picker/issues/586#issuecomment-372713868)

### DIFF
--- a/DayPickerInput.d.ts
+++ b/DayPickerInput.d.ts
@@ -1,0 +1,2 @@
+import { DayPickerInput } from './types/DayPickerInput'
+export default DayPickerInput


nucleartux:
@gpbl please merge, without this PR I can't use this library on my current project :(

kryops:
@nucleartux  you should be able to add the typings manually as a workaround.

1) Add a file containing the type definition to a folder that you use for types (e.g. `src/@types`):

[src/@types/react-day-picker.d.ts]
```ts
declare module 'react-day-picker/DayPickerInput' {
    import { DayPickerInput } from 'react-day-picker/types/DayPickerInput'
    export default DayPickerInput
}
```

2) Add the directory to `typeRoots` in your tsconfig.json:
```
"typeRoots": ["node_modules/@types", "src/@types"],
```

nnic:
This file does not seem to be present in the the npm module - does it need to be explicitly added to the package.json?

kryops:
@nnic sorry my bad :disappointed: 
I created #784 to add it.

PR #769, opened 2018-08-10
fix typo in 'monday'

jsjoeio:
Hmm... I read this:

>* 🙏🏽 Please **DO NOT** build files or update the docs in your pull request – as this may cause git conflicts. Thanks!

How else should I propose a typo fix for the docs? I'll add this here and wait to hear from you :)


gpbl:
Thanks! Well that disclaimer is for people changing the API, I'll clarify this a bit :)

PR #774, opened 2018-08-20
Add onClose event handler for DayPickerInput overlay

jwm0:
Adds a prop for DayPickerInput that fires once `hideDayPicker()` is called.

This is a crucial handler for validation purposes - especially the `touched` attribute which detects whether user focused the input field. As of now the only way to set it was to use onBlur method (via inputProps) and it would fire each time the overlay is clicked - like when clicking between the months which in turn would toggle the validator prematurely. This approach makes it easier to detect when a user actually focused the input and didn't select a date.

I had to update enzyme to run the tests successfully as the function prop is invoked in setState callback (fixed in 3.4.3).


src/DayPickerInput.js
```diff
@@ -115,6 +115,7 @@ export default class DayPickerInput extends React.Component {
     }),
 
     onDayChange: PropTypes.func,
+    onClose: PropTypes.func,
```
gpbl: We should really call it `onDayPickerHide` since "hide" is the verb we use to hide the overlay and we have a method called `hideDayPicker`.

src/DayPickerInput.js
```diff
@@ -312,7 +313,9 @@ export default class DayPickerInput extends React.Component {
     if (this.state.showOverlay === false) {
       return;
     }
-    this.setState({ showOverlay: false });
+    this.setState({ showOverlay: false }, () => {
+      if (this.props.onClose) this.props.onClose();
```
gpbl: I wonder if this is the place where we should call the event handler. I believe we should put it in a `componentDidUpdate`:

```jsx
componentDidUpdate(prevProps, prevState) {
    if (this.props.onDayPickerHide && this.state.showOverlay && !prevState.showOverlay) {
        this.props.onDayPickerHide();
    }
}
```

What do you think?

jwm0: Not sure about this, it seems to overlook a simple solution. We want it to fire just after the DayPicker is hid and that's when the callback is fired. While it would also do that in `componentDidUpdate` from a performance point of view it means that every time the component is updated it would check the if statement - which I do not think is necessary. 
What are your thoughts?

gpbl: I share with you the concerns about performance (which are more ideal, though). The reason I was thinking about this change, is that we must be sure the overlay is _always_ hidden only by calling `hideDayPicker` method, and not just by setting the state (which won't call the event handler). So in conclusion, we can go with this. We should just take extra care and hide the overlay only by calling that method.

PR #775, opened 2018-08-22
Add undefined and null as allowed modifier types

Lavoaster:
Examples like: https://react-day-picker.js.org/examples/selected-range-enter/ show that modifiers are in fact supported as undefined or nullable. They do not work in typescript without some fair amount of work to ensure the value isn't null, when in fact the library does handle those types.


gpbl:
Thanks! We are planning a release in the next days, so stay tuned.

Lavoaster:
Ah excellent news :)

PR #777, opened 2018-08-25
Change formatting to local em-sizes

signalwerk:
Problem: I think the component-styling looks great. But if I need to place it smaller or bigger in my design I have to change all the default stylings.

Suggested solution: if the styling is on the root-level set by `rem` (line: 5) and in the component by `em` I just need to change `.DayPicker {  font-size: 2rem; }` and the whole calendar will scale in proportion to it.

Compatibility: This change should not break any existing styling. People who did manual styling will not be affected, since they reset the default styling. And people who made no changes will not notice, since it's visually equal to the existing codebase.

Sidenote: on [Line 88](https://github.com/gpbl/react-day-picker/blob/master/src/style.css#L88) there was already a `em` format introduced.

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -2,12 +2,13 @@
 
 .DayPicker {
   display: inline-block;
+  font-size: 1rem;
 }
 
 .DayPicker-wrapper {
   position: relative;
   user-select: none;
-  padding-bottom: 1rem;
+  padding-bottom: 1em;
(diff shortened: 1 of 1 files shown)


gpbl:
Good idea, thanks! I think we should switch completely to `em` (so, even the root element) in the next major release – what do you think?

Issue #789, opened 2018-09-10
Re-enable handleDayClick for disabled dates


dewey92:
Hi, thanks for this great library beforehand, I really appreciate your work!

However, there's an issue I'd like to address to you regarding this commit (https://github.com/gpbl/react-day-picker/commit/808da17299653462b4f65b89a40d43f16279cc39#diff-e15d85e29b8c7310cbf7799c62521619) which basically ignores `this.props.onDayClick` when the selected date is disabled. In the application I'm developing, I really want to show a message explaining **why** a particular date is disabled (i.e we don't deliver the product on Sunday, etc) so the end-user can clearly know the reason.

Previously this worked:
```ts
onDayClick = (date: Date, modifiers: DayModifiers) => {
  if (modifiers.disabled) {
    // If user clicks Sunday,
    this.setState({ errMessage: 'No Sunday delivery' });
    // If user clicks 25 Dec
    this.setState({ errMessage: 'No Chrismast delivery' });
    // ...etc

    return;
  }

  this.setState({ errMessage: '' });
};
```

But now it doesn't. So, is there any other way to achieve this?

gpbl:
Yes you are right! That shouldn't have been merged, my bad. Could you revert it? I will do it in the next days otherwise. Thanks!

dewey92:
@gpbl should we close this issue?

gpbl:
Sure! FIx published in `v7.2.4`.

PR #791, opened 2018-09-12
Revert disabled check for handleDayClick

dewey92:
action taken from this issue https://github.com/gpbl/react-day-picker/issues/789 😉


gpbl:
Thanks! Patch will be published in few hours!

Issue #759, opened 2018-07-19
DayPickerInput documentation about className is incorrect


mnemanja:
This is what the documentation states:
![screen shot 2018-07-19 at 12 55 13](https://user-images.githubusercontent.com/1423218/42938603-1ac76ffe-8b53-11e8-8983-38ac7d424e67.png)
This refers to the React.PropTypes definition within the component.

However, the definition of the className in the `.ts` file speaks differently:
![screen shot 2018-07-19 at 12 57 26](https://user-images.githubusercontent.com/1423218/42938678-63570b58-8b53-11e8-8d75-663e2e8c3f05.png)

Another problem is that all of the properties listed under the `ClassNames` interface are mandatory.

Also, when one uses the `classNames={{ container: "someClassName" }}` it automatically removes the `overlayWrapper` class... What's up with that?

RDP: v7.1.0
Typescript v2.9.2

gpbl:
Maybe we shouldn't import `ClassNames` here:

https://github.com/gpbl/react-day-picker/blob/master/types/DayPickerInput.d.ts#L4

and use a new `InputClassNames`. Could you open a PR and add it? Thanks.

> Also, when one uses the classNames={{ container: "someClassName" }} it automatically removes the overlayWrapper class... What's up with that?

Please provide an example con CodeSandbox thanks!

PR #796, opened 2018-09-14
Fix DayPickerInput classNames TS definitions

davidspiess:
This pull request fixes the `classNames` property for the `DayPickerInput` component.
Discussed in #759.


davidspiess:
Any updates here?

davidspiess:
@gpbl Another option would be removing the definition files from the library so we TS users can constantly update them on definitely typed.

PR #805, opened 2018-09-27
[Types] renderDay function, modifiers argument

srigaux:
The second argument of function `renderDay(date, modifiers)` should be of type `DayModifiers` instead of type `Modifiers`.


gpbl:
Thanks a lot!

Issue #130, opened 2016-02-18
UTC mode


jonathanong:
i store my dates in GMT. my dates do not care about timestamps. i noticed that when using this library, dates are always shown based on the current timezone. this is annoying because from PST, when i select a date, the day before is actually selected in the calendar.

is there a way we can set the mode to GMT? my current work around is to create new date objects based on the current locale. works, but i'm not a fan of it.

thanks

gpbl:
Hi @johnrhampton

Could you provide an example of this behavior and what do you expect instead? Thanks!

johnrhampton:
Hi @gpbl.. I think you may have tagged me on accident.. I do love this library and would be more than happy to contribute. great stuff!

gpbl:
thanks @johnrhampton, I'm so sorry I meant to answer to @jonathanong :smile: i clicked on the wrong user!

johnrhampton:
no worries @gpbl!

janmyler:
Hi @gpbl, I think I know what @jonathanong thinks, an example using moment:
1. you have a date-time value stored as UTC/GMT string: `"2016-02-22T00:00Z"`,
2. you are in New York (GMT-5 offset).

```
const date = '2016-02-22T00:00Z';
let m = moment(date);
let mu = moment.utc(date);

m.format('YYYY-MM-DD HH:mm');
// output: "2016-02-21 19:00"
mu.format('YYYY-MM-DD HH:mm');
// output: "2016-02-22 00:00"
```

When you create an instance of `Date` object using UTC/GMT format, the value is shifted to the local timezone to represent a correct local time, which causes problems when you compare dates in modifiers (e.g. `isSameDay` for selected day).

I hope my explanation is understandable at least a little bit. :smile:

jonathanong:
oh sorry. i forgot to follow up. yes, what @janmyler says :) i always handle all my dates with `moment.utc()`.

another option is just to pass a date format like `YYYY-MM-DD` to date picker instead of a `Date` object so that it never picked up the locale and because the time isn't relevant.

gpbl:
Thanks for the feedback! Sorry for the delay, as I'm currently on vacation.

So I take it, the issue is related to the [DateUtils](https://github.com/gpbl/react-day-picker/blob/master/src/DateUtils.js#L29)'s `isSameDay` function - right? It would help to have some code to test this behavior, as I don't understand the role of moment.js here. Or just wait some more days, when back I'll try to reproduce the issue by myself :)

[...]

PR #810, opened 2018-10-03
Fix timezone dependant keys

tobilen:
relates to #130

Currently, the keys used for the `Week`components in the `Month` component are different depending on the timezone of the engine that compiles the code.

This causes a problem when creating snapshots of the day-picker, since they may be compared between machines with different timezones.

to demonstrate this behavior, i've added a snapshot test that just snapshots the `key` prop of the first week of a given month. the snapshot was recorded using UTC time,and it will fail on any machine running the test suite with a different timezone.

the actual implementation consists of stripping the timezone away from the date before converting it to a timestamp and using it as a key.

this method could be implemented for any other dates as well. i am using it together with https://www.npmjs.com/package/jest-date-serializer to have full utc snapshotting.

### DIFF
--- a/src/Month.js
+++ b/src/Month.js
@@ -180,9 +180,12 @@ export default class Month extends Component {
             if (showWeekNumbers) {
               weekNumber = DateUtils.getWeekNumber(week[6]);
             }
+            const dateWithoutOffset = new Date(
+              week[0].getTime() - week[0].getTimezoneOffset() * 60000
+            );
             return (
               <div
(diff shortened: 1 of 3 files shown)


tobilen:
@gpbl anything i can do to help getting this merged?

tobilen:
@gpbl is this something i can expect to get merged? otherwise i would need to fork

tobilen:
@gpbl can you give an update on this?

Issue #815, opened 2018-10-05
DayPickerInput prefers typed value over controlled value 


MitchRivet:
reproduction here: https://codesandbox.io/s/0yjvkzm3vn

I know this is a very specific example, but it has come up in my particular use case. In the example above, I have a DayPickerInput with a controlled value. When I type an invalid date into the input, and the value is updated (from outside of the component) the typed valid is not cleared. This is the opposite behavior when the user changes the date from the input's overlay. 

This happens because of the order of variables in the OR condition here: https://github.com/gpbl/react-day-picker/blob/master/src/DayPickerInput.js#L569

It seems like switching the order of those two variables would be the desired approach (this.state.value is the source of truth for the input's value) 

![sharedstateexample](https://user-images.githubusercontent.com/13023846/46565101-0dc41e00-c8c1-11e8-9a4f-424b52755849.gif)

walterdn:
I believe this is the same issue as:
https://github.com/gpbl/react-day-picker/issues/804

ivan-nejezchleb:
We have the same issue in our application.

In version 7.0.7 it works fine. 
But in 7.2 there is no way how to reset the text. It is caused by merge of https://github.com/gpbl/react-day-picker/pull/778

I propose different solution, but I am not sure, if this doesn't break some other use case
- clear inner state 'typedValue' to undefined in componentDidUpdate, when props.value has changed (or at least when has changed to some valid date).
https://github.com/gpbl/react-day-picker/blob/master/src/DayPickerInput.js#L189

Issue #804, opened 2018-09-26
Can't set empty value for input ( reproduction )


wenzoto:
There was a [similar error](https://github.com/gpbl/react-day-picker/issues/666), but could not reproduce it.

https://codesandbox.io/s/7m6q1y8xmj

### reproduction

<a href="https://gyazo.com/5bbb2bc21ffb6a2bede4045e91cf7ed3"><img src="https://i.gyazo.com/5bbb2bc21ffb6a2bede4045e91cf7ed3.gif" alt="Image from Gyazo" width="538"/></a>

cjnickel:
Ran into this issue as well. Looks like PR #816 should fix this assuming it gets merged.

Bereware:
This would be a temporary solution.
This error is the one that shows when you try to modify the state:
 _**Warning: A component is changing a controlled input of type undefined to be uncontrolled**_

I have solved it this way, it is not recommended to use refs but ... xD 
```js
<DayPickerInput
    value={from}
    ref={(el) => (this.from = el)}
.../>
<DayPickerInput
    value={to}
    ref={(el) => (this.to = el)}
.../>
handleResetClick = () => {
        // this fixed the encapsulated input uncontrolled bug
        this.from.setState({ value: '', typedValue: '' });
        this.to.setState({ value: '', typedValue: '' });
    };
```
I hope help you

serban-condrea:
In order to avoid refs, I used another little hack 😅

```
<DayPickerInput
    value={to || ' '}
.../>
```

gpbl:
This has been fixed in the latest releases - please upgrade thanks

owen-stellar:
Though I've upgraded to 7.4.8 I don't see this being resolved. When I clear out the value (in a blur function when the entry is deemed to be invalid) the typedValue remains the same. Is there something more that needs to be done besides simply upgrading to 7.4.8?

tanelih:
The latest release still exhibits this problem when clearing a value that is not a valid Date.

I've played around changing the following line to also set the typed value to the given value & the problem seems to go away.

https://github.com/gpbl/react-day-picker/blob/v7/src/DayPickerInput.js#L198

```
if (isDate(value)) {
  newState.value = formatDate(value, format, dayPickerProps.locale);
} else {
  newState.value = value;
  newState.typedValue = value;
}
```

However I'm not a 100% certain this is the best solution, but it seems to work for my use case, where dates are stored in the form state as plain strings.

PR #816, opened 2018-10-06
When DayPicker input state is controlled, correctly set the input text on 'value' prop updates

MitchRivet:
Reported as an issue with explanation here: https://github.com/gpbl/react-day-picker/issues/815

Essentially, my issue is that when DayPickerInput is controlled (i.e. takes a 'value' prop) and an invalid date is entered into the input box, an update to the value prop does not clear/update the text within the input (even though the overlay correctly updates the state)

The main issue in this case happens here: https://github.com/gpbl/react-day-picker/blob/master/src/DayPickerInput.js#L576

In this case, this.state.typedValue is defined and remains inside the input, even though this.state.value is the correct state of the input. 

The other changes address warnings that are shown in the console as a result. This message happens when you initialize a controlled input with an undefined value: 

screen shot 2018-10-05 at 5 13 31 pm

Also, I added a test to cover this scenario.

### DIFF
--- a/src/DayPickerInput.js
+++ b/src/DayPickerInput.js
@@ -486,10 +486,13 @@ export default class DayPickerInput extends React.Component {
       } else if (selectedDays) {
         selectedDays = null;
       }
+
       this.setState(
-        { value: '', typedValue: undefined, selectedDays },
+        { value: '', typedValue: '', selectedDays },
         this.hideAfterDayClick
       );
+
+
       if (onDayChange) {
         onDayChange(undefined, modifiers, this);
       }
@@ -566,7 +569,7 @@ export default class DayPickerInput extends React.Component {
           ref={el => (this.input = el)}
           placeholder={this.props.placeholder}
           {...inputProps}
-          value={this.state.typedValue || this.state.value}
+          value={this.state.value || this.state.typedValue}
           onChange={this.handleInputChange}
           onFocus={this.handleInputFocus}
           onBlur={this.handleInputBlur}
--- a/src/DayPickerInput.js
+++ b/src/DayPickerInput.js
@@ -126,6 +126,7 @@ export default class DayPickerInput extends React.Component {
   static defaultProps = {
     dayPickerProps: {},
     value: '',
+    typedValue: '',
     placeholder: 'YYYY-M-D',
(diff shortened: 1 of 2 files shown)


andrewsuzuki:
Looks solid to me, @gpbl can you review?

jordan-jarolim:
First of all, thank you for your work! Will this be merged any soon? :)

MitchRivet:
@jordan-jarolim thx, hopefully soon. haven't heard from any of the maintainers. going to reach out in the gitter and see if there planning on updating/releasing anytime soon

tanelih:
Related to #804 - with this PR how would one go about setting an empty value for the input (e.g. when resetting the form state). Since empty values (empty strings) are evaluated as "falsy", the value passed to the actual input element will always end up being `typedValue`. This leads to a situation where setting the value programmatically to an empty string will still leave the `typedValue` intact inside the component's state - thus rendering it instead of the "reset" value.

I did some digging around and found out that one could "reset" the `typedValue` in the `componentDidUpdate` function:

```javascript
componentDidUpdate (prevProps) {
  // ...
  if (value !== prevProps.value) {
    if (this.props.value === "") {
      newState.typedValue = ""
    }
    // ...
  }
  // ...
}
```

I'm not sure whether this is something to consider. I might also be using this component wrong. :smile:

aekt:
@tanelih I totally agree with you! It seems like no way to reset the input field now... Based on this PR, setting `value` state to `" "` can indeed empty the field, but it may not be very satisfying if one wants placeholder to be shown. So, we probably need some way to reset `typedValue`?

Wouldn't it be a good if we directly expose a method to reset this `typedValue` state? I think such way will be more flexible and explicit.

KumailP:
Isn't this a somewhat core functionality, to be able to reflect the state in the input on update? Either way, great work. Really hope this gets merged soon.

dmitrykrylov:
Hey! Is somebody going to merge this PR? DayPickerInput can't provide good UX now because it allows an invalid date string to be kept as an input value and there is no even way to control it :(

hovoodd:
@gpbl merge please 😢

mdeeter:
Can this get merged?

PR #828, opened 2018-11-13
Fix wrong type definition on LocaleUtils

lukyth:
Every `locale` has `en` as the default value so they're optional. Source: https://github.com/gpbl/react-day-picker/blob/master/src/addons/MomentLocaleUtils.js#L5


gpbl:
Awesome, thanks!

PR #831, opened 2018-11-21
Adds onDayPickerShow to complement onDayPickerHide

kylekthompson:
👋 Hello.

While working with `react-day-picker`, I ran into a situation where I wanted to know whether the day picker was open. To help with figuring that out, this PR exposes an `onDayPickerShow` callback to complement the `onDayPickerHide` callback that already exists.

Typically I'd instead opt for keeping the `showOverlay` state local to my own component and use this as a controlled component, but it doesn't look like that is possible (if we'd rather have the ability to do that, I'm willing to open a PR or modify this one to support that instead, but I think we'd still need this callback to know when to set that state to `true`)

### DIFF
--- a/src/DayPickerInput.js
+++ b/src/DayPickerInput.js
@@ -116,6 +116,7 @@ export default class DayPickerInput extends React.Component {
 
     onDayChange: PropTypes.func,
     onDayPickerHide: PropTypes.func,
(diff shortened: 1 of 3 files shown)


gpbl:
Good idea, thanks!

PR #832, opened 2018-11-28
Add style prop to DayPickerInput

kirillku:
I found it would be useful to have `style` prop passed down to container div of `DayPickerInput`. Sometimes it's much more handy to just apply some small inline styles instead of creating or modifying css classes.

```javascript
// for example if you want to make only this one specific `DatePickerInput` fluid
<DayPickerInput style={{ width: '100%' }} />
```

### DIFF
--- a/src/DayPickerInput.js
+++ b/src/DayPickerInput.js
@@ -108,6 +108,7 @@ export default class DayPickerInput extends React.Component {
     component: PropTypes.any,
     overlayComponent: PropTypes.any,
 
+    style: PropTypes.any,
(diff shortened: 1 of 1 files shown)


src/DayPickerInput.js
```diff
@@ -108,6 +108,7 @@ export default class DayPickerInput extends React.Component {
     component: PropTypes.any,
     overlayComponent: PropTypes.any,
 
+    style: PropTypes.any,
```
gpbl: Maybe:

```suggestion
    style: PropTypes.object,
```

PR #834, opened 2018-11-28
Add missing utils export in TS definitions

### LINKED ISSUE #812: LocaleUtils should be exported from types/index.d.ts
Previous versions of react-day-picker exported the LocaleUtils interface in the TypeScript type definitions. This was changed at some point, and LocaleUtils is no longer exported.

Without this exported type, it is still possible, but cumbersome, to obtain the type as follows:
```ts
import { DayPickerProps } from "react-day-picker";
type DayPickerLocaleUtils = Exclude<DayPickerProps["localeUtils"], undefined>;
```

It would be much more convenient if LocaleUtils was exported by "react-day-picker".

Similarly, the other interfaces defined in utils.d.ts should probably also be exported for convenience. This would be as simple as adding the following line to index.d.ts:
```ts
export * from './utils';
```

davidspiess:
At the moment it's not possible to import `LocaleUtils`, `DateUtils`, `ModifiersUtils` in Typescript projects, because the utils export is missing in the definition file.
Fixes #812;

image

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -5,3 +5,4 @@ export default DayPicker;
 
 export * from './common';
 export * from './props';
+export * from './utils';


DevanB:
Came here to find why these types weren't being exported. Happy to see a PR is opened. When can it be merged in? 😄

johnjesse:
Can you get to them temporarily by doing using the statics on DayPicker e.g. `typeof DatePicker.LocaleUtils`

mhafer:
@DevanB I used  DayPicker.DateUtils.addDayToRange(); not LocalUtils

PR #844, opened 2018-12-12
Mark handled keydown events as handled

johnjesse:
The Arrow keys are used by the browser to scroll, and `DayPicker` uses them to navigate months and years on the body of the `DayPicker`. Currently these events are not marked as handled so the browser does it's default action - which is scrolling when in a container that scrolls, and they bubble to anyone else listening higher up the tree. I've used the cancel event helper to stop that happening.

Here is a sandbox that highlights the issue https://codesandbox.io/s/1q60jqr9ml


gpbl:
Hey thank you – sorry for the late merge.

Issue #837, opened 2018-12-03
RTL keyboard handling issues


johnjesse:
Hello,

First, let me say that so far I have found this library to be great!

I have an issue when using the day picker in RTL. the keyboard handling is counter-intuitive. Moving the arrow keys left/right moves the focus of the date in the opposite direction.
See: https://codesandbox.io/s/l9wj0ow469 - use the code-pen to move around the day picker with the arrow keys and you'll see what I mean.

It'd be nice if Day picker used the dir prop and then decided what to do based on that, I think I could submit a PR for that if we agreed that should be the approach. Also, the docs for the dir prop appears to be non-existant - http://react-day-picker.js.org/api/DayPicker#dir

gpbl:
Fixed with #845

PR #845, opened 2018-12-12
Switch what left and right keys do in RTL

johnjesse:
Fix for #837:

Basically, it used the dir prop to decide how to handle the left and right arrows being pressed, switching what they do in RTL.

I've also added the missing `dir` prop to the `DayPickerProps` interface.

The docs for dir usage are missing right now... it would be good to add them, I'd be happy to do that as part of this PR - although the PR template warns against making any changes to the docs


gpbl:
Awesome, thank you – sorry for being so late on this!

PR #854, opened 2019-01-11
Add isDate and isSameMonth to type definition

rnons:
These two functions were added in v7.1.6, but type is missing


gpbl:
Thank you!

Issue #748, opened 2018-07-09
Stateless function as component prop triggers a warning


elamperti:
CodeSandbox demo: https://codesandbox.io/s/zqx94zxy0l

Following the [example in the documentation](http://react-day-picker.js.org/api/DayPickerInput#component), adding a `component` property raises an error in the console saying:

> Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail.
>
>Check the render method of `DayPickerInput`.
>    in component (created by DayPickerInput)
>    in div (created by DayPickerInput)
>    in DayPickerInput (created by Example)
>    in div (created by Example)
>    in Example

I've noticed that @gengue had the same issue as well some months ago ([Gitter message about it](https://gitter.im/gpbl/react-day-picker?at=5a1c3b278b3a9e2c0c2344c6))

Transforming the component passed to the picker to a class seems to fix this problem (see [this other example](https://codesandbox.io/s/vn5yn388q5)) although it doesn't look good and it's different from what the documentation suggests.

Could you provide a better solution or update the documentation to reflect this? Thanks in advance! :)

gpbl:
Hey thanks for reporting this! The problem with stateless function is that we can't know how to focus the input field, so we use a ref there. This won't give any error from the `v7.1.10` however the field won't be focused when the user chooses a day. We could add a condition to check if the component is stateless and skip the `ref` at all...

Sawtaytoes:
I'm also seeing the warning in `v7.1.10`.

Why not use `document.querySelector` instead of `ref` as an option?

It's not manipulating the DOM, just selecting a field that has the exact same name as you're passing to the `input` anyway. That way, I could give the component `selector="[name='someFieldName']"` or just `name="someFieldName"`, and it would do something as simple as:

```js
document
.querySelector('[name="someFieldName"]')
.focus()
```

Or is there some way I could just do this myself so long as the warning goes away?

gpbl:
> Why not use `document.querySelector` instead of ref as an option?

This can be a solution, but it is not easy as it seems – since it is possible to change the component of the input element, we really don't know what's inside there, which `name` it has (if it has one) etc.

A solution could be using `findDOMNode`, at 

```js
get input() {
   return React.findDOMNode(this).querySelector('input')
}
```

which will cover the most of the cases (but not the ones with multiple input fields in it).

[...]

PR #862, opened 2019-02-05
Accept Both Functions and Components for DayPickerInput

glockjt:
Problem:
React was throwing an error when a function was passed into the `component` property for the `DayPickerInput`
```js
<DayPickerInput component={props => <input {...props} />} />
```
This address #748 

Explanation:
The [docs example for component](http://react-day-picker.js.org/api/DayPickerInput) shows the above code. However it is using it and expecting a React Component, thus resulting in React throwing the following error:

image

Resolution:
Add a check to determine if the component is a function or not and if it is a function render the function by passing all `inputProps`.  If the component is a react component render as a component.


src/DayPickerInput.js
```diff
@@ -562,18 +562,33 @@ export default class DayPickerInput extends React.Component {
...
+            ...inputProps,
+            value: this.state.typedValue || this.state.value,
+            onChange: this.handleInputChange,
+            onFocus: this.handleInputFocus,
+            onBlur: this.handleInputBlur,
+            onKeyDown: this.handleInputKeyDown,
+            onKeyUp: this.handleInputKeyUp,
+            onClick: !inputProps.disabled ? this.handleInputClick : undefined,
+          })
+        )}
```
panczarny: I strongly recommend to extract common props to the const.
```
const props = {
            placeholder: this.props.placeholder,
            ...inputProps,
            value: this.state.typedValue || this.state.value,
            onChange: this.handleInputChange,
            onFocus: this.handleInputFocus,
            onBlur: this.handleInputBlur,
            onKeyDown: this.handleInputKeyDown,
            onKeyUp: this.handleInputKeyUp,
            onClick: !inputProps.disabled ? this.handleInputClick : undefined,
};
```
and then spread it to child.

src/DayPickerInput.js
```diff
@@ -562,18 +562,33 @@ export default class DayPickerInput extends React.Component {
...
-          value={this.state.typedValue || this.state.value}
-          onChange={this.handleInputChange}
-          onFocus={this.handleInputFocus}
-          onBlur={this.handleInputBlur}
-          onKeyDown={this.handleInputKeyDown}
-          onKeyUp={this.handleInputKeyUp}
-          onClick={!inputProps.disabled ? this.handleInputClick : undefined}
-        />
+        {this.props.component &&
+        {}.toString.call(this.props.component) !== '[object Function]' ? (
```
panczarny: Why not just `typeof this.props.component === 'function'`?

glockjt: This way is more verbose.  Not really relevant here but one good example is that it will return `Array` for `Array` instead of `Object`.  A better way would be to actually have a function that would extract out `Function`.

src/DayPickerInput.js
glockjt: extracted.

saintplay:
Any progress on merging this?

gpbl:
This will be solved in the upcoming v8: see #942 . Thanks!

Issue #819, opened 2018-10-18
Datepicker doesn't hide on outside click when using custom overlay


peterbartels:
Try it here:
http://react-day-picker.js.org/examples/input-custom-overlay

- Click on the textbox to open the datepicker
- Click on arrow-right (to go to the next month)
- click outside the datepicker
- it doesn't close (only when selecting a date)

Same problem occurs when using a custom navbar

matthova:
Ran into the same problem. Going to do a bit of digging to see if I can find out why, but let me know if you've found a workaround

matthova:
PR: https://github.com/gpbl/react-day-picker/pull/871

matthova:
As described in #871 I'm not sure how to patch the bug while maintaining standard behavior.
Instead  I'm going to create a fork where `showOverlay` is driven by props.

TSilviu:
Hi there! I ran into the same issue when providing a custom input component to the `DayPickerInput` component.

@matthova Thanks for looking closer into this, I can see you tried a couple of implementations.

Are there any plans to address this issue? What about known workarounds? In my opinion, this is quite a serious problem as I can imagine a considerable amount of users will want to provide their own input or overlay components.

matthova:
@TSilviu the custom implementation shown in the examples in #872 has been working great for us.

MuhammadSaadQadeer:
Is this fixed ? @matthova I tried your custom example but its not working for me, any insights on this ? Also I am using this with formik

sakerrr:
+1

PR #871, opened 2019-02-22
Hide overlay on overlayblur

matthova:
This PR addresses the bug described in issue #819.  

Old behavior: When using a custom overlay, after clicking on anything in the overlay, then clicking away, the overlay does not close. Once the overlay has been focused, the only way to make the overlay close is to click on the input and then click away.

New behavior: After clicking on anything in the overlay, then clicking away, the overlay will close.

Note: With this PR - If the user clicks on the input field, after clicking on something inside of the overlay, the overlay will momentarily close and then reopen. This is not ideal, but it is better than the current behavior. If there is a preferred way to avoid calling `setState` when clicking on the input field, let's do that instead 👍.

🚀 Amazing library 🚀  I hope this PR helps

### DIFF
--- a/src/DayPickerInput.js
+++ b/src/DayPickerInput.js
@@ -394,6 +394,7 @@ export default class DayPickerInput extends React.Component {
     // focusing it
     this.overlayBlurTimeout = setTimeout(() => {
       this.overlayHasFocus = false;
+      this.setState({ showOverlay: false });
     }, 3);
   }
 


matthova:
Tested this a bit more. It solves one particular issue, but creates a bunch more issues with the standard React-Day-Input. Will revisit this once I have more time, but please let me know if a fix comes to mind.

matthova:
I'm not sure how to patch the bug while maintaining standard behavior.
Instead of pursuing fixing the bug as described  in #819 I'm going to create a fork where `showOverlay` is driven by props.

nevace:
@matthova I had this same issue a while ago, stumble across this while looking for a way to keep the overlay always open.

I ended up doing this to get around it, not ideal but might help

```
handleClickOutside = ({ target }) => {
    //this.datepicker is a ref to the container DayPickerInput is in and this.dayPickerInstance is the ref to DayPickerInput
    this.datepicker && !this.datepicker.contains(target) && this.dayPickerInstance.hideDayPicker();
};

componentDidMount() {
    document.addEventListener('click', this.handleClickOutside, true);
}

componentWillUnmount() {
    document.removeEventListener('click', this.handleClickOutside, true);
}
```

PR #872, opened 2019-02-25
Controlled showOverlay state

matthova:
This PR allows the showOverlay prop to drive showOverlay state of day-picker-input. It uses React's getDrivedStateFromProps lifecycle, but it could be changed to use componentWillReceiveProps if necessary.  

I've updated the input-custom-overlay.js code to remove the bug as described in #819.


matthova:
Closing again. I should have put my other work in a different branch 🤷‍♂️

PR #873, opened 2019-02-27
Add formatDate and parseDate to typings

no23reason:
The `formatDate` and `parseDate` are not mentioned in the typings for LocaleUtils although they [contain them](https://github.com/gpbl/react-day-picker/blob/master/src/addons/MomentLocaleUtils.js#L49-L61). This means the code in [the example](http://react-day-picker.js.org/examples/input-moment/):

```ts
import MomentLocaleUtils, {
  formatDate,
  parseDate,
} from 'react-day-picker/moment';
```

 ends up with an error in TypeScript. If this is merged, this code could be replaced by

```ts
import MomentLocaleUtils from 'react-day-picker/moment';
// MomentLocaleUtils contains formatDate and parseDate even in TypeScript
```


lukewis:
It seems that this "worked" in v7.2.4, but broke with v7.3.0

ludaval:
You declared the formatDate signature as:
`formatDate(date: Date, format?: string | string[], locale?: string): string;`
But given that the [moment.js 'string + format' method signature](https://momentjs.com/docs/#/parsing/string-format/) declares the format to be a string and not a string[], shouldn't you change it to:
`formatDate(date: Date, format?: string, locale?: string): string;` .
Also, the method parseDate [can also return undefined](https://github.com/gpbl/react-day-picker/blob/master/src/addons/MomentLocaleUtils.js), so the return type should be changed to `Date | undefined` .. IMO

singingwolfboy:
`formatDate` and `parseDate` are _not_ exported in the [`LocaleUtils`](https://github.com/gpbl/react-day-picker/blob/master/src/LocaleUtils.js) file. They _are_ exported in the [`MomentLocaleUtils`](https://github.com/gpbl/react-day-picker/blob/master/src/LocaleUtils.js) file, however. I made a different pull request (#895) that reflects this situation more accurately.

gpbl:
Is this fixed via #899?

estephinson:
> Is this fixed via #899?

I would say not, in #899 there is no change to the typings, the issue is purely formatDate and parseDate are not exposed. Correct me if I'm wrong, but there is no change in #899 to include formatDate and parseDate?

estephinson:
Current workaround for me is to simply define the functions in the project yourself, as thankfully they're relatively simple:

```javascript
function formatDate(date, format = 'L', locale = 'en') {
    return moment(date)
        .locale(locale)
        .format(Array.isArray(format) ? format[0] : format);
}

function parseDate(str, format = 'L', locale = 'en') {
    const m = moment(str, format, locale, true);
    if (m.isValid()) {
        return m.toDate();
    }
    return undefined;
}
```

Issue #861, opened 2019-02-01
The default `initialMonth` is determined on import instead of on mount


batmantis:
The default value of the `initialMonth` prop is determined when the file is imported instead of when the component is mounted.  This is the opposite of what I would expect to happen.  For example, if the component is loaded close to midnight on the last day of the month and the component is mounted the following month, then the month will be wrong.

It's a little tough to demo the behavior, but this should give a general idea: https://codesandbox.io/s/nrxzq0223l

I plan to create a PR to fix this.

gpbl:
Hey this is an interesting bug. Thanks for catching it up and the repro.

PR #874, opened 2019-02-27
Remove initialMonth from defaultProps

batmantis:
This is a fix for https://github.com/gpbl/react-day-picker/issues/861.  

I wasn't able to come up with a nice way to test that the "initial month" of the calendar is based on `Date.now()` when the component is mounted instead of `Date.now()` when the component is imported.  I think it would require some monkey patching of `Date` and I didn't want to add `timemachine` as a dependency.  In addition, I was running into stack overflows related to some kind of circular logic between `timemachine` and `react-day-picker`.

This change worked for me as expected when I manually tested it.

I'm happy to make any additional changes based on feedback.

Fixes: #861


batmantis:
@gpbl What info is needed for this?

batmantis:
I resolved the merge conflict.  I think the typescript test failure is not related to this PR, I will try tests again when it is fixed.

batmantis:
Tests are now passing again after merge

PR #883, opened 2019-03-26
Added formatDate&parseDate definitions for MomentLocaleUtils

Jefski14:
type definitions for `formatDate` and `parseDate` were missing, so I added them.

As described I added missing type definitions to the MomentLocaleUtils.

Since I'm fairly new to writing typescript definitions it would be great if someone who's good with TS reviews my change.


Jefski14:
Just saw that there is already a PR ( #873  )
Closing this PR

PR #885, opened 2019-03-27
Add TypeScript declaration for utils.js

adam187:
Add missing declaration file for `utils.js` file.

```typescript

import { localeUtils, dateUtils, modifiersUtils } from 'react-day-picker/utils'

```

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "moment.js",
     "moment.d.ts",
     "utils.js",
+    "utils.d.ts",
     "types"
   ],
   "directories": {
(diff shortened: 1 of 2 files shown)


DylanVann:
There doesn't need to be a declaration file for every source file, there just needs to be declarations for the public API

adam187:
@DylanVann This file [utils.js](https://github.com/gpbl/react-day-picker/blob/master/utils.js) actually looks to me like public API.  

Currently, you can't use utils function like in docs in typescript because of type/export conflicts but looks like [this PR](https://github.com/gpbl/react-day-picker/pull/899/files) should solve it.

```typescript
import { DateUtils } from "react-day-picker";
```

`'DateUtils' only refers to a type, but is being used as a value here.ts`

DylanVann:
Yeah I’m aware of the issue I’m just not sure this will do anything to help, also aware of that PR since I made it :)

gpbl:
#899 has been merged 🎊

PR #893, opened 2019-04-11
add missing utils typescript definitions

ludaval:
Not very much to explain actually. It seemed to me that some type definition were missing, because all the other functions in MomentLocaleUtils.js were declared correctly in the utils.d.ts definition file, except the ones that i added. So I added them.


ludaval:
this is a dupe of #873 . closing

PR #909, opened 2019-07-06
Add tabIndex to DayPickerProps

clarityflowers:
It looks like this is [included in the propTypes](https://github.com/gpbl/react-day-picker/blob/421af9f56b1b07ed58f864678b3bbb8617cdaff7/src/DayPicker.js#L92) for the component, but didn't get added to the typescript typings.


gpbl:
Hey thank you for the PR! I'm trying to resume this project , sorry for the late answer.

PR #918, opened 2019-08-07
Add style type to DayPickerInputProps

KonstantinKudelko:
So simple hotfix.


types/props.d.ts
```diff
@@ -170,6 +170,7 @@ export interface DayPickerInputProps {
   component?: any;
   overlayComponent?: any;
 
+  styles?: object;
```
chrise86: I think this should be `style`, not `styles`.

Possibly even using `any` too:
```
style?: any
```

PR #923, opened 2019-08-22
Make `format` optional in DayPickerInput

TomPridham:
* Please explain clearly your changes.

the `format` prop in the `DayPickerInput` component is not required as far as i can tell in my project. this just updates the types to match


gpbl:
Thanks!

PR #928, opened 2019-09-03
Add missing locale utils type definitions

RobertMaciejewski:
So, today I wanted to add date formatting in a project using typescript and while [following the documentation](https://react-day-picker.js.org/examples/input-moment) I've noticed that there are some definitions missing for `MomentLocaleUtils`. So here's a small fix :wink:

### DIFF
--- a/types/utils.d.ts
+++ b/types/utils.d.ts
@@ -24,6 +24,8 @@ export const LocaleUtils: {
     string,
     string
   ];
+  formatDate(date: Date, format?: string, locale?: string): string;
+  parseDate(str: string, format?: string, locale?: string): string;
 };
 
 export const DateUtils: {


faqini:
Stumbled on the same problem. This fixes it, merge please 🙏

faqini:
Stumbled on the same problem. Merge please 🙏

types/utils.d.ts
```diff
@@ -24,6 +24,8 @@ export const LocaleUtils: {
     string,
     string
   ];
+  formatDate(date: Date, format?: string, locale?: string): string;
+  parseDate(str: string, format?: string, locale?: string): string;
```
gpbl:
```suggestion
  parseDate(str: string, format?: string, locale?: string): Date | undefined;
```

faqini:
Did another test with the latest version of TS 3.6.4, I see no error is raised if you just use `Date` as a return type for parseDate. This was addressed here:
https://github.com/gpbl/react-day-picker/commit/28c4282330d79704504b04be0e7e3b160029370f

I guess you can close this one.

PR #935, opened 2019-09-20
Upgrade dev dependencies

CodeWitchBella:
This PR upgrades all devDependencies to their latest version and makes changes required to make it work.

## Babel

Upgraded to babel 7, replaced packages with their new `@babel/*` equivalent and removed `transform-object-rest-spread` because that's there by default.

## Eslint

Upgraded to latest versions and turned off errors due to new preset which this package violates. (can be reenabled later, but I wanted to focus on upgrading tools without changing code)

## Jest

Fixed deprecation warning after update

## Webpack

Added explicit mode setting and installed webpack-cli.


CodeWitchBella:
Seems like this is already included in next branch. Closing

PR #937, opened 2019-09-26
Add click handler for week day

isabeau-bergeron:
Add a click handler for the week day label (such as Su, Mo, Tu, ...).

There is already a click handler for the Month label and Week label. This is a need for us so I thought I would share it!


isabeau-bergeron:
@gpbl Why close the PR?

PR #943, opened 2019-10-09
The small fix: unnecessary bracket in the example

Diligens:
The small fix: unnecessary bracket in the example


gpbl:
Thank you!

Issue #942, opened 2019-10-08
React DayPicker 8


gpbl:
v8 will be a TypeScript rewrite and will introduce some major, needed changes.

* See project https://github.com/gpbl/react-day-picker/projects/7
* Temporary website: https://react-day-picker-next.netlify.app
* Latest docs: https://github.com/gpbl/react-day-picker/tree/master/website/docs

## Testing v8

Add the next version to the dependencies:

```bash
yarn add react-day-picker@next
```

## v8 Objectives

* Rewrite the package to follow the latest React patterns
* Improve API and customization with custom components
* Resolve [TypeScript typing issues](https://github.com/gpbl/react-day-picker/issues?utf8=✓&q=is%3Aissue+label%3Asrc%3ATypeScript+) by rewriting it in TypeScript
* Replace `DayPickerInput` with a more flexible alternative, e.g. an hook. 
  - The component has a [maintenance cost](https://github.com/gpbl/react-day-picker/issues?q=is%3Aissue+label%3Asrc%3ADayPickerInput) I can't handle
* use [date-fns](https://date-fns.org/) as date utility ~~peer dependency~~
* have a browserstack functional tests
* have a new website 

### Notes

* `DayPickerInput` seems very popular and removing it can make difficult the upgrade.
* Need to be React +16.8
* Likely we will need to maintain v7 for a while

## Please help 🙏🏾

**With more than 1M download per month, react-day-picker needs love**

It is time consuming to work alone on this project, organize the repository, write documentation, the functional tests, and the upgrade guides (not native speaker here). 

Please help, talk on https://spectrum.chat/react-day-picker 

I started this project five years ago as a way to contribute to the awesome React community. I've learnt a lot, I had a lot of fun, and met the best developers thanks to it. 

However, after all these years, the old codebase is not much fun to maintain and after so many changes, the code is not as good as it should. Developers started filling issues on Github, and the glorious OSS experience I had so far hit me to the point I didn't want to read Github notifications anymore.

Surprisingly tho, downloads never stopped to grow and this library is still popular! People continue to send PRs, and some even wrote me that it would be sad to see this library to die. These nice messages have been highly motivational so I'm here again ❤️

shelver2004:
hello @gpbl i need some help with captionElement for 2 calendars for date range, how can i have changeable month and year dropboxes? do you have an example for date range date picker?

KyorCode:
Will there be a `useInput` hook Ranges?

[...]

PR #944, opened 2019-10-10
v8.0.0 🚢

gpbl:
See issue: https://github.com/gpbl/react-day-picker/issues/942 

This is still a work in progress and code may change substantially, still I'd love some reviews of the work so far...


src/Head.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import { getWeekdaysNames } from './utils/getWeekdaysNames';
+import { HeadProps } from '../typings/react-day-picker';
+
+export const Head: React.FC<HeadProps> = props => {
+  const { locale, showWeekNumber, dayPickerProps } = props;
+  const { classNames, styles, formatWeekdayName } = dayPickerProps;
+  const weekdayNames = getWeekdaysNames(locale, formatWeekdayName);
+  return (
+    <thead style={styles.head} className={classNames.head}>
```
ryanburr: Is there a benefit to using the `style` prop over `className`? Doesn't `style` take precedent over anything applied with `className`? I know this can make it difficult to override styles. Just trying to avoid someone trying to override with `!important` at all costs.

src/Navigation.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  const handlePrevClick = (e: React.MouseEvent<HTMLButtonElement>): void => {
+    onMonthChange && prevMonth && onMonthChange(prevMonth, e);
+    onPrevClick && prevMonth && onPrevClick(prevMonth, e);
+  };
+  const handleNextClick = (e: React.MouseEvent<HTMLButtonElement>): void => {
+    onMonthChange && nextMonth && onMonthChange(nextMonth, e);
+    onNextClick && nextMonth && onNextClick(nextMonth, e);
+  };
+  const handleStartClick = (e: React.MouseEvent<HTMLButtonElement>): void => {
```
ryanburr: Does assigning function like this create new references ever render? I've always defined functions like so...
```
export const MyComp: React.FC<MyCompProps> = () => {
	// ... implementation
	
	return <div onClick={handleOnClick} />;

	function handleOnClick() {
		// handler
	}
};
```

src/Day.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+import { getDayProps } from './helpers/getDayProps';
+import { DayProps } from '../typings/react-day-picker';
+export const Day: React.FC<DayProps> = props => {
+  const { day, modifiers, dayPickerProps } = props;
+  const { locale, formatDay } = dayPickerProps;
+
+  const { containerProps, wrapperProps } = getDayProps(
```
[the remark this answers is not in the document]
gpbl: I was wondering if it is worth to add it or not. My conclusion is to see the userland feedback first – i.e. better keep performance optimizations after the major release.

src/Head.tsx
gpbl: Good question! These styles are user defined inline-styles: they come from the optional `styles` prop. Example: https://rdp-v8.netlify.com/docs/styling#using-inline-styles

ryanburr: Fair enough. If this is something the user has full control over, seems alright to me

src/Navigation.tsx
ryanburr: Probably goes along with your performance optimization response from my other comment... carr yon

gpbl: That is an interesting solution, do you have more insights about that pattern?

Anyway, yes it creates new references. We may want to use `useCallback` if worth it (but I don't think it is).

package.json
```diff
@@ -1,106 +1,69 @@
...
+    "rollup-plugin-typescript2": "^0.24.3",
+    "ts-jest": "^24.1.0",
+    "tslib": "^1.10.0",
+    "typescript": "^3.6.4"
   },
-  "dependencies": {
-    "prop-types": "^15.6.2"
+  "peerDependencies": {
+    "date-fns": "^2.2.1",
+    "react": "^16.10.2"
```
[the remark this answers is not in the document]
gpbl: Sure thanks!

gpbl:
@ryanburr I really appreciate your review , can i have more? :P Thanks a lot!

src/Navigation.tsx
ryanburr: I just read through [this article](https://kentcdodds.com/blog/usememo-and-usecallback) that has a really good explanation of what's going on here.

In summary, `useMemo` is going to be better, but the optimizations aren't worth the code complexity in this scenario. 

As far as defining a function scoped to the component (like my example) and what you have, I haven't really found any articles or done extensive perf testing myself. I think it stems from the `eslint(react/jsx-no-bind)` rule I usually code with that yells at me.

gpbl:
> I'm curious as to why you went with the namespace for types? Nothing wrong with it. I'm personally a fan of specific named imports for everything, but then again it's not my library :p

@ryanburr I'm new to TypeScript and I am changing my mind pretty often :) 

I'm trying to understand what are the best practices for a react npm library in typescript. Do you think just named imports are enough? I thought namespace was a good idea, but I have no clue what are the consequences :)

ryanburr:
@gpbl I've seen both as far as namespaces go. It's really just preference. I don't think there's a benefit to one or the other. Just subjective readability

rollup.config.js
```diff
@@ -0,0 +1,89 @@
...
+        exports: 'auto',
+        sourcemap: true,
+      },
+    ],
+    external: id =>
+      !id.startsWith('\0') && !id.startsWith('.') && !id.startsWith('/'),
+    plugins: [
+      resolve(),
+      typescriptPlugin,
+      replaceDevelopment,
```
TrySound: NODE_ENV replacement should be done on user side. It should be replaced only for umd which usually run directly in browser.

rollup.config.js
```diff
@@ -0,0 +1,89 @@
+import { sizeSnapshot } from 'rollup-plugin-size-snapshot';
+import autoExternal from 'rollup-plugin-auto-external';
```
TrySound: Not used

rollup.config.js
```diff
@@ -0,0 +1,89 @@
...
+      {
+        file: pkg.main,
+        format: 'cjs',
+        exports: 'auto',
+        sourcemap: true,
+      },
+      {
+        file: pkg.module,
+        format: 'esm',
+        exports: 'auto',
```
TrySound: Omit this option in both cases. Auto is default. If you will add default export (please don't). Rollup will warn you.

rollup.config.js
```diff
@@ -0,0 +1,89 @@
+import { sizeSnapshot } from 'rollup-plugin-size-snapshot';
+import autoExternal from 'rollup-plugin-auto-external';
+import copy from 'rollup-plugin-copy';
+import replace from 'rollup-plugin-replace';
+import { terser } from 'rollup-plugin-terser';
+import resolve from 'rollup-plugin-node-resolve';
```
TrySound: No reason in node resolve plugin as you do not have any dependency (except peer ones).

gpbl:
Thanks @TrySound you are the boss! See the changes here :)
https://github.com/gpbl/react-day-picker/pull/944/commits/f26da848a7791bca4f332230d7e35b10d69e858a

PR #947, opened 2019-10-21
Remove `typeof` for LocaleUtils

Kjaer:
In the #873 , `LocaleUtils` is not a `const` variable and it is and `Interface` there no need to use `typeof` for it's type usage. (This is breaking a build pipeline when tsconfig is set `strict:true`)

That's why I remove them.

### DIFF
--- a/types/DayPicker.d.ts
+++ b/types/DayPicker.d.ts
@@ -12,7 +12,7 @@ import {
 
 export default class DayPicker extends React.Component<DayPickerProps, any> {
   static VERSION: string;
-  static LocaleUtils: typeof LocaleUtils;
+  static LocaleUtils: LocaleUtils;
   static DateUtils: typeof DateUtils;
   static ModifiersUtils: typeof ModifiersUtils;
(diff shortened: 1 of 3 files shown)


Kjaer:
Forget it 😅 
I saw you remove those file just before I created the PR.
Such a pity for me but decision is yours :)

gpbl:
Sorry @Kjaer I missed this PR for some reasons!

gpbl:
Could you help with the conflicts?

KjaerBL:
I'd love to, but they have been removed.(https://github.com/gpbl/react-day-picker/commit/884dbea6ed5e60c8ce724ebd33c9abf3d3e5c646)

I am not sure, do you still want them or not?

gpbl:
We need to fix this: https://github.com/gpbl/react-day-picker/issues/948 :) but i have no time for now ...

BlackDark:
You should change the target branch to `v7` i guess

gpbl:
The fix should be already on master – i just need to test it properly before publishing!

PR #954, opened 2019-11-05
Reverse onMonthChange check.

KyorCode:
Small bug which disabled navigation when you use the useInput hook.


gpbl:
What is this change for? 🤔

KyorCode:
on your V8 / master branch. If the currentMonth is controlled, the internal function should be called so the currentMonth state gets set. I updated the file.

PR #967, opened 2019-12-11
a11y fixes for v7

hannahkowen:
Problem:
1.	The full date of a day cell in the calendar day picker grid is not announced with NVDA, just the number of the day (e.g. 11 instead of Wednesday November 11th, 2019)
2.	The month name is not announced when changing between months

Solution:
1.	Some screen readers (NVDA) do not read ariaLabels directly on `<div>`s. 
a.	Place the visible text in a span hidden from screen readers
b.	have the screen readers read from a separate span that contains the ariaLabels.
c.	Add css class to hide only the screen reader span from display.
2.	Add aria-live=”polite” to caption so that it reads the month when changing between months.

### DIFF
--- a/src/Caption.js
+++ b/src/Caption.js
@@ -51,7 +51,7 @@ export default class Caption extends Component {
       onClick,
     } = this.props;
     return (
-      <div className={classNames.caption} role="heading">
+      <div className={classNames.caption} role="heading" aria-live="polite">
         <div onClick={onClick} onKeyUp={this.handleKeyUp}>
           {months
             ? `${months[date.getMonth()]} ${date.getFullYear()}`
--- a/src/Day.js
+++ b/src/Day.js
@@ -137,7 +137,6 @@ export default class Day extends Component {
         tabIndex={tabIndex}
         style={style}
         role="gridcell"
-        aria-label={ariaLabel}
         aria-disabled={ariaDisabled}
         aria-selected={ariaSelected}
         onClick={handleEvent(onClick, day, modifiers)}
@@ -150,7 +149,8 @@ export default class Day extends Component {
         onTouchStart={handleEvent(onTouchStart, day, modifiers)}
         onFocus={handleEvent(onFocus, day, modifiers)}
       >
-        {children}
+        <span className="DayPicker-ScreenReaderOnly">{ariaLabel}</span>
+        <span aria-hidden="true">{children}</span>
       </div>
     );
   }
--- a/test/daypickerinput/rendering.js
+++ b/test/daypickerinput/rendering.js
@@ -119,7 +119,9 @@ describe('DayPickerInput', () => {
       wrapper.instance().showDayPicker();
       wrapper.update();
       expect(wrapper.find('.DayPicker-Day--selected')).toHaveLength(1);
-      expect(wrapper.find('.DayPicker-Day--selected')).toHaveText('15');
+      expect(wrapper.find('.DayPicker-Day--selected')).toHaveText(
+        'Fri Dec 15 201715'
+      );
     });
     it('should clear timeouts when component unmounts', () => {
       const container = document.createElement('div');
--- a/src/style.css
+++ b/src/style.css
@@ -210,3 +210,16 @@
   background: white;
   box-shadow: 0 2px 5px rgba(0, 0, 0, 0.15);
 }
+
+/* DayPicker a11y helpers */
+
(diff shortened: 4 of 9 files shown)


src/Caption.js
```diff
@@ -51,7 +51,7 @@ export default class Caption extends Component {
       onClick,
     } = this.props;
     return (
-      <div className={classNames.caption} role="heading">
+      <div className={classNames.caption} role="heading" aria-live="polite">
```
gpbl, line 54: @hannahkowen this is great thanks!

src/Day.js
```diff
@@ -150,7 +149,8 @@ export default class Day extends Component {
         onTouchStart={handleEvent(onTouchStart, day, modifiers)}
         onFocus={handleEvent(onFocus, day, modifiers)}
       >
-        {children}
+        <span className="DayPicker-ScreenReaderOnly">{ariaLabel}</span>
+        <span aria-hidden="true">{children}</span>
```
gpbl, line 153: I fear this line will be a breaking change for many. Could we just add `aria-live="polite"` and the "screen-reader-only part – or are these changes related?

src/Day.js
```diff
@@ -150,7 +149,8 @@ export default class Day extends Component {
         onTouchStart={handleEvent(onTouchStart, day, modifiers)}
         onFocus={handleEvent(onFocus, day, modifiers)}
       >
-        {children}
+        <span className="DayPicker-ScreenReaderOnly">{ariaLabel}</span>
```
gpbl, line 152: User can set their own [classNames](http://react-day-picker.js.org/api/DayPicker#classNames): this className should be added to the list.

test/daypickerinput/rendering.js
```diff
@@ -119,7 +119,9 @@ describe('DayPickerInput', () => {
       wrapper.instance().showDayPicker();
       wrapper.update();
       expect(wrapper.find('.DayPicker-Day--selected')).toHaveLength(1);
-      expect(wrapper.find('.DayPicker-Day--selected')).toHaveText('15');
+      expect(wrapper.find('.DayPicker-Day--selected')).toHaveText(
+        'Fri Dec 15 201715'
```
gpbl, line 123: This change is unexpected :)

src/Day.js
trichards-shedd: hey, friend! I'm one of Hannah's coworkers helping get this PR up to snuff. Is the issue that `Day`'s `children` would be pushed down a level in the DOM tree? We placed `children` outside of the span tag again with no poor effects from the screenreader's perspective, if that's acceptable.

src/Day.js
trichards-shedd: this className is always required to hide this from a typical user's view. are you asking that a key on the `classNames` object be added for the screenreader text?

test/daypickerinput/rendering.js
trichards-shedd: unfortunately, no matter how hard we hide it, the text the screenreader reads is still present. if they were two separate spans, we could select into them and verify each's content. could also use `toIncludeText` instead but that feels flakier :^(

src/Day.js
gpbl: Yes! Every class name used in the component can be replaced with a custom one via that `classNames` prop. We need to add it here https://github.com/gpbl/react-day-picker/blob/v7/src/classNames.js.

src/Day.js
gpbl: My problem is that I cannot add breaking changes to v7, as i have no time left :)

Could you open an issue describing better the problem?

My concern is that this change may break other DayPicker implementations: e.g. what happens to localized calendars? Must the existing DayPickers add an `ariaLabel` after this change?

test/daypickerinput/rendering.js
gpbl: The tests are there to not break them... if they break, then is a breaking change, and I cannot work on a major release right now :(

gpbl:
Let's do this:

- open a PR with only the `aria-live=”polite”` change. This is not breaking so it is easy to publish.
- open a new issue describing the other problem. I don't understand, for example, if this is only for the NVDA readers. 

I'd prefer to find a solution that is not breaking. I am working on v8 so I don't have time for this. If you prefer [help me there](https://github.com/gpbl/react-day-picker/projects/7)!

Thanks!

PR #1002, opened 2020-03-11
Added TodayButtonElement Prop for daypicker

kevinlandsberg:
Added a property for the daypicker to render a custom todayButtonElement inside the footer. If the property ```todayButtonElement``` exists, we will render the custom component instead of the default ```<button>```.


kevinlandsberg:
No Answer...

PR #1004, opened 2020-03-30
aria-disabled and aria-selected break when using custom classNames prop

camflan:
When checking if an item is selected or disabled, we need to use the classNames given to the component rather than hardcoded strings. When using hardcoded strings, it will break when custom classNames are provided


gpbl:
Thanks a lot @camflan ! Sorry for the delay.

gpbl:
This should be fixed in v7.4.1!

PR #1008, opened 2020-04-10
Remove initialMonth from defaultProps (#2)

batmantis:
Original Issue: https://github.com/gpbl/react-day-picker/issues/861
Original PR: https://github.com/gpbl/react-day-picker/pull/874

I originally submitted a PR to fix this bug last year, however I mistakenly edited the wrong source files (`/lib/src`) instead of (`/src`).  Consequently this bug fix never made it into v7.4.  This PR makes the same bug fix, but in the correct file.

### DIFF
--- a/src/DayPicker.js
+++ b/src/DayPicker.js
@@ -132,7 +132,6 @@ export class DayPicker extends Component {
   static defaultProps = {
     classNames,
     tabIndex: 0,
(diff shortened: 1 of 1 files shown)


gpbl:
Thanks – how did i miss that in the review :)

Issue #948, opened 2019-10-21
LocaleUtils in not a value, causing types to be broken


MhMadHamster:
In this pr: https://github.com/gpbl/react-day-picker/pull/873 `LocaleUtils` was changed from value to type (interface), i'm not sure if it's intended, but in the latest version v.7.4.0 types are broken, cause there are a lot of `typeof LocaleUtils` which expects value, not the type.

P.S. consider adding tests for types

PiTomar:
Can confirm, same issue here.

madebyew:
This is an issue for me, too.

gpbl:
Sorry for the bugged release, people, but I need help maintaining v7 as I'm busy working with the next major version. Please branch off `v7` branch to send a PR, thanks 🙏🏾 
I'll see if I can work on this today.

zvertovsek:
Any updates on the fix?

tpetrina:
Downgrade to 7.3.2 and it should work.

n-romaniv:
Thanks for the great library, guys! @gpbl, any updates?

kevinlandsberg:
Still no updates... :S

gpbl:
This should be fixed in v7.4.7, hope so. If not please open another issue, this one was lost into the list of untriaged stuff :) Thanks @MhMadHamster!

Issue #1009, opened 2020-04-13
Types destroyed v7.4.1


tonoslav:
**Describe the bug**
Getting a lot of wrong errors from types

1. Using **DayPickerInput** with **dayPickerProps** prop, keep getting error that **renderDay**, **renderWeek** are required , but should they be?

2. ```js
    import { DateUtils } from 'react-day-picker'
    DateUtils.isDate(...) 
    ```
    **'DateUtils' only refers to a type, but is being used as a value here.**
    **Fix:** (removed **export** before **type**) +
    ![image](https://user-images.githubusercontent.com/3399021/79157131-545ad880-7dd4-11ea-849c-b2ede1314483.png)

3. **DayPickerInputProps** props **onDayPickerShow**, **onDayPickerHide** are required but should they?

mr-wildcard:
Also experiencing 1️⃣. I had to hide `dayPickerProps` prop of `DayPickerInput` component behind a `@ts-ignore`

sbusch:
I think the 1st one should be fixed by my PR https://github.com/gpbl/react-day-picker/pull/1012 (merged quickly by @gpbl, thank you!), not yet published to npm.

Regarding 2nd bug, I fixed the same error for another export (https://github.com/gpbl/react-day-picker/pull/1011), so the fix here could be similar?!

3rd: the JS implementation checks if the props are set before calling those event handlers, so I think they are not mandatory and should be fixed similar to 1.

redaxmedia:
I experienced the same with `MomentLocaleUtils` ... that broken release should be unpublished

andrew-r-king:
FYI: A workaround is to stay at the previous version by removing "^" in front of the version in your package.json:

```
 "react-day-picker": "7.4.0",
```

gpbl:
I'm sorry for that! The hard life of a maintainer :) I should never trust PRs without proper testing and the release was rushed up.

Setting up a typescript app to test it, let see if I come out. 

Please send PRs to help as I'm a bit busy these days 🙏🏽

bengry:
Part of this is already fixed by https://github.com/gpbl/react-day-picker/pull/1012
I'll send a PR to fix the `DateUtils` and `ModifierUtils` export in a bit and update here.

redaxmedia:
You could just `npm unpublish react-day-picker@7.4.1 --force` like I suggested above... anyway, I think this is only allowed during first 24h or something.

gpbl:
[...]

PR #1014, opened 2020-04-15
Fix TypeScript types ruined in v7.4.1

bengry:
Fixes issues 2 and 3 raised in https://github.com/gpbl/react-day-picker/issues/1009

### DIFF
--- a/types/props.d.ts
+++ b/types/props.d.ts
@@ -176,13 +176,13 @@ export interface DayPickerInputProps {
 
   classNames?: InputClassNames;
 
-  onDayChange: (
+  onDayChange?: (
     day: Date,
     DayModifiers: DayModifiers,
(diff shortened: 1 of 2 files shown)


bengry:
Sure thing. Is this going to be released as part of v7.4.2? I was waiting for 7.4.1 initially for #1004, but since types are broken there it's a blocker for my use-case. I also think it'd be good to push this and maybe even deprecate v7.4.1 since it's broken when using TypeScript in some scenarios (namely, when you don't pass some optional props and/or rely on the *Utils fixed in this PR), at least without a `@ts-ignore` on those usages.

MhMadHamster:
This pr fixes #948 as well, i suppose?

bengry:
> This pr fixes #948 as well, i suppose?

Yes

PR #1016, opened 2020-04-16
(v7) Fix types, UMD build

### LINKED ISSUE #981: unpkg usage throwing react type is invalid error
**Describe the bug**
When I try to use react-day-picker I get the follow errors in my console
```
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
```

**To Reproduce**
I can reproduce the error in the sandbox here -https://codesandbox.io/s/react-day-picker-base-vt99v

**Expected behavior**
Should be able to use the react-day-picker normally

gpbl:
I've been investigating the recent issues with types etc. There were multiple issues that were related how we build files.

This should fix #981 and #1009


types/MomentLocaleUtils.d.ts
```diff
@@ -1,2 +1,2 @@
-import { LocaleUtils } from './utils';
+import { LocaleUtils } from './LocaleUtils';
 export const MomentLocaleUtils: LocaleUtils;
```
nstepien, line 2: @gpbl 
This type isn't correct, when trying to import
```ts
import utils from 'react-day-picker/moment';
```

The resulting type for `utils` is `{ MomentLocaleUtils: LocaleUtils }`.
I think the proper fix would be:

```ts
export default LocaleUtils;
```

gpbl: Hey thanks for the heads up! releasing a patch...

gpbl: Should be fixed in v7.4.7.

types/index.d.ts
```diff
@@ -1,4 +1,7 @@
+export as namespace DayPicker;
+
 export { default } from './DayPicker';
-export * from './common';
-export * from './props';
-export * from './utils';
+export * from './Props';
```
mikaelpaj, line 4: The props declaration filename is still in lowercase, this seems to break some types.

types/MomentLocaleUtils.d.ts
nstepien: Thanks! It's working great.

types/index.d.ts
gpbl: Thanks @mikaelpaj checkout v7.4.8 – should have fixed it.

types/MomentLocaleUtils.d.ts
gpbl: phew!!

types/index.d.ts
mikaelpaj: Fixed our issue, thanks for the quick reaction!

PR #1019, opened 2020-04-17
Replace rollup with webpack

gpbl:
I've been hit by few problems with Rollup and Typescript:

- https://rollupjs.org/guide/en/#error-this-is-undefined
- https://github.com/rollup/plugins/issues/243

so I prefer to switch back to webpack which seems having better Typescript support from community.

This PR adds also support for prettier.


TrySound:
Please don't. This is the problem of typescript which adds "this" to es modules. This does not exist in es modules. Smart guys told me babel should be used instead.

TrySound:
Your users will anyway face this issue when use rollup as app bundler. Here it just exposes existing problem.

gpbl:
Thanks for chiming in, @TrySound ! 

https://github.com/rollup/plugins/issues/243 was a blocker here — i’m quicker with webpack, so for now I prefer to use that as i need to continue with v8.

PR #1024, opened 2020-04-21
Rename types file to uppercase because linux imports are case sensitive

masiulis:
Hey @gpbl I noticed that the newest version was failing on our CI as it couldn't find the props.d.ts. Everything works fine on mac, because mac imports are not case sensitive.


gpbl:
Thanks – published as v7.4.8!

PR #1038, opened 2020-05-04
Add aria-polite to announce month change during navigation

kfsass:
Hiya! Friend from Shedd here again - it turns out the other issue was actually fixed in a later version of v7 (and we just needed to update), so here's a PR just for the `aria-live="polite"` bit!  This allows the NVDA screenreader to announce the changing text in the month header. Thank you for your work! 💖

### DIFF
--- a/src/Caption.js
+++ b/src/Caption.js
@@ -51,7 +51,7 @@ export default class Caption extends Component {
       onClick,
     } = this.props;
     return (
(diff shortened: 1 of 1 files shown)


gpbl:
Woah that's a good news! I had other alternatives ready – like using the `formatDay` prop :) Thanks for this.

gpbl:
Trying to publish this today...

PR #1048, opened 2020-06-23
fix(Caption): improve accessibility roles for Caption-Month relationship

Michael-M-Judd:
- Fixes an accessibility issue with having `role="heading"` on the caption column. Usually heading roles described semantically is the same as h1, h2, etc, so in this case we don't want to break page flow and should use a relationship with aria-describedby instead.

Storybook accessibility audit report:
Before:
Screen Shot 2020-06-23 at 11 54 49 AM

After:
Screen Shot 2020-06-23 at 11 35 49 AM

### DIFF
--- a/src/Caption.js
+++ b/src/Caption.js
@@ -52,7 +52,7 @@ export default class Caption extends Component {
       onClick,
     } = this.props;
     return (
-      <caption
+      <div
         className={classNames.caption}
         id={captionId(date)}
         aria-live="polite"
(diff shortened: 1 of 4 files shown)


gpbl:
Closing as [we've released v8](https://github.com/gpbl/react-day-picker/discussions/1412): this issue belongs to a version which is not longer maintained, and it is likely fixed. In case, please open a new issue. Thanks!

PR #1049, opened 2020-07-01
Minor a11y updates

parkeragee:
- `<Day />` - Removing the role of gridcell because screen readers do not recognize it as a button
- `<DayPicker />` - Adding a few needed `aria` attributes

Resources - https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html

### DIFF
--- a/src/Day.js
+++ b/src/Day.js
@@ -136,9 +136,7 @@ export default class Day extends Component {
         className={className}
         tabIndex={tabIndex}
         style={style}
-        aria-label={ariaLabel}
-        aria-disabled={ariaDisabled}
-        aria-selected={ariaSelected}
+        role="gridcell"
         onClick={handleEvent(onClick, day, modifiers)}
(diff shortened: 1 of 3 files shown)


src/Day.js
```diff
@@ -136,7 +136,6 @@ export default class Day extends Component {
         className={className}
         tabIndex={tabIndex}
         style={style}
-        role="gridcell"
```
liamsc, line 139: @parkeragee do we need a role of `"button"` here if it's supposed to function as one?

parkeragee: Adding a role of button throws a linting error stating that `aria-selected` is not supported by the role "button". So, we'd need to remove that `aria-selected` attribute. I think just removing the `gridcell` role would allow the screen reader to read the content while not introducing a breaking change (removing the aria-selected) attr.

gpbl:

PR #1057, opened 2020-08-24
Fix month prop for DatePickerInput

kaciakmaciak:
The `month` prop is being ignored if used in `DatePickerInput`. According to docs: http://react-day-picker.js.org/api/DayPicker#month if `month` is set the calendar should open with the specified month displayed.

To see the bug in action https://codesandbox.io/s/y70dr


gpbl:
Thanks! This is by design.

PR #1065, opened 2020-10-09
Corrected typing for RangerModifier

joebenn:
Came across a typing issue when using react-day-picker in our Typescript project.

Specifically, making use of the util function `addDayToRange`. The typing for `addDayToRange` is defined as:
`addDayToRange(day: Date, range: RangeModifier): RangeModifier;`. 
where `export interface RangeModifier {
  from: Date;
  to: Date;
}`

However the function `addDayToRange`, implemented in `DateUtils.js`expects (and handles) null date objects (from and to) objects, which contradicts the current typing for `RangeModifier`

I simply changed the typing to allow for the dates to be `undefined` within the `RangeModifier` object, and this solved the typing issues we were experiencing.

I also noticed `RangeModifier` was defined twice, so I removed the duplicate.

Cheers!

### DIFF
--- a/types/Modifiers.d.ts
+++ b/types/Modifiers.d.ts
@@ -1,11 +1,6 @@
 export interface RangeModifier {
-  from: Date;
-  to: Date;
-}
-
-export interface RangeModifier {
-  from: Date;
-  to: Date;
+  from: Date | undefined;
+  to: Date | undefined;
 }
 
 export interface BeforeModifier {


elvorad:
I have the same TypeScript issue and would like to use the change that joebenn has requested – thank you for reaching in this merge request, joebenn! Dear maintainers, would somebody of you please be so kind as to approve this merge request and release a new version?

The tutorial site found under http://react-day-picker.js.org/examples/selected-range has an example in which `from` and `to` are set to `undefined`, yet the current type definition does not support this use case and throws false positives.

types/Modifiers.d.ts
```diff
@@ -1,11 +1,6 @@
 export interface RangeModifier {
-  from: Date;
-  to: Date;
-}
-
-export interface RangeModifier {
-  from: Date;
-  to: Date;
+  from: Date | undefined;
+  to: Date | undefined;
```
mhaidamashko, line 3: If you want set undefined to type property better to use `?` if it's possible.
```
export interface RangeModifier {
  from?: Date;
  to?: Date;
}
```
Nevertheless, I think there should be 
```
export interface RangeModifier {
  from: Date | null;
  to: Date | null;
}
```
Because In realisation of the function `addDayToRange` it has fallback to nulls.
```
export function addDayToRange(day, range = { from: null, to: null }) {}
```

PR #1083, opened 2021-01-25
Wrapper - add the ability to remove the tabIndex

sivanharel:
When we are talking about a calendar without an input there is no need to add a focus on the wrapper element at all. So we need the ability to remove the tabindex property from this element at all.
Thanks!


sivanharel:
🙏

gpbl:

PR #1175, opened 2021-02-28
Removes unnecessary node requirement

lukebatchelor:
Hiya,

Just wanted to make a quick PR to remove the node requirement from this package. Since it's a frontend package there shouldn't be a requirement for a specific version of node to be installed, it can make it a hassle to install.

No worries if you don't want to merge.

Cheers

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,5 @@
 {
   "engines": {
-    "node": "14",
     "yarn": "2"
   },
   "name": "@react-day-picker/monorepo",


gpbl:
Hey thanks – the wrong version was fixed in https://github.com/gpbl/react-day-picker/commit/35db4c7415efa42ecb69036acfaa9a1bb664f10e, this one is for the yarn workspace and it's OK like this :)

beta.17 should contain the fix!

PR #1176, opened 2021-03-04
Selecting multiple days / hooks version

avomakesart:
**What**:

Recently I was working with the library, particularly using the example of Selecting multiple days, and I realized that in the documentation only the component is used with class components, since the requirement I had was only to use hooks, I took on the task to implement the same component but using hooks.

I have just added a new component, do not modify anything in the documentation.

**Why**:

The user may be able to use the component not only with class components also with the new react hooks.


gpbl:

PR #1179, opened 2021-03-10
Reference the commonjs and ESM outputs in the package.json

rzane:
This module ships with a dist folder, but the package.json does not reference any of the build outputs.

On a related note, I wonder if you've considered using a tool like [`microbundle`](https://github.com/developit/microbundle) to simplify the build process? It makes the build process a little less painful, but I can also imagine it being too opinionated for some projects.

### DIFF
--- a/packages/react-day-picker/package.json
+++ b/packages/react-day-picker/package.json
@@ -12,7 +12,8 @@
   "bugs": {
     "url": "https://github.com/gpbl/react-day-picker/issues"
   },
-  "main": "lib/index.js",
+  "main": "dist/index.cjs.js",
+  "module": "dist/index.esm.js",
   "types": "lib/index.d.ts",
   "scripts": {
     "develop": "concurrently --names='tsc-alias,tsc' 'yarn tsc-alias -w' 'yarn tsc -w'",


gpbl:
I will remove the dist files – the ESM output is already in the `lib` – a good reason to need the others?

rzane:
I think so? Node versions that don't support import/export are still pretty widespread. I believe the `main` field is supposed to reference commonjs, while the `module` field is supposed to reference ESM.

I'm on Node 14. Here's what happens when I try to require this package:

```
> require('react-day-picker');
/Users/rzane/Workspace/day-picker/node_modules/react-day-picker/lib/index.js:1
export * from './DayPicker';
^^^^^^

Uncaught SyntaxError: Unexpected token 'export'
```

While I was running my app via webpack, everything worked fine. Webpack transpiled the modules, so that everything worked nicely. However, when I went to run my test suite, I was greeted with the error above.

gpbl:
Thanks @rzane for replaying back, I've been reviewing your report too quickly, sorry for that!

The challenge is to _not use_ a web bundler, and rely just on the TypeScript compiler. But I'm still not sure what use case this choice will cause to explode.

So I've been studying a bit the `package.json` world, and I believe we want an "hybrid package" - that exports both CommonJS and ES modules. This is not that easy, but thanks to [Rauschmayer's hints](https://2ality.com/2019/10/hybrid-npm-packages.html) I think I fixed it in 0f8a75eb23605e0a0d0105f66f92fd6bcda1933c.

I'm deploying to npm this update as beta.20 or 21, please report back if this version fixes it.

gpbl:
Thanks for the help @rzane,`npm install react-day-picker@next` (beta.24) should fix it!

PR #1198, opened 2021-03-29
Update Props.d.ts

shamoh19:
Fix: Typescript Error

DayPicker is not typescript compatible. Was getting error in modifier


ajroberts0417:
Does #1217 solve your problem @shamoh19 ?

gpbl:
I don't think setting `any` is the fix we are looking for. Thanks anyway!

Issue #1205, opened 2021-04-06
Wrong export (v7.4.10)


pustovalov:
`7.4.9`

```
built modules 624 KiB [built]
  modules by path ./node_modules/react-day-picker/build/*.js 66.8 KiB 14 modules
```

```
import DayPicker from 'react-day-picker'

// all works fine
```

`7.4.10`

```
built modules 598 KiB [built]
    ./node_modules/react-day-picker/lib/react-day-picker.min.js 41.1 KiB [built] [code generated]
```

```
import DayPicker from 'react-day-picker'

// an error
```

pustovalov:
Looks like after the `browser` field was added, the webpack started to use `react-day-picker.min.js` source

Used such workaround:

```
module.exports = {
  resolve: {
    mainFields: ['module', 'main', 'browser'],
  },
}
```

maybe this solution will help someone

gpbl:
Closing as [we've released v8](https://github.com/gpbl/react-day-picker/discussions/1412): this issue belongs to a version which is not longer maintained, and it is likely fixed. In case, please open a new issue. Thanks!

Issue #1039, opened 2020-05-26
Wrong export (v 7.4.1 or higher )


pavel-sindelka:
WARNING in ./node_modules/react-day-picker/src/moment.js 1:0-73
"export 'MomentLocalUtils' (reexported as 'default') was not found in './addons/MomentLocaleUtils'
 @ ./node_modules/react-day-picker/build/addons/MomentLocaleUtils.js
 @ ./node_modules/react-day-picker/moment/index.js

andrei10k:
@pavel-sindelka why did you close this? i can still reproduce it wit 7.4.10

PR #1206, opened 2021-04-06
Fix export issue

pustovalov:
I think this should fix:
Closes https://github.com/gpbl/react-day-picker/issues/1205
Closes https://github.com/gpbl/react-day-picker/issues/1039

How webpack uses export keys:
https://github.com/webpack/webpack/issues/4674#issuecomment-639669772

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "Flexible date picker component for React",
   "browser": "lib/react-day-picker.min.js",
   "main": "build/index.js",
+  "module": "build/index.js",
   "types": "types/index.d.ts",
   "style": "lib/style.css",
   "files": [


zivni:
Hi. 

I still get the error, as my setup cannot use the "module" field. So I removed it from my webpack resolve. 

The question is why do you need the "browser" field as those are meant only for bundlers. As I understand "lib/react-day-picker.min.js" is not meant to be used by bundlers but included in a`<script>` tag. 

As you can see from the [spec](https://github.com/defunctzombie/package-browser-field-spec#overview), the "browser" field is to be used in situations where some part of the code is not meant to be used in the browser and some is. e.g. when some parts in the code are meant to work only in node-js server environments. In this case the "browser" field is pointing to the files that doesn't include the server code (like relaying on "fs").

The "browser" field is not meant to point to the  stand alone version. 

As this package is meant to be used only in the browser. It is best to only include the "main" field.

pustovalov:
@zivni, you can try this way:

```
  resolve: {
    mainFields: ['module', 'main', 'browser'],
  },
```

zivni:
Tried it and it didn't work for me as there are packages that use the "browser" field for what it meant to be, so if I change the order, their code with the node specific Apis is getting into the browser bundle.

pustovalov:
I think you can manually resolve dist file of `react-day-picker` in your webpack config and choose what suits you

zivni:
I used this webpack setting to overwrite the "script" field.
```
resolve: {
    alias: {
        "react-day-picker": "react-day-picker/build",
    },
}
```

I still think that it is better to follow the intended use of the fields, instead of making users use such a workaround.

PR #1230, opened 2021-06-08
Fix DateUtils.addDayToRange second argument typing

CyrilQuandalle:
It's to fix the typing on DateUtils.addDayToRange which doesn't accept undefined as second argument whereas it is perfectly able to handle it.

### DIFF
--- a/types/DateUtils.d.ts
+++ b/types/DateUtils.d.ts
@@ -1,7 +1,7 @@
 import { RangeModifier } from './Modifiers';
 
 export const DateUtils: {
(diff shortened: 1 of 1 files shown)


gpbl:
Thanks!

CyrilQuandalle:
> Thanks!
You're welcome ;-)

lloydstubber:
Hi there, I see this has been fixed but I can't seem to see it in the latest tagged version for v7. Could this please be released? Thanks

PR #1238, opened 2021-06-30
Update Navigation.tsx

brentvb:
This change fixes the issue that the 'nav_button_previous' className is added to both nav buttons instead of only the previous button.


gpbl:
Good catch, thanks. I was sure I had tests wrote for this :(

PR #1240, opened 2021-07-11
Allow Head to be replaced

Itherael:
Currently `Head` cannot be replaced with a custom component as per [documentation](https://react-day-picker-next.netlify.app/api/interfaces/components#head). This PR allows a custom `Head` component to be used.


tjfred35 (approved):
LGTM. This is the only component in the Components interface that doesn't follow the pattern

gpbl (approved):

PR #1262, opened 2021-10-15
v7: Added support for MouseEnter and MouseLeave events on WeekNumber

gabsoftware:
I added support for the MouseEnter and MouseLeave events for week numbers. This can be useful to pre-select the week days on hovering the week number.

A little GIF is better than a thousand words: https://gyazo.com/9db1f69153293697f96465e3f5f9bf1f


gpbl:

Issue #1250, opened 2021-08-14
(v.8) Broken day layout for January for locales where the first day of week is Monday


dmitriifrlv:
**Describe the bug**
Day layout breaks for January(not for every year though but it was noticed for January 2021, 2022, 2023 ) for locales where the first day of the week is Monday(for several months)
<img width="388" alt="Screenshot 2021-08-14 at 20 39 16" src="https://user-images.githubusercontent.com/65853205/129451560-0426fbbf-fabc-4db9-b6b7-ef5782475d49.png">

**To Reproduce**
Clone [this CodeSandbox](https://codesandbox.io/s/broken-day-layout-9f9l2?file=/src/App.js) to reproduce the bug.

**Expected behavior**
The layout where the first days of the month are on the first row.

<img width="413" alt="Screenshot 2021-08-14 at 20 44 42" src="https://user-images.githubusercontent.com/65853205/129451716-e056804a-7081-4bec-9850-d80bded88fd9.png">

Gertjan1996:
@dmitriifrlv did you find a workaround for this problem?

tjfred35:
Oh, interesting...the week number is 52
<img width="380" alt="Screen Shot 2021-11-02 at 7 49 28 AM" src="https://user-images.githubusercontent.com/2838338/139840895-d857580d-d380-4251-9535-bf059dd3ec92.png">

tjfred35:
I've got a fix here: https://github.com/gpbl/react-day-picker/pull/1270. Would appreciate some review.

Issue #1222, opened 2021-05-17
v8. Days ordering for locales where week starts on Monday (January 2021-23)


alesvojta:
**Describe the bug**
When using locale option where week starts at Monday (eg cs or fr), dates for beginning of the month are aligned at the end of the table for January 2021-2023.

**To Reproduce**
https://codesandbox.io/s/daypicker-locale-week-starts-monday-85bb8

**Expected behavior**
Days are aligned in correct order 1, 2, 3, 4, 5, ..., 28, 29, 30, 31.

**Screenshots**
![image](https://user-images.githubusercontent.com/1668328/118465814-8ee32300-b702-11eb-9e87-c4ed0625508f.png)

**Additional context**
Used beta 29 but experiencing it since alpha 45.

brentvb:
As a workaround you can set the firstWeekContainsDate property of your locale to 1. As far as I am aware the only downside of the workaround would be that you can't show the correct weekNumber according to your locale.

tjfred35:
I think this might be the same root cause as https://github.com/gpbl/react-day-picker/issues/1207

tjfred35:
Yeah same issue
<img width="361" alt="Screen Shot 2021-11-02 at 7 54 09 AM" src="https://user-images.githubusercontent.com/2838338/139841503-5f15d616-47c1-4d80-af6f-523ada1cca81.png">

tjfred35:
I've got a fix here: https://github.com/gpbl/react-day-picker/pull/1270. Would appreciate some review.

Issue #1207, opened 2021-04-07
v8. Markup is broken with "showOutsideDays" and "fixedWeeks" options in December 2022


jodaka:
**Describe the bug**
Layout is broken with `"react-day-picker": "8.0.0-beta.29"` under certain conditions. 

**To Reproduce**
reproduced here: 
https://codesandbox.io/s/daypicker-basicscustomization-weeknumber-example-forked-7ej5b?file=/example.tsx
using next month button scroll to December 2022. While other months are rendered just fine, this one brakes layout

**Screenshots**
![image](https://user-images.githubusercontent.com/83644/113855575-dc39af80-97a8-11eb-8fd1-3c79a9a8d34f.png)

**Additional context**
used beta 29 (as 30 doesn't seemed to be published to npm at the moment of writing)

roblotter:
<img width="867" alt="Screenshot 2021-05-19 at 14 29 32" src="https://user-images.githubusercontent.com/50020244/118805590-c1318380-b8ae-11eb-9299-bec8763304f6.png">

I can confirm that with the same version `8.0.0-beta.29`

I'm using `showOutsideDays=true` and `fixedWeeks=true`. You can see in the screenshot above, the row contains 14 cells for days, instead of just 7. Basically it should not try to show the next week.

roblotter:
For now I worked around this one with the following css:

```css

    .rdp-row {
      td:nth-child(8),
      td:nth-child(9),
      td:nth-child(10),
      td:nth-child(11),
      td:nth-child(12),
      td:nth-child(13),
      td:nth-child(14) {
        display: none;
      }
    }
```

roblotter:
Also had to add this:

```css
rdp-with_weeknumber .rdp-row td:nth-child(8) {
   display: table-cell;
}
```
to make week numbers show correctly

tjfred35:
In the code in getWeeks.ts, the following lines (two places) can be removed. The problem is that the 25th - 31st of Dec 2022 are _already_ week 53 -- this logic seems to want to make week 1 into (pseudo-) week 54, but I don't see why it shouldn't just be week 1 -- it would be simpler that way, and fix this bug.

```
if (week === 1 && getMonth(date) === 11) {
   week = 53;
}
```

tjfred35:
That's the issue...but after review of https://github.com/gpbl/react-day-picker/issues/1222 and others, I now understand why the code above exists...since we iterate over an object whose keys are the week numbers from date-fns
[...]

PR #1270, opened 2021-11-02
(fix) Week numbers may break at end/begin of the years

tjfred35:
Fixes some weird edge cases where weeks at the beginning or end of the year could render in the wrong order, or share the same table row.

This fixes the following issues:
https://github.com/gpbl/react-day-picker/issues/1207
https://github.com/gpbl/react-day-picker/issues/1222
https://github.com/gpbl/react-day-picker/issues/1250

The problem came from using week number in a dictionary, because at the beginning and end of years, the natural order of a date-fns week number is not the order that we want to render the weeks in.

So instead of returning an object, we return an array of a new MonthWeek object from getWeeks function, to maintain order. Furthermore, we remove the previous attempt to fix this ordering problem that was the source of one of the above issues where two weeks would render into one row. By removing this week transformation code, we can reduce the overall complexity of what a week number is.


packages/react-day-picker/src/components/Table/Table.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+const po = new PageObjects(new Date());
+
+describe('It should render weeks at the boundaries of years correctly', () => {
+  test('If the first days of the month start at week 52 from the previous year, week 52 should be rendered on the 1st row', () => {
+    customRender(<Table displayMonth={new Date(2022, 0)} />, { locale: enGB });
+    const week1 = po.weeks[0].querySelectorAll('td');
+    const day6 = week1[5];
+    const day7 = week1[6];
+    expect(day6).toHaveTextContent('11st January (Saturday)');
+    expect(day7).toHaveTextContent('22nd January (Sunday)');
```
tjfred35, line 17: I would love suggestions on how to structure these tests better -- right now the text content is from an aria-hidden element (`"2"`) and hidden element that is presumably read by screen readers (`"2nd January (Sunday)"`) -- I don't know to what extend that construct in this v8 beta is accessible.

packages/react-day-picker/src/components/Table/Table.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  test('If the last week of the year is 1, it should be rendered on the 5th row', () => {
+    customRender(<Table displayMonth={new Date(2021, 11)} />, {
+      locale: enUS
+    });
+    const week5 = po.weeks[4].querySelectorAll('td');
+    const day29 = week5[0];
+    const day30 = week5[1];
+    expect(day29).toHaveTextContent('2626th December (Sunday)');
+    expect(day30).toHaveTextContent('2727th December (Monday)');
```
tjfred35, line 41: This test passes with the previous code because of this code:
```
if (week === 1 && getMonth(date) === 11) {
  week = 53;
}
```
With the proposed code changes, the above code has been removed, but the test still passes, and the other two edge case tests do too.

However, removing that code above now renders the last month of the year with a week number of one (see screenshot):
Screen Shot 2021-11-02 at 3 09 22 PM

I am making the assumption that the reason for the code above was to fix the original ordering bug, not to apply any desired transformations to the week numbers.

The week number, 1, without the transformation to map it to 53, is now consistent between the December 2021 and the January 2022 calendars, so I think it's a reasonable change of behavior to the rendering of the week numbers that also fixes these other edge cases. Additionally, making week number transformations significantly increases the complexity of the definition of a week number, so removing the transformation is an improvement as far as reducing complexity is concerned.

packages/react-day-picker/src/components/Table/utils/getWeeks.ts
```diff
@@ -14,51 +13,44 @@ import { getOutsideEndDays } from './getOutsideEndDays';
 import { getOutsideStartDays } from './getOutsideStartDays';
 
 /**
- * The weeks belonging to a month. Each key of the returned object is the
- * week number of the year.
+ * The date-fns week and the dates for that week
  */
-export type MonthWeeks = { [weeknumber: string]: Date[] };
-
+export type MonthWeek = { weekNumber: number; dates: Date[] };
```
tjfred35, line 18: Since this type and the `getWeeks` function were previously publicly exported, this change is a breaking change, but I assumed that is ok because we are still in beta.

packages/react-day-picker/src/components/Table/Table.test.tsx
UlyssesInvictus: Is there a reason the text of the actual rendered dates (the non-hidden elements) isn't preferred here? Are they not actually visible?

Also, is this react testing library?

packages/react-day-picker/src/components/Table/Table.test.tsx
UlyssesInvictus: As long as it's documented in the (eventual) library docs, I don't see that as a breaking change. It's a lot easier to understand "week # = latest week # within this 7 day period" than the edge case of the end of the year.

The only thing that gives me pause is that a week 53 is extremely rare but technically possible. However, the number 53 isn't special in itself. Are there any known use cases where someone has to detect whether something is the last week, and is currently reliant on looking for a hardcoded value of 52/53 ?

packages/react-day-picker/src/components/Table/Table.test.tsx
melinayoung-toast: In addition to these tests, have you thought about including snapshot tests? They can be used to verify that the behavior doesn't change in the future https://jestjs.io/docs/snapshot-testing

jericopingul-toast: One way to select the cell in a particular role is something like
```js
const table = screen.getByRole('table')
const firstBodyRow = within(table).getAllByRole('row')[2]
within(firstBodyRow).getByRole('cell', {
  name: /3rd january \(monday\)/i
})
```
Since there are a lot of nesting `within` can be useful:
https://testing-library.com/docs/dom-testing-library/api-within/

jericopingul-toast: This might also work if you want to select a row with a matching date
```js
const table = screen.getByRole('table')
within(table).getByRole('row', {
  name: /3rd january \(monday\)/i
})
```
The testing playground is a pretty nice tool to play around with selecting elements and gives recommendations
https://testing-library.com/docs/queries/about/#screenlogtestingplaygroundurl

tjfred35: Yes -- but then how to assert that particular element is in the right position in the grid?

jericopingul-toast: Good point, can we use something like the the row/col indexes?
```js
const table = screen.getByRole('table')
const firstBodyRow = within(table).getAllByRole('row')[2]
const targetDate = within(firstBodyRow).getAllByRole('cell')[0]
```
There might be a better way than this...

packages/react-day-picker/.eslintrc
```diff
@@ -4,6 +4,9 @@
     "ecmaVersion": 2018,
     "sourceType": "module"
   },
+  "env": {
+    "jest": true
+  },
```
gpbl, line 9: There's already an override for the test files. This should not be needed.

```suggestion
```

packages/react-day-picker/src/components/Table/utils/getWeeks.ts
```diff
@@ -14,51 +13,44 @@ import { getOutsideEndDays } from './getOutsideEndDays';
...
+  const weeks: MonthWeek[] = [];
+  let lastWeek: MonthWeek = { weekNumber: -1, dates: [] };
   for (let i = 0; i <= diff; i++) {
     const date = addDays(monthStart, i);
-    let week = getWeek(date, { locale });
-    if (week === 1 && getMonth(date) === 11) {
-      week = 53;
-    }
-    const weekStr: string = week.toString();
+    // const year = getYear(date, { locale })
```
gpbl, line 35:
```suggestion
```

packages/react-day-picker/src/components/Table/Table.test.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react';
+
+import { enGB, enUS } from 'date-fns/locale';
+import { customRender, PageObjects } from 'test';
+
+import { Table } from './Table';
+
+const po = new PageObjects(new Date());
```
gpbl, line 8: We need to freeze date with timekeeper: check out how the test is implemented here: https://github.com/gpbl/react-day-picker/blob/95ec208ccbf97f4c440a196c24c809219d5d7bd3/packages/react-day-picker/src/components/Caption/Caption.test.tsx#L17-L18

packages/react-day-picker/src/components/Table/Table.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+describe('It should render weeks at the boundaries of years correctly', () => {
```
gpbl, line 10: I'm not sure this is the test where we should test this behavior. I would add a test to `getWeeks` instead.

packages/react-day-picker/src/components/Table/Table.test.tsx
gpbl: @tjfred35 are you sure from the screen above the week numbers are correct? For December 2021 , my calendar app shows 48–1 for (week starting from Monday) or 47–52 (week starting from Sunday). But never the first week is 49. We should double check this...

packages/react-day-picker/src/components/Table/utils/getWeeks.ts
```diff
@@ -14,51 +13,44 @@ import { getOutsideEndDays } from './getOutsideEndDays';
...
+): MonthWeek[] {
   const monthStart = startOfMonth(month);
   const monthEnd = endOfMonth(month);
 
   const diff = differenceInCalendarDays(monthEnd, monthStart);
 
-  const weeks: MonthWeeks = {};
-  let lastWeekStr = '';
+  const weeks: MonthWeek[] = [];
+  let lastWeek: MonthWeek = { weekNumber: -1, dates: [] };
```
gpbl, line 32: 🤔 this shouldn't be:

```suggestion
  let lastWeek: MonthWeek = { weekNumber: 0, dates: [] };
```

gpbl (commented):
Hey thanks for your contribution! Week numbers are a beast. They depends from the first day of the week or the locale.

- are the week number still correct? My calendar app says other numbers but it maybe because locale / first day of week
- align the test of `Table.tsx` to our pattern: for example, [Caption.test.tsx](https://github.com/gpbl/react-day-picker/blob/95ec208ccbf97f4c440a196c24c809219d5d7bd3/packages/react-day-picker/src/components/Caption/Caption.test.tsx)
  - in particular make sure to freeze the time
- i think if we change `getWeeks` we must add a test for that file instead of testing table.

packages/react-day-picker/src/components/Table/utils/getWeeks.ts
gpbl: I think it is a pretty nice solution. I wasn't sure at first about the `Object`, which i think I choose for performance reasons.

packages/react-day-picker/src/components/Table/Table.test.tsx
tjfred35: Week numbers are wild, I guess! On the published beta in codesandbox I can get the following

**locale={enUS}**
Screen Shot 2021-11-04 at 8 34 40 AM

**locale={enGB}**
Screen Shot 2021-11-04 at 8 37 53 AM

https://codesandbox.io/s/daypicker-basics-customization-weeknumber-example-forked-k9o5n

What locale / configuration do you use to get 47-52?

packages/react-day-picker/src/components/Table/Table.test.tsx
tjfred35: Good to know. I ended up moving the tests to getWeeks.test.ts where they got much simpler -- no react necessary!

packages/react-day-picker/src/components/Table/Table.test.tsx
tjfred35: Good idea. That cleaned things up considerably

packages/react-day-picker/src/components/Table/utils/getWeeks.ts
tjfred35: I wanted this initial state to be null, but then I didn't want to deal with null later -- this initial state is never added to the list, and so I wanted to be good and sure the week number was invalid. I changed it to zero, based on your suggestion, but I have assumed zero is never valid.

tjfred35 (commented):
Thanks for the review @gpbl. I think this makes things much cleaner, at least as far as the tests are concerned.

packages/react-day-picker/src/components/Table/utils/getWeeks.test.ts
```diff
@@ -0,0 +1,33 @@
+import { enGB, enUS } from 'date-fns/locale';
+import { getWeeks } from './getWeeks';
+
```
gpbl, line 3: We need to "freeze the time" here as well (like in `Caption` with the tk package).

packages/react-day-picker/src/components/Table/utils/getWeeks.test.ts
```diff
@@ -0,0 +1,33 @@
...
+describe('It should render weeks at the boundaries of years correctly', () => {
+  test('If the first days of the month start at week 52 from the previous year, week 52 should be rendered on the 1st row', () => {
```
gpbl, line 5: We should update the tests to follow this format:

```ts
describe('when the first days of the month start at week 52 from the previous year', () => {
  /* some setup here */
  test('the 1st row should render 52', () => {
    expect(); // etc
```

gpbl (changes requested):
LGTM - could we improve the test as suggested? Eg the timekeeper. 

Thanks!

packages/react-day-picker/src/components/Table/utils/getWeeks.test.ts
tjfred35: I'll try to have just one expect per test

packages/react-day-picker/src/components/Table/utils/getWeeks.test.ts
tjfred35: I think we may only need to freeze time in the case that we are testing code that uses `new Date()`.

packages/react-day-picker/src/components/Table/Table.test.tsx
gpbl: > What locale / configuration do you use to get 47-52?

Interesting. The "Region" in my Mac OS is set to Spain. Indeed the week number changes. If there are issues we will find them out but it looks like it is expected :)

packages/react-day-picker/src/components/Table/utils/getWeeks.ts
gpbl: Yeah maybe a reduce there would have worked better.

packages/react-day-picker/src/components/Table/utils/getWeeks.test.ts
gpbl: Would effects happen even if the `new Date()` is used in the test? Anyway thanks for the heads up!

packages/react-day-picker/src/components/Table/utils/getWeeks.test.ts
gpbl: Yeah! We should improve the test to be more descriptive, we can do that during some clean-up…

gpbl (approved):

PR #1274, opened 2021-11-04
Fix range selection selected modifiers

tjfred35:
Additionally, add a missing focused state for selected days to the default stylesheet and fix the two navigable selecting-days-range doc files to allow for de-selecting a range without creating errors

### DIFF
--- a/packages/react-day-picker/src/hooks/useModifiers/useModifiers.ts
+++ b/packages/react-day-picker/src/hooks/useModifiers/useModifiers.ts
@@ -49,9 +49,6 @@ export function useModifiers(date: Date): UseModifiers {
     modifiers.range_start = rangeSelect.modifiers.range_start ?? [];
     modifiers.range_middle = rangeSelect.modifiers.range_middle ?? [];
     modifiers.range_end = rangeSelect.modifiers.range_end ?? [];
-    modifiers.disabled = modifiers.disabled.concat(
-      rangeSelect.modifiers.disabled ?? []
-    );
   }
 
   const status = getModifierStatus(date, modifiers);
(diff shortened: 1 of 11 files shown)


gpbl (commented):
Thanks, before reviewing... 

I am sure testing-library has an easier way to test a hook, or we should write a small util to reduce the boilerplate. 

Here (and in fee other PRs) we should simplify the tests before merging. I will push something soon about this...

tjfred35:
Thanks for the pre-review review :). I found a nice react-hooks testing library that makes the tests much cleaner.

packages/react-day-picker/src/hooks/useModifiers/useModifiers.ts
```diff
@@ -41,17 +41,14 @@ export function useModifiers(date: Date): UseModifiers {
     );
   } else if (isDayPickerRange(context)) {
     modifiers.selected = modifiers.selected.concat(
-      multipleSelect.modifiers.selected ?? []
+      rangeSelect.modifiers.selected ?? []
```
gpbl, line 44: That was the bug ?

website/docs/basics/selecting-days-range.tsx
```diff
@@ -10,13 +10,17 @@ export default function Example() {
...
+  const [range, setRange] = React.useState<DateRange>(defaultSelected);
+
+  console.log(range);
 
   let footer = 'Please pick the first day.';
-  if (range.from && !range.to) footer = 'Please pick the last day.';
-  if (range.from && range.to)
+  if (range && range.from && (!range.to || isSameDay(range.from, range.to))) {
+    footer = format(range.from, 'PPP');
+  } else if (range && range.from && range.to) {
```
gpbl, line 20: 💭  Should we make easier to implement this with a small utility?

gpbl (approved):
Cool thanks!

PR #1275, opened 2021-11-04
(fix) useDay `isOutside` is always true

tjfred35:
This is useful for custom implementations that use the useDay hook. For example, my custom day component uses:

```
  const isRangeStart = day.modifiers.range_start
  const isRangeMiddle = day.modifiers.range_middle
  const isRangeEnd = day.modifiers.range_end
  const isSelected = day.modifiers.selected
  const isToday = day.modifiers.today
  const isDisabled = day.modifiers.disabled
  const isOutside = day.isOutside
```

except mine uses a workaround:
```
  // TODO: Fix this in useDay
  // const isOutside = day.isOutside
  const isOutside = !isSameMonth(props.date, props.displayMonth)
```


gpbl (approved):
Thanks, here as well we could use the custom render from #1274...

tjfred35:
Yep. Will update the test to use the testing library we introduced.

gpbl:
Thanks! (I've updated the test a bit in 6a2f4a0)

tjfred35:
Thanks! I like `toBe(true)` better than `toBeTruthy()`.

Issue #1245, opened 2021-07-17
Is there a way to unselect a day without user interaction


ablajievschii:
### Discussed in https://github.com/gpbl/react-day-picker/discussions/1234

<div type='discussions-op-text'>

<sup>Originally posted by **ablajievschii** June 11, 2021</sup>

Hi all, tried to get an answer in discussion but no help there. Hope I can get an answer here.

I'm using DayPickerInput in my React application and I need to unselect a day without user interaction. I've already tried setting selectedDay to undefined, but it does not work. Setting input element value to empty string does not work either.
I know about clickUnselectsDay, but how to unselect without user interaction ? Is there a way to simulate a click or maybe use some public function on DayPickerInput instance ?</div>

tjfred35:
I was going to suggest that you should use the "controlled" mode where you pass in `selected`...

```
export default function Example() {
  const [days, setDays] = React.useState([]);

  const footer =
    days.length > 0
      ? `You picked ${days.length} day(s).`
      : `Please pick one or more days.`;

      const removeLastSelected = () => {
        setDays(days.slice(0, days.length - 1))
      }

  return (
    <div>
      <DayPicker mode="multiple" selected={days} onSelect={setDays} footer={footer} />
      <button disabled={days.length === 0} onClick={removeLastSelected}>Remove most recent selection</button>
    </div>
  );
}
```

...but it doesn't work! Maybe there is no true "controlled" mode, because it's blending the selected prop here with the internal selected state...

We should make it so that if `selected` is not undefined then we are in a "controlled" mode and the internal state is not used.

tjfred35:
Ok, I have a solution for this:

https://github.com/gpbl/react-day-picker/pull/1276

tjfred35:
@ablajievschii This is now possible in the v8 beta. I think your comments are about v7, so going to change the labels to match that assumption.

gpbl:
Closing as [we've released v8](https://github.com/gpbl/react-day-picker/discussions/1412): this issue belongs to a version which is not longer maintained, and it is likely fixed. In case, please open a new issue. Thanks!

PR #1276, opened 2021-11-04
(feat) Allow fully "controlled" range and multiple modes

tjfred35:
I’m pretty sure this addresses the feature request here:

https://github.com/gpbl/react-day-picker/issues/1245

This change is to allow `selected` prop to be used for all the modes as a way for the state to be externally managed (“controlled”). This allows users to manipulate this state “without user interaction”.

Additionally, I’m hoping we can rename mode="uncontrolled" to mode="custom" -- this seems to better express how the documentation already refers to this mode, and doesn’t get mixed up with the other common definition in react component libraries like downshift and even elsewhere in react-day-picker:

- controlled means the user passes in the state
- uncontrolled means the component manages the state for the user, but the user can pass in some default state

This is the definition that react day picker uses for months. From the docs:

> To control the current month, use `month` (as opposed to `defaultMonth`) and `onMonthChange` to handle the current month.

I'm still looking into the implementation of this controlled pattern for month (useNavigationState.ts) -- it's a little different from the controlled implementation I’ve written for `selected`, so let's discuss which one we prefer. I will look into the downshift implementation too.

 And I haven't implemented this fully yet -- need to implement this same pattern for `mode='single'` for completeness.


tjfred35:
@gpbl Ok this is ready for a review -- I've looked at downshift implementation of its "controlled" props:

https://github.com/downshift-js/downshift/tree/a6bcd2247f4757b9a6c4077fdd9cf07554049054/src/hooks/useSelect#control-props

It's much more complicated, but it ultimately boils down to the same thing -- when the user passes in the `selectedItem` and other props, it uses those in lieu of its own internal state.

gpbl (commented):
Thanks for your work! That `useSelect` in each provider is a red flag. I think we need to review or refactor something first – let keep this on hold while I review the codebase :)

tjfred35:
The dynamic resolution of types based on the value of mode is really interesting, but also complex enough that it’s beyond my technical ability to consolidate these three methods so I’m interested to see if you can come up with a solution

tjfred35:
Also, I just realized that the examples I wrote for the documentation were broken. I just fixed them.

gpbl:
> The dynamic resolution of types based on the value of mode is really interesting, but also complex enough that it’s beyond my technical ability

Well that's not good 😅 Let's block for a moment this and review the architecture.

packages/react-day-picker/src/contexts/SelectMultiple/SelectMultipleProvider.tsx
```diff
@@ -96,3 +90,27 @@ export function SelectMultipleProvider({
...
+    Date[] | undefined
+  >(initialSelected || undefined);
+
+  const selectedDays =
+    controlledSelectedDays === undefined
+      ? uncontrolledSelectedDays
+      : controlledSelectedDays;
+
+  return [selectedDays, setSelectedDays];
+};
```
tjfred35, line 116: I’m going to try something like this:
```
type ControllableProps<T> {
  prop?: T
  defaultProp?: T
}

…
const selectedProps: ControllableProps<Date[]> = isDayPickerMultiple(initialProps)
  ? { prop: selected, defaultProp: defaultSelected }
  : { }
  
const [selectedDays, setSelectedDays] = useControllablePropsState(selectedProps)
```

tjfred35:
@gpbl Ok, I was able to extract a hook that encapsulates the logic that was previously copy / pasted across three components. Does this unblock this from moving forward?

packages/react-day-picker/src/types/DayPickerBase.ts
```diff
@@ -193,12 +193,12 @@ export interface DayPickerBase {
   /**
    * Toggle the controlled selection mode.
    *
-   * - `uncontrolled`: disable the controlled selection mode
+   * - `custom`: customize what is selected
```
gpbl, line 196: To investigate - what should be the default mode?

gpbl (approved):
Thanks I could check this out better, looks like it is laying out a better foundation.

I have some feedback for the `useControllableProp` hook. Will open a PR soon.

PR #1282, opened 2021-11-16
(package) Add eslint-plugin-react-hooks

tjfred35:
I found a couple issues with the new docusaurus updates:

1. There is a hook in website/src/pages/render.tsx (useLocation) that React wants us to hoist to the top of the `Render` function. It spits out an error in the javascript console
2. The svg path for the previous and next icons has no height and width and does not render inside an iframe in the latest version of chrome (see screenshot)
Screen Shot 2021-11-16 at 3 36 40 PM

I think we might be able to use the following plugin to catch the first issue. It certainly caught a few issues in the react-day-picker package, which I've fixed below.

I have no idea how to fix the second issue. It works fine in firefox

### DIFF
--- a/packages/react-day-picker/.eslintrc
+++ b/packages/react-day-picker/.eslintrc
@@ -8,6 +8,7 @@
   "extends": [
     "eslint:recommended",
     "plugin:@typescript-eslint/recommended",
+    "plugin:react-hooks/recommended",
     "prettier",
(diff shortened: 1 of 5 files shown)


gpbl (approved):
Thanks. I've seen too a weird behavior with the arrows after the update, will check it out!

PR #1283, opened 2021-11-21
(chore) Refactor `useControllablePropState`

gpbl:
This PR refactors `useControllablePropState`:

- renamed `useControllablePropState` to `useControlledValue`
- updated its signature – by using two arguments we can fix naming inconsistencies
- added a test: hopefully I understood how it is intended to work, @tjfred35 would you give a look [to the test](https://github.com/gpbl/react-day-picker/blob/5ddbfdde4f9272e0fecd1a2a5f96d858e291bef0/packages/react-day-picker/src/hooks/useControlledValue/useControlledValue.test.ts)?
- the return type is now casted as `<T>`. It helps reducing the type-checking in the components – still, it is left to the implementer to make sure these types are valid. I don't like this much, but I couldn't find a better solution.


packages/react-day-picker/src/contexts/SelectMultiple/SelectMultipleProvider.tsx
```diff
@@ -22,14 +21,23 @@ export function SelectMultipleProvider({
...
-  const [selectedDays, setSelectedDays] = useControllablePropState(
-    isDayPickerMultiple(initialProps)
-      ? {
-          value: initialProps.selected,
-          defaultValue: initialProps.defaultSelected
-        }
-      : { defaultValue: undefined }
-  );
+  const [selectedDays, setSelectedDays] = useControlledValue<
+    Date[] | undefined
```
tjfred35, line 25: I think I might know a way to fix this. We can’t conditionally call a hook, but we can conditionally call a subcomponent for which typescript is confident about the narrowing type checked by isDayPickerMultiple. At that point the values of the parameters to useControlledValue will be known to typescript

packages/react-day-picker/src/hooks/useControlledValue/useControlledValue.ts
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+
+export type DispatchStateAction<T> = React.Dispatch<React.SetStateAction<T>>;
+
+/**
+ * Helper hook for using controlled/uncontrolled values from a component props.
+ *
+ * When a component is controlled, pass the controller value and use the
+ * returned setter to update it.
```
tjfred35, line 9: The tests look good, but this comment doesn’t match my expectation. The setter does nothing in the controlled state.

packages/react-day-picker/src/contexts/Navigation/useNavigationState.ts
```diff
@@ -27,7 +28,6 @@ export function useNavigationState(): [
 
   // Update month if updated from context.
   React.useEffect(() => {
-    // TODO: return void also if `context.defaultMonth`?
```
tjfred35, line 30: I removed this React.useEffect, because it's no longer necessary to use the internal state if it is controlled. Instead we just ignore the internal state

packages/react-day-picker/src/contexts/SelectMultiple/SelectMultipleProvider.tsx
```diff
@@ -22,14 +21,23 @@ export function SelectMultipleProvider({
...
+  };
+
+  if (!isDayPickerMultiple(initialProps)) {
+    const contextValue = { selected: undefined, modifiers };
+    return (
+      <SelectMultipleContext.Provider value={contextValue}>
+        {children}
+      </SelectMultipleContext.Provider>
+    );
+  }
```
tjfred35, line 40: This is what I was talking about in my earlier comment:
```
export function SelectMultipleProvider({
  initialProps,
  children
}: {
  initialProps: DayPickerProps;
  children: React.ReactNode;
}): JSX.Element {
  if (!isDayPickerMultiple(initialProps)) {
    return (
      <SelectMultipleContext.Provider value={EMPTY_SELECT_MULTIPLE_CONTEXT}>
        {children}
      </SelectMultipleContext.Provider>
    );
  }
  return <SelectMultipleProviderInternal
    initialProps={initialProps}
    children={children}
  />
}

const EMPTY_SELECT_MULTIPLE_CONTEXT = {
  selected: undefined,
  modifiers: {
    selected: [],
    disabled: []
  }
}

function SelectMultipleProviderInternal({
  initialProps,
  children
}: {
  // vvv This is a safe cast because of type predicates in isDayPickerMultiple
  initialProps: DayPickerBase & DayPickerMultiple;
  children: React.ReactNode;
}): JSX.Element {
  // vvv It means we don't need the unsafe cast here:
  const [selectedDays, setSelectedDays] = useControlledValue(
    initialProps.defaultSelected, initialProps.selected);

  const handleDayClick: DayClickEventHandler = (day, modifiers, e) => {
    initialProps.onDayClick?.(day, modifiers, e);

```

tjfred35: I'll put up a PR

tjfred35: Here's a complete view compared to master: https://github.com/gpbl/react-day-picker/pull/1285

tjfred35: Here's a view of just my refactor: https://github.com/gpbl/react-day-picker/pull/1286

packages/react-day-picker/src/contexts/Navigation/useNavigationState.ts
tjfred35: Is there a reason that you added it back, or perhaps it was just a merge commit that added it back

packages/react-day-picker/src/hooks/useControlledValue/useControlledValue.ts
gpbl: Thanks! I've updated it in https://github.com/gpbl/react-day-picker/pull/1283/commits/c5b562c688f3fe3e02bfdae6010110d81e0952ee

packages/react-day-picker/src/contexts/Navigation/useNavigationState.ts
gpbl: > Is there a reason that you added it back, or perhaps it was just a merge commit that added it back

Are we branching from the same master? I see that line added by me 8 months ago :)

But anyway you are right it can be safely removed - and if not, it is because we are missing the test :/

gpbl: Removed in 29043dc2dce92fe450b215e60271a5e532d43483.

packages/react-day-picker/src/contexts/SelectMultiple/SelectMultipleProvider.tsx
gpbl: Thanks will follow up there!

PR #1286, opened 2021-11-22
chore: stronger type-safety for useControlledValue

tjfred35:
This PR adds internal provider components that can take advantage of the type predicates to safely narrow the typescript types so that they can pass in known types to the useControlledValue hook, instead of using unsafe casts.


gpbl:
Could we reduce the diff to only the `useControlledValue` change? Because I am not sure it works.

PR #1290, opened 2021-11-24
(chore) Refactor DayPickerProps types, remove uncontrolled selections

gpbl:
We have been dealing with the "controlled" or "uncontrolled" behavior but I don't think it is necessary in case of selections –  by looking to our examples, we always need a "controlled" mode. This however was a bit confusing with the current typings of the props.

* Renamed `DayPickerBase` to `DayPickerProps` and made the four interfaces for the selection modes (`DayPickerSingleProps`,`DayPickerMultipleProps`, etc.) extend `DayPickerProps` itself.
* Removed the `defaultSelected` prop.


tjfred35 (commented):
I like the direction this is going in 👍

packages/react-day-picker/src/types/DayPickerProps.ts
```diff
@@ -1,15 +1,240 @@
...
+  /**
+   * A map of labels creators used for the ARIA labels attributes.
+   */
+  labels?: Partial<Labels>;
+
+  /**
+   * The text direction of the calendar. Use `ltr` for left-to-right (default)
+   * or `rtl` for right-to-left.
+   */
+  dir?: string;
```
tjfred35, line 215: Could this be:
```
dir?: 'ltr' | 'rtl'
```

gpbl: I'd keep the HTML default for the HTMLDivElement which is `string`.

tjfred35: Ok fair

tjfred35:
I can’t think of a reason to use defaultSelected, but if I can we can always add it back. I do have a reason to use defaultMonth.

gpbl:
> I can’t think of a reason to use defaultSelected, but if I can we can always add it back. I do have a reason to use defaultMonth.

Sure, @tjfred35 `defaultMonth` is still there.

tjfred35:
Is “being able to play with DayPicker out of the box without wiring up an external React.useState” a reason to continue to support an uncontrolled mode?

website/examples/navigation-disabled.tsx
```diff
@@ -1,8 +1,13 @@
...
+import { DayPicker } from "react-day-picker";
 
 export function Example() {
   return (
-    <DayPicker fromYear={2015} toYear={2025} disableNavigation mode="single" />
+    <DayPicker
+      fromYear={2015}
+      toYear={2025}
+      disableNavigation
+      mode="single"
```
tjfred35, line 10: Here is an uncontrolled example that may no longer work as intended

website/examples/customization-fixedweeks.tsx
```diff
@@ -2,5 +2,5 @@ import React from "react";
 import { DayPicker } from "react-day-picker";
 
 export function Example() {
-  return <DayPicker showOutsideDays fixedWeeks />;
+  return <DayPicker fixedWeeks />;
```
tjfred35, line 5: I thought fixedWeeks also required showOutsideDays. Did you change that?

website/docs/guides/custom-selections.md
```diff
@@ -1,17 +1,14 @@
...
-to custom selection mode or – to keep the built-in modes – use a custom Day component.
+To implement a selection behavior by yourself, you can either switch to custom
+selection mode or – to keep the built-in modes – use a custom Day component.
 
 ## Switching to the custom mode
 
-1. set the selection mode to `mode="custom"`
-2. store the selected dates in the parent’s state
-3. use the `selected` and `onDayClick` props
+To switch to custom selection, the `mode` propo to `custom`:
```
tjfred35, line 11: “set the” … “prop”…

website/examples/custom-selection-single.tsx
```diff
@@ -1,19 +1,18 @@
...
-    : "Please pick a day.";
+  let footer = "Please pick a day.";
+  if (selected) {
+    footer = `You picked ${selected.toLocaleDateString()}.`;
+  }
   return (
     <DayPicker
-      mode="custom"
-      selected={selectedDay}
-      onDayClick={handleDayClick}
```
tjfred35, line 16: We don’t need onDayClick anymore? I’d like to test this a little bit. I’m not looking at the code right now, but I’m unfamiliar with the code calls onSelect in the custom mode

website/examples/customization-fixedweeks.tsx
gpbl: Good catch thanks, restored in b2ba94237a8427021dac77a4baaafd953352ddef.

website/docs/guides/custom-selections.md
gpbl: Thanks fixed in b508d948432d138abed260923935b652f7a5fca4.

website/examples/custom-selection-single.tsx
gpbl: Sorry for the confusion, that was a wrong copy/paste. The example is restored in dd748c5851f53e52d20326c7cb4d9c874c934cf1.

website/examples/navigation-disabled.tsx
gpbl: I prefer this version with no "magic": without specifying `selected` or `onSelect`, nothing happens. 

(I will remove the mode prop from this example)

gpbl:
> Is “being able to play with DayPicker out of the box without wiring up an external React.useState” a reason to continue to support an uncontrolled mode?

To me yes so far – what do you think? :)

gpbl:
@tjfred35 I could fix some bugs that should address your concerns about the click events (without the tests this is getting complicated).

gpbl:
I'm merging this as it is required for the next PRs.

PR #1292, opened 2021-11-25
chore: re-include main field

mayteio:
Re-include `main` field in package.json for CJS bundlers


gpbl:
Thanks, I was publishing this just in the same moment: de3336ab4f4c2b0a1a1b0f2fd5b22b5f96b7ed4c 

beta.31 should include the fix!

PR #1293, opened 2021-11-25
(chore) useInput cleanup

gpbl:
Some cleanup for the `useInput` hook. Tests coming soon...


packages/react-day-picker/src/hooks/useInput/useInput.ts
```diff
@@ -3,26 +3,80 @@ import * as React from 'react';
...
-} from './types';
-import { UseInput } from './types/UseInput';
+  DayClickEventHandler,
+  DayPickerProps,
+  MonthChangeEventHandler
+} from '../../types';
 import { isValidDate } from './utils/isValidDate';
 
-const DefaultFormat = 'PP';
+/** Represent the props to attach to the input field. */
```
gpbl, line 14: I did not like the idea of placing a type in each file, so when I met them – I move to the file where they are used.

packages/react-day-picker/src/hooks/useInput/useInput.ts
```diff
@@ -3,26 +3,80 @@ import * as React from 'react';
...
+  /** Event handler for the blur event. */
+  onBlur: React.FocusEventHandler;
+  /** Event handler for the change event. */
+  onChange: React.ChangeEventHandler<HTMLInputElement>;
+  /** Event handler for the focus event. */
+  onFocus: React.FocusEventHandler;
+  /** The value of the input field */
+  value: string;
+  /** The input field placeholder */
+  placeholder?: string;
```
gpbl, line 25: The placeholder prop was missing in the previous return value. I added it here.

PR #1299, opened 2021-11-29
feat: remove the day_today default class name and provide some type h…

tjfred35:
…ints for the internal modifiers

This allows for a bit of type hinting, as shown in the screenshot below:

Screen Shot 2021-11-29 at 9 31 04 AM

It also removes the `day_today` type hint as shown in the screenshot below, which was incorrectly shown, and moves 'today' to an internal modifier, which is how "today" is actually invoked.

Screen Shot 2021-11-29 at 9 33 40 AM

Lastly, this provides type hints to anyone using ModifierStatus. For example, in my custom Day class:

Screen Shot 2021-11-29 at 9 35 33 AM

### DIFF
--- a/packages/react-day-picker/src/contexts/DayPicker/defaultClassNames.ts
+++ b/packages/react-day-picker/src/contexts/DayPicker/defaultClassNames.ts
@@ -43,11 +43,9 @@ export const defaultClassNames: Required<ClassNames> = {
   nav_icon: 'rdp-nav_icon',
 
   row: 'rdp-row',
-  row_head: 'rdp-row_head',
   weeknumber: 'rdp-weeknumber',
   cell: 'rdp-cell',
 
   day: 'rdp-day',
-  day_outside: 'rdp-day_outside',
-  day_today: 'rdp-day_today'
+  day_outside: 'rdp-day_outside'
 };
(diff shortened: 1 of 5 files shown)


gpbl:
Where did the CI go? :(

gpbl (approved):

packages/react-day-picker/src/contexts/DayPicker/defaultClassNames.ts
```diff
@@ -43,11 +43,9 @@ export const defaultClassNames: Required<ClassNames> = {
...
 
   row: 'rdp-row',
-  row_head: 'rdp-row_head',
   weeknumber: 'rdp-weeknumber',
   cell: 'rdp-cell',
 
   day: 'rdp-day',
-  day_outside: 'rdp-day_outside',
-  day_today: 'rdp-day_today'
+  day_outside: 'rdp-day_outside'
```
gpbl, line 50: The typecheck of this file is failing.

gpbl (changes requested):
We need to fix the types: `yarn workspace react-day-picker typecheck`

gpbl (approved):

Issue #1279, opened 2021-11-06
Add keyboard navigation for navigating between months


tjfred35:
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html

- Page up (subtract one month)
- Page down (add one month)
- Shift + Page up (subtract one year)
- Shift + Page down (add one year)
- Home (go to first day of week)
- End (go to last day of week)

rolandsfr:
Hey @gpbl, I'm ready to give this issue a shot.

**TODO**  
- [ ] Create navigation for all the 6 types of actions
- [ ] Add a  ``keybaordNavigation`` prop that will take a boolean and enable/disable the keyboard navigation (let me know if you do not want to have a different name for this prop or do not want to have it at all)

gpbl:
@rolandsfr awesome, would you start from a PR where you add the relative todo tests? The file to change for now is this:

https://github.com/gpbl/react-day-picker/blob/master/website/examples/keyboard.test.tsx

For example:

```ts
 describe('when the PageUp is pressed', () => {
        beforeEach(pressPageUp);
        test.todo('should display the previous month');
});
```

Ayan-web:
is someone working on this issue?

PR #1300, opened 2021-11-29
feat: add extended keyboard navigation

rolandsfr:
Referencing issue #1279 

**Changes**
 - Created test todos to later implement the keyboard navigation itself
 - Added the keyboard navigation functions


website/examples/keyboard.test.tsx
```diff
@@ -199,3 +205,35 @@ describe.each(['ltr', 'rtl'])(
     });
   }
 );
+
+describe('keyboard actions', () => {
```
gpbl, line 209:
```suggestion
```

website/examples/keyboard.test.tsx
```diff
@@ -199,3 +205,35 @@ describe.each(['ltr', 'rtl'])(
...
+  describe('when the Page Up is pressed', () => {
```
gpbl, line 210: All these `describes` should go inside the first `describe`, as they must run the `setup` before each and test the rtf direction too.

website/examples/keyboard.test.tsx
```diff
@@ -199,3 +205,35 @@ describe.each(['ltr', 'rtl'])(
...
+    beforeEach(pressPageUp);
+    test.todo('Should display the previous month');
+  });
+
+  describe('when the Page Down is pressed', () => {
+    beforeEach(pressPageDown);
+    test.todo('Should display the next month');
+  });
+
+  describe('when the combination Shift + Page Up is pressed', () => {
```
gpbl, line 220:
```suggestion
  describe('when Shift + PageUp is pressed', () => {
```

website/examples/keyboard.test.tsx
```diff
@@ -199,3 +205,35 @@ describe.each(['ltr', 'rtl'])(
...
+    beforeEach(pressShiftPageUp);
+    test.todo('Should display the previous year');
+  });
+
+  describe('when the combination shift + page down is pressed', () => {
```
gpbl, line 225:
```suggestion
  describe('when Shift + PageDown is pressed', () => {
```

gpbl (changes requested):
Thanks very helpful. See my review. 

Could you please replace the `when the` with just `when` in the whole file? Thanks 🙏

rolandsfr:
Committed the changes.
 - reworded `describe` blocks
 - put the `describe` blocks into the global `describe` with the setup function initialized

gpbl (approved):
Thanks looking good!

gpbl:
@rolandsfr feel free to continue with your work thanks, now we need to make these tests to pass.

rolandsfr:
I am having some problems setting up the project on my local machine.
Do you mind sharing the setup process? I have figured out the way to implement the features, it's just that I keep getting different errors related to packages.  
Does one single `npm i` in the root dir do the work or should do something additional?

rolandsfr:
> > I am having some problems setting up the project on my local machine. Do you mind sharing the setup process? I have figured out the way to implement the features, it's just that I keep getting different errors related to packages. Does one single `npm i` in the root dir do the work or should do something additional?
> 
> image this is what im getting on either `npm run develop` or ` yarn develop`

got it to work dw

gpbl:
@rolandsfr is there some change to this repo we could apply,  to help Windows users? Mind to share your solution here? I'm sorry you had this issue.

rolandsfr:
I think I just simply messed up the dependencies in some magical way, so to fix it I simply 
 1. made a hard reset of the project
 1. ran both `npm install` and `yarn` in the root dir
 1. used `yarn develop` </br>
 
... and everything started to work as it should.

My suggestion is to create a brief guide of how to get this project up and running locally on one's machine.
Also, I figured I could use the /render?example={example} endpoint on the website to test out the package while developing it only after quite some time (quite unintuitive if I say so myself).

rolandsfr:
@gpbl, I wanted to clarify what is the wanted behavior of navigation between months and years?
2 types of behavior can be implemented, when the user presses, for example, presses PageUp:
1. Go back a month and save the initially focused day
2. Go back a month and focus(/select?) on the day of this month that matches the initial day

tjfred35:
The arrow keys and (theses key too) should move the focused day. Selection is achieved with space or enter

rolandsfr:
@tjfred35 ok perfect, thank you

gpbl:
> My suggestion is to create a brief guide of how to get this project up and running locally on one's machine.

Thanks! I'm trying to maintain such guide here: https://react-day-picker-next.netlify.app/development appreciate the feedback.

gpbl:
@rolandsfr putting this on draft until ready 👍

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -41,11 +67,22 @@ export function FocusProvider({
   children: React.ReactNode;
 }): JSX.Element {
   const [focusedDay, setDay] = React.useState<Date | undefined>();
-  const { goToMonth, displayMonths } = useNavigation();
+  const [tempoFocusedDay, setTempo] = React.useState<Date | undefined>();
```
gpbl, line 70: What do you mean with tempo focused day? Why we need to add this?

```suggestion
```

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -41,11 +67,22 @@ export function FocusProvider({
...
+  const initialFocusChanged = React.useRef(false);
```
gpbl, line 71: Why do we need to track the initial focus? I wasn't expecting that...

```suggestion
```

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -41,11 +67,22 @@ export function FocusProvider({
...
-  const focus = (date: Date) => setDay(date);
+  const tempoFocusDay = (date: Date) => setTempo(date);
+  const focus = (date: Date, keepInitial: boolean = false) => {
+    setDay(date);
+  };
+
+  const changeInitialFocus = (date: Date): void => {
+    setDay(date);
+    setTempo(date);
+  };
```
gpbl: Please remove this

```suggestion
```

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -59,35 +96,148 @@ export function FocusProvider({
   const focusDayBefore = () => {
     if (!focusedDay) return;
     const before = addDays(focusedDay, -1);
-    setDay(before);
+    changeInitialFocus(before);
```
gpbl, line 99: Please remove these here and below:

```suggestion
```

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -59,35 +96,148 @@ export function FocusProvider({
...
   };
 
+  const focusFirstDayOfWeek = (): void => {
+    if (!focusedDay) return;
+    const dayToFocus = addDays(
+      startOfWeek(addDays(focusedDay, -1), {
+        weekStartsOn: 0
+      }),
+      1
+    );
```
gpbl, line 128: Why going to the day before? (maybe is right)

```suggestion
  const focusFirstDayOfWeek = (): void => {
    if (!focusedDay) return;
    const dayToFocus = startOfWeek(focusedDay);
```

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -59,35 +96,148 @@ export function FocusProvider({
...
+    goToMonth(dayToFocus);
```
gpbl, line 129: This should stay the same as the other functions above:

```suggestion
    switchMonth(dayToFocus, numberOfMonths);
```

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -59,35 +96,148 @@ export function FocusProvider({
...
+    changeInitialFocus(dayToFocus);
```
gpbl, line 130:
```suggestion
```

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -59,35 +96,148 @@ export function FocusProvider({
...
+    if (!focusedDay) return;
+    const dayToFocus = addDays(
+      endOfWeek(addDays(focusedDay, -1), {
+        weekStartsOn: 0
+      }),
+      1
+    );
+    goToMonth(dayToFocus);
+    changeInitialFocus(dayToFocus);
+  };
```
gpbl, line 143: Same here, please refactor as `focusFirstDayOfWeek`

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -59,35 +96,148 @@ export function FocusProvider({
...
+        ? tempoFocusedDay
+          ? addDays(startOfMonth(month), tempoFocusedDay.getDate() - 1)
+          : month
+        : addDays(
+            startOfWeek(addWeeks(startOfMonth(month), 3)),
+            tempoFocusedDay ? tempoFocusedDay.getDay() : focusedDay.getDay()
+          ),
+      true
+    );
+  };
```
gpbl: Remove this. Write instead verbose functions as the current code style:

```suggestion
```

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -59,35 +96,148 @@ export function FocusProvider({
...
+
+  const focusMonthBefore = () => {
+    focusByUnit({
+      direction: 'backwards',
+      unit: 'month'
+    });
+  };
```
gpbl, line 203: These and the other focus function should be written with the same code style as the others:

```suggestion
  const focusMonthBefore = () => {
    if (!focusedDay) return;
    const monthBefore = addMonths(focusedDay, -1);
    setDay(monthBefore);
    switchMonth(monthBefore, numberOfMonths);
   };
```

gpbl (changes requested):

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
rolandsfr: The initial focus changes every time the user navigates with arrow keys or selects a specific day.  
In order for navigation by months and years to focus on the same date as the initial focus (before PageUp, PageDowm, and other related keys were pressed) it needs to separate the initial focus state and the dynamic focus states. Meaning, when the user presses PageUp several times, the calendar will go months back and still focus on the **same** date that was saved before any of the extended keyboard navigation keys were pressed.

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
rolandsfr: These are functions that implement the reset of the initial focus and their deletion will result in broken behavior.

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
rolandsfr: For some reason, date fns lib makes it so going to the start/end of the week returns one day after the beginning/end of the week so that's why I have one additional day added

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
rolandsfr: Alright, will do.

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
gpbl: > going to the start/end of the week returns one day after the beginning/end of the week

Do they go at 0.00 and 23:59 respectively? Worth investigating.

This is sure worth investigating in another instance

tjfred35 (commented):
I don’t fully understand tempo concept or the initial focus concept yet, but I had a similar idea where we would maintain the “target” focused day, so that even when no day was in focus, the tabIndexes would be set such that upon user hitting tab key, they would hit the target focus day (instead of the first or last day of the month, depending on the direction of the tab or shift-tab focus)

packages/react-day-picker/src/components/Day/hooks/useDay.tsx
```diff
@@ -146,6 +144,7 @@ export function useDay(
       range.onDayClick?.(date, modifiers, e);
     }
     context.onDayClick?.(date, modifiers, e);
+    changeInitialFocus(date);
```
tjfred35, line 147: Wouldn’t the day already be focused by this click event? Or is it that our internal focus state doesn’t know?

tjfred35: When the user clicks on the button I think the focus event will trigger and this code will set the focus day?

https://github.com/gpbl/react-day-picker/blob/9c5789c78d5cc447be2b5a77af3b406ab8fc5099/packages/react-day-picker/src/components/Day/hooks/useDay.tsx#L152

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -106,96 +101,49 @@ export function FocusProvider({
...
 
-  const focusFirstDayOfWeek = (): void => {
+  const focusStartOfWeek = (): void => {
     if (!focusedDay) return;
-    const dayToFocus = addDays(
-      startOfWeek(addDays(focusedDay, -1), {
-        weekStartsOn: 0
-      }),
-      1
-    );
```
gpbl: Here I really could not understand: adding a day, removing a day?

gpbl: Here and the others – I don't understand: why adding days and removing days?

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -106,96 +101,49 @@ export function FocusProvider({
...
     switchMonth(dayToFocus, numberOfMonths);
     focus(dayToFocus);
   };
 
   const focusMonthBefore = (): void => {
     if (!focusedDay) return;
 
-    const month = subMonths(focusedDay, 1);
-    switchMonth(month, numberOfMonths);
-    const switchDay = Boolean(getDaysInMonth(month) < focusedDay.getDate());
```
gpbl: I couldn't understand what is a `switchDay` here.

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -106,96 +101,49 @@ export function FocusProvider({
...
-
-    focus(
-      !switchDay
-        ? month
-        : addDays(
-            startOfWeek(addWeeks(startOfMonth(month), 3)),
-            focusedDay.getDay()
-          )
```
gpbl: Why adding 3 weeks and get the start of the week of that?

gpbl (approved):
Thanks I've been updating the code a bit give a look.

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
rolandsfr: In case the day focused is 31th/30th or 29th - if that date does not exist in the prev/following month - it makes it to focus on the same day of the week

rolandsfr: otherwise the focus will disappear

PR #1301, opened 2021-11-30
fix: labelDay missing the modifiers argument

tjfred35:
This becomes a typescript error when we more strongly type ModifierStatus (the second parameter) -- see https://github.com/gpbl/react-day-picker/pull/1299.

### DIFF
--- a/packages/react-day-picker/src/components/Day/Day.test.tsx
+++ b/packages/react-day-picker/src/components/Day/Day.test.tsx
@@ -1,3 +1,35 @@
+import React from 'react';
+import { es, enUS } from 'date-fns/locale';
+import { Day } from './Day';
+import { customRender } from '../../test';
+import { DayPickerProps } from '../../types';
+
+const renderNovember182021 = (contextValue: DayPickerProps) => {
+  return customRender(
+    <Day displayMonth={new Date(2021, 10)} date={new Date(2021, 10, 18)} />,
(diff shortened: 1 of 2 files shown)


packages/react-day-picker/src/components/Day/Day.test.tsx
```diff
@@ -1,3 +1,35 @@
...
+    expect(container).toHaveTextContent('18th November (Thursday)');
+  });
+});
+describe('when the locale is es', () => {
+  const locale = es;
+  test('the day should have aria text content in spanish', () => {
+    const { container } = renderNovember182021({
+      locale
+    });
+    expect(container).toHaveTextContent('18º noviembre (jueves)');
```
tjfred35, line 29: This is the test that was failing for me before the fix to the parameters below.

gpbl (changes requested):
Thanks! 

Please remove the test, which will be done eventually for `labelDay` ("expect that `format` is called with options`).

tjfred35:
I had a feeling you wouldn't like this test. Anyway, I've fixed it here too: https://github.com/gpbl/react-day-picker/pull/1299. So I'll close this PR.

gpbl:
> 



> I had a feeling you wouldn't like this test.

 already thinking to a eslint plugin to help sharing these best practices 🦹🏼‍♀️

PR #1314, opened 2021-12-04
feat: add DayPickerSingleRequiredProps

gpbl:
This new type will make the type check to _not accept_ `undefined` as `selected` value while `mode="single"` and `required={true}`. Not sure if is the correct approach, and if we should add something similar for the Mul

The test case is included in the PR: it removes `undefined` from the example. cc @tjfred35

### DIFF
--- a/packages/react-day-picker/src/DayPicker.tsx
+++ b/packages/react-day-picker/src/DayPicker.tsx
@@ -5,7 +5,8 @@ import {
   DayPickerMultipleProps,
   DayPickerProps,
   DayPickerRangeProps,
-  DayPickerSingleProps
+  DayPickerSingleProps,
+  DayPickerSingleRequiredProps
 } from './types';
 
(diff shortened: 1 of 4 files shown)


tjfred35 (commented):
This seems ok to me, but I would wonder if it's the correct behavior. Radio buttons are "required", or can be made required, in that you cannot deselect a value once selected, but that doesn't mean they must have an initial value.

gpbl:
>  Radio buttons are "required", or can be made required, in that you cannot deselect a value once selected, 

I wanted also to follow that behaviour but it is not playing well with the types. That you cannot deselect once selected seems counter-intuitive to me... 

My concern here is more adding an extra interface just for that – we should do this with props for the other selection modes, or are there better alternatives? I've blocked myself with the "multiple" mode which does not have `required` but `min`.

Sometimes I think we should throw warnings in development but I'd not go so far...

PR #1315, opened 2021-12-05
chore: re-enable @typescript-eslint/no-unused-vars

gpbl:


packages/react-day-picker/src/contexts/DayPicker/DayPickerProvider.tsx
```diff
@@ -55,6 +55,7 @@ export function DayPickerProvider(props: DayPickerProviderProps): JSX.Element {
     modifiers.disabled.push({ after: toDate });
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
gpbl, line 58: I don't like adding this, but how to remove the props and make type check happy?

PR #1319, opened 2021-12-06
fix: missing today modifier

tjfred35:
I recently changed the evaluation of `context.modifiers.today` from `undefined` to `[]`. This caused `today: context.modifiers.today ?? [context.today]` line to evaluate to `[]` instead of `[context.today]`, meaning the modifier class name never got applied to the current day.

I set up a high-level test to ensure this regression doesn't happen again, and implemented it two different ways (the test allows us to play around with the implementation as much as we want, right?)


packages/react-day-picker/src/hooks/useModifiers/useModifiers.ts
```diff
@@ -21,9 +21,7 @@ export function useModifiers(date: Date): UseModifiers {
 
   // TODO: modifiers from context should not be overridden
```
tjfred35, line 22: FWIW, I think the only modifiers that need overriding now are `disabled`, `range_start`, `range_middle`, and `range_end`.

Pretty sure these lines:
```
    modifiers.selected = modifiers.selected.concat(
      singleSelect.modifiers.selected ?? []
    )
```
can all now be deleted and still work

tjfred35: (out of scope here I think though)

test-integration/examples/modifiers-today.test.tsx
```diff
@@ -16,7 +16,7 @@ beforeEach(() => {
 describe('when the today date is clicked', () => {
   beforeEach(() => clickDay(today));
   test('should update the footer', () => {
-    expect(getTableFooter()).toHaveTextContent('Try clicking the today’s date');
+    expect(getTableFooter()).toHaveTextContent('You clicked the today’s date');
```
tjfred35, line 19: This test should have caught it, but instead it was expecting the same result as the test below it.

packages/react-day-picker/src/DayPicker.test.tsx
```diff
@@ -1,3 +1,23 @@
...
+import { DayPicker } from './DayPicker';
+
+const today = new Date(2022, 5, 10);
+const tomorrow = new Date(2022, 5, 11);
+freezeBeforeAll(today);
+
+describe('when rendering a month that contains today', () => {
+  beforeEach(() => render(<DayPicker month={today} mode="single" />));
+  test('it should add the default class name for today', () => {
+    expect(getDayButton(today)).toHaveClass('rdp-day_today');
```
tjfred35, line 14: I think this is a nice high-level test that tests not the boolean value for the `today` modifier status, but that the class name is applied...seems ok to keep from my perspective, even though we also have a test for modifiers-today.tsx

packages/react-day-picker/src/contexts/DayPicker/DayPickerProvider.tsx
```diff
@@ -54,6 +54,7 @@ export function DayPickerProvider(props: DayPickerProviderProps): JSX.Element {
   if (toDate) {
     modifiers.disabled.push({ after: toDate });
   }
+  modifiers.today = [today];
```
tjfred35, line 57: Following the advice of `// TODO: modifiers from context should not be overridden` in `useModifiers`, I am properly assigning the modifiers to `[today]` here (the modifiers from the context), so that we are not overriding them in `useModifiers`

packages/react-day-picker/src/hooks/useModifiers/useModifiers.ts
gpbl: @tjfred35 please remove it then :) Thanks!

test-integration/examples/modifiers-today.test.tsx
```diff
@@ -13,10 +13,20 @@ beforeEach(() => {
...
 });
 
+describe('when rendering a month that contains today', () => {
+  test('it should add the default class name for today', () => {
+    expect(getDayButton(today)).toHaveClass('rdp-day_today');
+  });
+  test('it should not add this class name for any day that is not today', () => {
+    const notToday = addDays(today, 1);
+    expect(getDayButton(notToday)).not.toHaveClass('rdp-day_today');
+  });
```
gpbl, line 23: This test is just checking that "tomorrow" doesn't have the class. Ideally here should be a `test.each` and run over all the days. As alternative, you could get the container from the setup render result, and check how many `.rdp-day_today` class are rendered 

Make the `setup` to return the container, [like here](https://github.com/gpbl/react-day-picker/blob/master/test-integration/examples/custom-day.test.tsx).

Then you can write instead:

```suggestion
test('the calendar should have only one ".rdp-day_today" class', () => {
    const todays = container.querySelectorAll('.rdp-day_today');
    expect(todays).toHaveLength(1);
});
```

gpbl (changes requested):
Awesome love we have a new test. I'd update the last one to be complete, tho… Thanks!

packages/react-day-picker/src/hooks/useModifiers/useModifiers.ts
tjfred35: Err! Ok...does that mean a big refactor is cool?

tjfred35: Alright...well...hope that wasn't premature, but I took a chance and made the big refactor. Based on some of the refactors you have been doing, I figured you would be comfortable reviewing it.

packages/react-day-picker/src/hooks/useDayModifiers/useDayModifiers.ts
```diff
@@ -0,0 +1,51 @@
+import { ModifierStatus } from '../../types/Modifiers';
+import { useDayPicker } from '../../contexts/DayPicker';
+import { useModifiers } from '../../contexts/Modifiers';
+import { getModifierStatus } from './utils/getModifierStatus';
+
+export type DayModifiers = {
+  /** The status of the modifiers */
+  modifiers: ModifierStatus;
```
gpbl, line 8: I don't like much a `modifiers` have a  `ModifierStatus` type, when we also have `Modifiers` type:

- rename `ModifierStatus` to `ModifiersStatus`
- evaluate this suggestion (but may be the API comes less elegant?)

```suggestion
  modifiersStatus: ModifiersStatus;
```

packages/react-day-picker/src/hooks/useDayModifiers/useDayModifiers.ts
```diff
@@ -0,0 +1,51 @@
...
+  /** The class names based on modifiers. */
+  modifierClassNames: string[];
+  /** The style based on modifiers. */
+  modifierStyle: React.CSSProperties;
+};
+
+/** Return the modifiers and its styles for the specified date. */
+export function useDayModifiers(date: Date): DayModifiers {
```
gpbl, line 16:
```suggestion
/** Return modifiers and styles for the specified day. */
export function useDayModifiers(day: Date): DayModifiers {
```

packages/react-day-picker/src/hooks/useDayModifiers/useDayModifiers.ts
```diff
@@ -0,0 +1,51 @@
...
+  const context = useDayPicker();
+
+  const modifiers = useModifiers();
```
gpbl, line 19: 🤔 Not clear the difference between `useModifiers` and `useDayModifiers`. Thinking to a better name for `useModifiers`...

packages/react-day-picker/src/types/DayPickerCustom.ts
```diff
@@ -1,10 +1,11 @@
 import { DayPickerProps } from './DayPicker';
+import { Matcher } from './Matchers';
 
 /** The props for the [[DayPicker]] component when using `mode="custom"`. */
 export interface DayPickerCustomProps extends DayPickerProps {
   mode: 'custom';
   /** The selected day(s). */
-  selected?: unknown;
+  selected?: Matcher | Matcher[];
```
gpbl, line 8: I think we can remove this line as inherited from `DayPickerProps`:

```suggestion
```

gpbl (changes requested):
Great work thanks! 👌

I haven't checkout the code yet, butt I already see we need to add more tests (ModifiersContext.tsx). I can help with that! Also I want to see if we can make clear the difference between the `useModifier` vs `useDayModifier`.

test-integration/examples/modifiers-today.test.tsx
```diff
@@ -1,22 +1,33 @@
...
 beforeEach(() => {
-  render(<Example />);
+  container = render(<Example />).container;
+});
+
+describe('when rendering a month that contains today', () => {
+  test('it should add the default class name for today', () => {
+    expect(getDayButton(today)).toHaveClass('rdp-day_today');
+  });
+  test('it should have exatct one ".rdp-day_today" class', () => {
```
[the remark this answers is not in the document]
tjfred35: Ah! I meant to say, "exactly"

packages/react-day-picker/src/hooks/useDayModifiers/useDayModifiers.ts
tjfred35: The line between what a "date" and what a "day" is is not clear to me. In Day.tsx, we seem to take a `date` and convert it into a `Day`:
```
const day = useDay(props.date, props.displayMonth, buttonRef);
```

useDay also uses `date`:
```
export function useDay(
  /** The day rendered in the month. */
  date: Date,


```
  const { modifiers, modifierClassNames, modifierStyle } =
    useDayModifiers(date);
```

So I think the parameter name should still be `date`

Comment (gpbl) [reply packages/react-day-picker/src/hooks/useDayModifiers/useDayModifiers.ts:16]: ahah yeah its not clear at all. I think if a hook is called "useDay" than it should have an argument called "day". Otherwise – which day is "using"?

**Comment by gpbl** [comment]
Going to help with the conflicts :)

File: packages/react-day-picker/src/contexts/Modifiers/ModifiersContext.tsx [hunk 11]
```diff
@@ -0,0 +1,93 @@
...
+    modifiers: customModifiers,
+    selected,
+    disabled,
+    hidden,
+    today,
+    fromDate,
+    toDate
+  } = {
+    ...context
+  };
```
gpbl, line 49: Why are you { ...context } here? 🤔

packages/react-day-picker/src/hooks/useDayModifiers/useDayModifiers.ts
```diff
@@ -0,0 +1,51 @@
...
+  modifierClassNames: string[];
+  /** The style based on modifiers. */
+  modifierStyle: React.CSSProperties;
+};
+
+/** Return the modifiers and its styles for the specified date. */
+export function useDayModifiers(date: Date): DayModifiers {
+  const context = useDayPicker();
+
+  // TODO: Rename this hook useModifiersContext?
```
gpbl, line 19: I am wondering too – if we should change the others too. Let's keep like this ATM

```suggestion
```

gpbl (approved):
Thanks, merging this

packages/react-day-picker/src/components/Day/hooks/useDay.tsx
```diff
@@ -20,7 +20,7 @@ import { useDayFocus } from './useDayFocus';
 import { SelectMultipleContextValue } from '../../../contexts/SelectMultiple/SelectMultipleContext';
 import { SelectRangeContextValue } from '../../../contexts/SelectRange/SelectRangeContext';
 import { SelectSingleContextValue } from '../../../contexts/SelectSingle/SelectSingleContext';
-import { useModifiers } from '../../../hooks/useModifiers';
+import { useDayModifiers } from '../../../hooks/useDayModifiers';
```
gpbl, line 23: I wish we could use @contexts, @hooks… how can we ?

tjfred35: Wouldn't webpack aliases work?

packages/react-day-picker/src/contexts/Modifiers/ModifiersContext.tsx
tjfred35: Because I have to pass in `context` in full to `isDayPickerMultiple(context)` below, I can't destructure it right away at the useDayPicker() call site.

packages/react-day-picker/src/components/Day/hooks/useDay.tsx
tjfred35: Nevermind, we're using rollup...

packages/react-day-picker/src/contexts/Modifiers/ModifiersContext.tsx
gpbl: OK but then better write 

```ts
  const {
    modifiers: customModifiers,
    // snip
  } = context;
```

without creating a new object.

PR #1321, opened 2021-12-07
chore: Make test:watch able to run against the source code

tjfred35:
Without this index.ts file, the module resolution code of jest
has to use the package.json file to resolve the modules, which
means jest is testing the code in the `dist` folder of react-day-picker
which is not rebuilt when running `test:watch`. This index.ts file
fixes that.

This allows for a testing technique called mutation testing, 
where we comment out parts of our production code and see
what tests fail (or don't fail).

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
     "build": "yarn workspace react-day-picker run build && yarn workspace website run build",
     "lint": "concurrently -n website,package \"yarn workspace website lint\" \"yarn workspace react-day-picker lint\"",
     "test": "yarn jest",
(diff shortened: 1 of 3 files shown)


gpbl:
This solution _may_ work but I prefer to keep the integration tests with the built version behind. I

 know we are losing the watch feature, but it is something I want to see better – without need to change the package itself. I actually think it is OK we test the build in the integration tests – the problem is how to get jest watch the built files :)

With #1324 we should be able to run `yarn test --watch` against website and trigger a new test from the command line or vscode.

Issue #1311, opened 2021-12-04
v8 unit tests umbrella


gpbl:
## Context 

Unit tests have been neglected during the complicate history of v8. We can't release it if we don't have a good test coverage. The internal API maybe still over engineered. We may need some refactoring while writing the tests.

Since there's still lot to test, I'd propose starting a TDD approach by writing first the todo tests. While writing the todo tests we can easily identify the weakness in the API.

## Unify testing environment

We are now testing both website and the package in two different ways: test utils, page objects, etc. Before completing the unit tests, we need to create an environment where we can write test more quickly. We need also to follow a stricter pattern so that the tests are more consistent each others.

## Add todo tests and refactor code

We need to scan the source, identify the missing todo tests, and take the opportunity to refactor and cleanup the code if needed. This means isolate better parts of the code in their own files or refactor props or returned values.

## Complete the test coverage

Once the todo tests are complete we can triage them,

### Environment
- [x] unify testing environment - #1313 
- [x] remove deprecated `PageObject`

### Components

- [X] Button
- [X] Caption
- [X] CaptionLabel
- [x] Day
- [x] DayContent
- [x] Dropdown
- [X] Footer
- [x] Head
- [X] IconDropdown
- [X] IconLeft
- [X] IconRight
- [X] Month
- [x] MonthsDropdown
- [x] Navigation
- [X] Root
- [x] Row
- [x] Table
- [x] WeekNumber
- [x] YearsDropdown

### Contexts

- [X] DayPicker
  - [X] Formatters
  - [X] Labels
- [X] Focus
- [x] Modifiers
- [x] Navigation
- [x] SelectMultiple
- [x] SelectRange
- [x] addToRange
- [x] SelectSingle

### Hooks

- [X] useActiveModifiers
- [X] useControlledValue
- [X] useDayEventHandlers
- [X] useDayRender
- [X] useSelectedDays

PR #1327, opened 2021-12-09
chore: Month component

gpbl:
In the context of #1311, extract a new component rendering the month from the `Root` component.


packages/react-day-picker/src/components/Month/Month.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+
+import { useDayPicker, useNavigation } from '../../contexts';
+import { Table } from '../Table';
+
+/** The props for the [[Month]] component. */
+export interface MonthProps {
+  displayIndex: number;
+  displayMonth: Date;
```
tjfred35, line 9: Seems a bit weird to pass in both of these and then also query for `displayMonths`...perhaps we only pass in displayMonth or displayIndex, and we compute the other based on displayMonths?

packages/react-day-picker/src/components/Month/Month.test.tsx
```diff
@@ -0,0 +1,188 @@
...
+import { customRender } from '@test/render';
+
+import { Month, MonthProps } from './Month';
+import { DayPickerProps } from '../../types';
+
+let root: HTMLDivElement;
+function setup(props: MonthProps, dayPickerProps: DayPickerProps) {
+  const renderResult = customRender(<Month {...props} />, dayPickerProps);
+  root = renderResult.container.firstChild as HTMLDivElement;
+  console.log(root.style);
```
tjfred35, line 12: Do we want this here?

tjfred35 (approved):
LGTM

packages/react-day-picker/src/components/Month/Month.test.tsx
gpbl: Sure no, added the eslint rule. Thanks!

packages/react-day-picker/src/components/Month/Month.tsx
gpbl: I'd prefer we keep `displayMonth` and remove `displayIndex`, then get the index from the hoo. 

Still I have no idea how to write the test: I think I should mock the contex. I'd keep it like it is for this PR: being internal component shouldn't affect too much. Feel free to give a look to see if we can remove one prop.

PR #1331, opened 2021-12-10
chore: path aliases

gpbl:


tjfred35 (approved):
LGTM

gpbl:
@tjfred35 I've found other parts that were going wrong. I pushed a completely different approach :)

It is a difficult task because we need to make vscode, tsc, types, jest and rollupjs happy _all together_. Also I want to reuse the same test utils from rdp in website.

I've set the `/src` dir as the `baseDir` in tsconfig. Now we can write absolute imports from that dir, like `import { something } from 'contexts/something'`. I've also removed few index.ts, as they were reducing the scope of our imports and make difficult to track.

PR #1339, opened 2021-12-11
chore: don't run workflows twice on pushes to pull request branches

tjfred35:
I think all pushes to a pull request branch will trigger via the pull_request event. Let's test and find out.


gpbl:
Looking good! the failed test is expected.

PR #1342, opened 2021-12-13
package: move from rollup to tsc

gpbl:
I've been fighting rollup.js for the whole weekend. I wondered if we really need a bundler! 

Here i'm trying to just use the TS compiler for build the package. Paired with [tsc-alias](https://github.com/justkey007/tsc-alias#readme) I could even maintain the absolute internal paths.


tjfred35 (approved):
This looks great. I verified that I can take the resulting build and use it in my component library.

PR #1343, opened 2021-12-13
feat: build a more intelligent target focused day

tjfred35:
This is based on

https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html

"When the dialog is opened, if the input field is empty, or does not contain a valid date,
then the current date is focused in the calendar. Otherwise, the focus is placed on the
day in the calendar that matches the value of the date input field."

This particular set of changes gets us half-way to the described state above, by allowing a
more intelligent tracking of which day in the grid should be the target of the focus when the user
presses the tab key.

The second part that will be required is some sort of `initialDayGridFocus` property that can be
used to grab the focus within a useEffect hook.

I've saved that for a follow up PR.

A third big of follow up work to meet the requirements of that wai-aria-practices example will
be to ensure that react-day-picker enables the dialog to implement a focus trap (https://github.com/focus-trap/focus-trap) -- might not be anything needed to build in react-day-picker, unless we took on the job of building a DayPickerDialog component of some sort...which similar to DayPickerInput, there might be strong support for, but it's a whole other ball of wax?


packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -64,7 +68,30 @@ export function FocusProvider({
...
+  >();
+
+  const targetFocusedDay =
+    focusedDay ??
+    (lastFocusedDay &&
+      displayMonths.some((displayMonth) =>
+        isSameMonth(lastFocusedDay, displayMonth)
+      ))
+      ? lastFocusedDay
+      : initialTarget;
```
tjfred35, line 89: The aria example has a somewhat more complex implementation where the targetFocusDay is moved up and down when the user navigates through months. The example has a visual indication of both the focused day (a gray outline) and the target focus day (a blue background). There is no text in the example that documents this requirement, so I've chosen not to support it: react-day-picker has no current requirement to make the target focus day visually obvious.

Since we have nice tests, we can always change the behavior we want later for the case when the user navigates to a new month.

packages/react-day-picker/src/contexts/Focus/getInitialTargetFocusedDay.tsx
```diff
@@ -0,0 +1,38 @@
...
+    if (!firstEnabledDay) {
+      firstEnabledDay = date;
+    }
+    date = addDays(date, 1);
+  }
+  if (today) {
+    return today;
+  } else {
+    return firstEnabledDay;
+  }
```
tjfred35, line 37: This is the code that implements the requirements, with the added complexity of handling disabled and hidden days, multiple months, and multiple selected days.

> if the input field is empty, or does not contain a valid date, 
> then the current date is focused in the calendar. Otherwise, 
> the focus is placed on the day in the calendar that matches 
> the value of the date input field.

https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html

packages/react-day-picker/src/components/Day/hooks/useDayFocus.ts
```diff
@@ -87,7 +88,9 @@ export function useDayFocus(
     }
   };
 
-  const isFocused = Boolean(focusedDay && !isSameDay(focusedDay, date));
+  const isFocusTarget = Boolean(
+    targetFocusedDay && isSameDay(targetFocusedDay, date)
```
gpbl, line 92:
```suggestion
    focusTarget && isSameDay(focusTarget, date)
```

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -11,12 +11,16 @@ import {
 } from 'date-fns';
 
 import { useDayPicker } from '../DayPicker';
+import { useModifiers } from '../Modifiers';
 import { useNavigation } from '../Navigation/useNavigation';
+import { getInitialTargetFocusedDay } from './getInitialTargetFocusedDay';
```
gpbl, line 16:
```suggestion
import { getInitialFocusTarget } from './getInitialFocusTarget';
```

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -11,12 +11,16 @@ import {
...
 
 /** Represents the value of the [[NavigationContext]]. */
 export type FocusContextValue = [
   /** The day currently focused */
   focusedDay: Date | undefined,
+  /** The day that is the target of focus in the day grid (tabIndex = 0) */
+  targetFocusedDay: Date | undefined,
```
gpbl, line 23:
```suggestion
  /** The day that is the target of focus in the day grid (tabIndex = 0) */
  focusTarget: Date | undefined,
```

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -64,7 +68,30 @@ export function FocusProvider({
   const { goToMonth, displayMonths } = useNavigation();
   const { numberOfMonths } = useDayPicker();
 
-  const blur = () => setDay(undefined);
+  const modifiersContext = useModifiers();
```
gpbl, line 71: Starting to think we should change all these to `useModifiersContext` etc...

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
```diff
@@ -64,7 +68,30 @@ export function FocusProvider({
...
+
+  const initialTarget = getInitialTargetFocusedDay(
```
gpbl, line 73:
```suggestion
  const initialFocusTarget = getInitialFocusTarget(
```

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
gpbl: Makes sense to me! 

I'd expand the code above without the "inliners" because it's hard to follow. Maybe it also deserve a new function and a test: this is an issue in few contexts already (how to unit test them) that we will hopefully improve with the time...

packages/react-day-picker/src/contexts/Focus/getInitialTargetFocusedDay.tsx
```diff
@@ -0,0 +1,38 @@
+import { addDays, endOfMonth, startOfMonth } from 'date-fns';
+
+import { getModifierStatus } from 'hooks/useDayModifiers/utils/getModifierStatus';
+import { Modifiers } from 'types/Modifiers';
+
+export function getInitialTargetFocusedDay(
```
gpbl, line 6: We need the jsdocs here. I am writing it right?

```suggestion
/** Returns the day that should be target of the focus when initially rendered according to... */
export function getInitialFocusTarget(
```

packages/react-day-picker/src/contexts/Focus/getInitialTargetFocusedDay.tsx
```diff
@@ -0,0 +1,38 @@
...
+  displayMonths: Date[],
+  modifiersContext: Modifiers
+) {
+  const firstDayInMonth = startOfMonth(displayMonths[0]);
+  const lastDayInMonth = endOfMonth(displayMonths[displayMonths.length - 1]);
+
+  let firstEnabledDay = undefined;
```
gpbl, line 13: With "enabled" do you mean "focusable"? Clarifying would help:

```suggestion
  let firstFocusableDay = undefined;
```

packages/react-day-picker/src/contexts/Focus/getInitialTargetFocusedDay.tsx
```diff
@@ -0,0 +1,38 @@
...
+  let today = undefined;
+  let date = firstDayInMonth;
+  while (date <= lastDayInMonth) {
+    const modifiers = getModifierStatus(date, modifiersContext);
+    if (modifiers.disabled || modifiers.hidden) {
+      date = addDays(date, 1);
+      continue;
```
gpbl, line 20: The joy of seeing a `continue` 👯

packages/react-day-picker/src/contexts/Focus/getInitialTargetFocusedDay.tsx
```diff
@@ -0,0 +1,38 @@
...
+    if (modifiers.selected) {
+      return date;
+    }
+    if (modifiers.today && !today) {
+      today = date;
+    }
+    if (!firstEnabledDay) {
+      firstEnabledDay = date;
+    }
+    date = addDays(date, 1);
```
gpbl, line 31: Cool!

packages/react-day-picker/src/contexts/Focus/getInitialTargetFocusedDay.tsx
gpbl: Love it ! good work!

website/test-integration/examples/keyboard-focus.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+  describe('pressing tab twice', () => {
+    beforeEach(() => {
+      renderDayPicker({ dir });
+      pressTab();
+      pressTab();
+    });
+    test('should focus on the next month button', () => {
+      expect(getNextButton()).toHaveFocus();
+    });
+  });
```
gpbl, line 52: NIT: We can also remove one rendering and just continue tabbing:

when tabbing once … when tabbing again… 

```suggestion
    test('should focus on the previous month button', () => {
      expect(getPrevButton()).toHaveFocus();
    });
    describe('pressing tab again', () => {
      beforeEach(() => {
        pressTab();
      });
    test('should focus on the next month button', () => {
      expect(getNextButton()).toHaveFocus();
    });
  });
```

website/test-integration/examples/keyboard-focus.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  describe('when tabbing into day grid', () => {
```
gpbl, line 54: Ah! I was expecting when tabbing 3 times. Shouldn't be the same result?

website/test-integration/examples/keyboard-focus.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+    beforeEach(() => {
+      renderDayPicker({
+        dir,
+        defaultMonth: today
+      });
+      tabToDayGrid();
```
gpbl, line 60: To be consistent i'd prefer here

```suggestion
      pressTab();
      pressTab();
      pressTab();
```

Yet if you merge the describes, you may have less `render` calls and simulate the actual user behavior:

```
describe… when rendered
  render()
  describe… when tabbed first time 
    pressTab()
      describe… when tabbed 2nd time
         pressTab()
```

(not strong opinionated about this, but the `tabToDayGrid`)

gpbl (commented):
Amazing thanks. I am the usual nitpicker with the test (which i'd improve) and dropping few comments

gpbl:
As a child feature, would make sense to add a `focusedDay` prop?

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
tjfred35: I agree about the size of that expression. How about:

```
  const isWithinDisplayMonths = (date: Date) =>
    displayMonths.some((displayMonth) => isSameMonth(date, displayMonth));

  const focusTarget =
    focusedDay ?? (lastFocusedDay && isWithinDisplayMonths(lastFocusedDay))
      ? lastFocusedDay
      : initialFocusTarget;
```

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
tjfred35: I think the problem comes from the fact that we ask users to give us `modifiers` (of type Matcher) and then we give users `modifiers` (of type ModifiersStatus) on the way out in `onDayClick`...we use the same name for two different things. But it's not a bad API! It's quite natural to use that word on both sides...

tjfred35: I can't think of anything better naming-wise

packages/react-day-picker/src/contexts/Focus/getInitialTargetFocusedDay.tsx
tjfred35: hah :)

packages/react-day-picker/test/po/po.ts
```diff
@@ -121,6 +121,18 @@ export function pressArrowDown() {
...
+export function pressShiftTab() {
+  userEvent.tab({ shift: true });
+}
+
+export function pressEnter() {
+  userEvent.keyboard('{enter}');
+}
+
 export function pressPageUp() {
   fireEvent.keyDown(getFocusedElement(), { key: 'PageUp', charCode: 33 });
```
tjfred35, line 137: I really like the userEvent API! I think for our purposes, `userEvent.keyboard` should be used instead of `fireEvent.keyDown` and instead of `userEvent.type` (used for text inputs or text areas).

https://testing-library.com/docs/ecosystem-user-event/#typeelement-text-options

tjfred35: I have half a mind to change it here, but it seems like scope creep, so I'll contain myself.

packages/react-day-picker/src/contexts/Focus/FocusContext.tsx
gpbl: Awesome thanks for the feedback! I've re-thought this countless times  :)

packages/react-day-picker/test/po/po.ts
gpbl: Make sense thanks. https://github.com/gpbl/react-day-picker/issues/1348

PR #1344, opened 2021-12-13
fix: small type fix on setSelected

Daniel-Novikov:
When using `useInput` hook, the returned function `setSelected` has incorrect type (doesn't allow 'undefined' as a value)


gpbl (approved):
Thanks. But why didn't tsc warn about this? Or is this expected?

Daniel-Novikov:
Not entirely sure...

The other solution tho is to type it like this:
```
setSelected: (date?: Date) => void;
const setSelected = (date?: Date) => {
```

Daniel-Novikov:
Or even better:
```
const setSelected: UseInput['setSelected'] = (date) => {
```

So that we don't need to type it twice

PR #1347, opened 2021-12-14
docs: add dialog example

tjfred35:
This is part 2, and it contains part 1 from here (https://github.com/gpbl/react-day-picker/pull/1343)

Just putting this up now if you want to play around with it


gpbl:
(reason I'd split this in two PRs is to help with the change log too)

gpbl:
In the dark website should change the background too:

Screenshot 2021-12-15 at 06 52 35

tjfred35:
I can split this into two parts

This API:  `initialFocus: Date` means that I have to know which date I want to focus on -- but that logic is complex, and contained within the FocusProvider, which my application will not have access to.

This is why I've gone with a boolean flag that says, "Focus on whatever you would have focused on if I had automatically tabbed three times..."

tjfred35:
Ok, here's part 2a, with tests, and so we can focus on getting the API right: https://github.com/gpbl/react-day-picker/pull/1351

Then part 2b can be about getting the website documentation right

gpbl:
> This API: initialFocus: Date means that I have to know which date I want to focus on -- but that logic is complex, and contained within the FocusProvider, which my application will not have access to.

Oh sorry I think I missed that part, so what is that prop for?

tjfred35:
The prop is there because I don’t know how to programmatically tab three times, I guess?

tjfred35:
I guess if we make the day picker a forward ref component I can possible use `ref.current.querySelector(…).focus()` instead

tjfred35:
Or actually I can just wrap the DayPicker in a dive with a ref…

gpbl:
> This is why I've gone with a boolean flag that says, "Focus on whatever you would have focused on if I had automatically tabbed three times..."

I think i expected that prop receiving a `Date` and thus the confusion :)

so this is blocked by #1351. 

PS. Could we branch this off from #1351 so to reduce the diff?

website/examples/date-picker-dialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { DayPicker } from 'react-day-picker';
+import { usePopper } from 'react-popper';
+
+import { format, isValid, parse } from 'date-fns';
+import FocusTrap from 'focus-trap-react';
+
+export default function Example() {
+  const [selected, setSelected] = React.useState<Date>();
+  const [inputValue, setInputValue] = React.useState<string>('');
+  const [isOpen, isOpenSet] = React.useState(false);
```
gpbl, line 11: I usually call the setter this way:

```suggestion
  const [isPopperOpen, setIsPopperOpen] = React.useState(false);
```

website/examples/date-picker-dialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  const refEl = React.useRef<HTMLDivElement>(null);
```
gpbl, line 13: According to their API this is called "reference":

```suggestion
  const popperReference = React.useRef<HTMLDivElement>(null);
```

website/examples/date-picker-dialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const buttonRef = React.useRef<HTMLButtonElement>(null);
+  const [popperEl, popperElSet] =
```
gpbl, line 15: Cool if we follow their API, this is called `popperElement`:

```suggestion
  const [popperElement, setPopperElement] =
```

website/examples/date-picker-dialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+    React.useState<HTMLDivElement | null>(null);
+
+  const closePopper = () => {
+    isOpenSet(false);
+    buttonRef?.current?.focus();
+  };
+
+  const { styles, attributes } = usePopper(
```
gpbl, line 23: Better don't remove the "scope" of these generic names. Do not destructure here:

```suggestion
  const popper = usePopper(
```

website/examples/date-picker-dialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+    refEl.current,
+    popperEl,
+    {
+      placement: 'bottom-start'
+    }
+  );
+
+  const dateFormat = 'P';
+  const formatDate = (date: Date) => format(date, dateFormat);
```
gpbl, line 32: NIT, too DRY: better just use 'P' (and below too)

```suggestion
  const formatDate = (date: Date) => format(date, 'P');
```

website/examples/date-picker-dialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  const handleInputChange = (
+    event: React.ChangeEvent<HTMLInputElement>
+  ) => {
+    const text = event.currentTarget.value;
+    setInputValue(text);
```
gpbl, line 38: NIT I think `text` as const name is confusing here. Better keep its original name!

```suggestion
    const { value } = event.currentTarget;
    setInputValue(value);
```

website/examples/date-picker-dialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const date = parse(text, dateFormat, new Date());
+    if (isValid(date)) {
+      setSelected(date);
+    } else {
+      setSelected(undefined);
+    }
+  };
+
+  const handleDaySelected = (date: Date | undefined) => {
```
gpbl, line 47:
```suggestion
  const handleDaySelect = (date: Date | undefined) => {
```

website/examples/date-picker-dialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+      >
+        Date
+        <input
+          type="text"
+          style={{
+            marginLeft: '4px',
+            padding: '6px 4px',
+            lineHeight: '24px',
+            fontSize: '16px',
+            height: '40px'
```
gpbl, line 75: 🤔 Thinking how to remove these style clutter and reduce the example number of lines..

website/examples/date-picker-dialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+          }}
+          placeholder="mm/dd/yyyy"
```
gpbl, line 77:
```suggestion
          placeholder={format(new Date(), 'P')}
```

website/examples/date-picker-dialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+          value={inputValue}
+          onChange={handleInputChange}
+        />
+        <button
+          ref={buttonRef}
+          type="button"
+          onClick={() => {
+            isOpenSet(true);
+          }}
```
gpbl, line 86: I'd put this in an `handleButtonClick` above...

website/examples/date-picker-dialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+          active
+          focusTrapOptions={{
+            initialFocus: false,
+            allowOutsideClick: true,
+            clickOutsideDeactivates: true,
+            setReturnFocus: () => buttonRef.current,
+            onDeactivate: () => {
+              closePopper();
+            }
+          }}
```
gpbl, line 117: NIT,

```suggestion
          focusTrapOptions={focusTrapOptions}
```

website/examples/date-picker-dialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+              ? `Choose date. Selected date is ${formatDate(
+                  selected
+                )}`
+              : 'Choose date'
+          }
+        >
+          🗓
+        </button>
+      </div>
+      {isOpen ? (
```
gpbl, line 106: NIT:

```suggestion
      {isOpen && (
```

so we can remove the `: null` below.

website/examples/date-picker-dialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+            height: '40px',
+            pointerEvents: isOpen ? 'none' : 'all'
+          }}
+          aria-label={
+            selected
+              ? `Choose date. Selected date is ${formatDate(
+                  selected
+                )}`
+              : 'Choose date'
+          }
```
gpbl, line 101: We can clean this up, with a `ariaLabel` const above:

```suggestion
          aria-label={ariaLabel}
```

gpbl (commented):
LGTM thanks a lot!

Since it is an example I expect people copy/paste I'm in extra nitpicking mode.

website/examples/date-picker-dialog.tsx
tjfred35: I agreed on ariaLabel, but for some reason, this isn't better to me to 🤷

website/examples/date-picker-dialog.tsx
tjfred35: yes, but then that changes the placeholder to today's day, which is too arbitrary in my mind for a placeholder -- not that pattern letters are much better...

website/examples/date-picker-dialog.tsx
tjfred35: It's hard to preview the website's css...I think there is some, but it doesn't load when I escape the sandpack context...

website/examples/date-picker-dialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+            isOpenSet(true);
+          }}
+          style={{
+            fontSize: '28px',
+            lineHeight: '28px',
+            marginLeft: '4px',
+            padding: '4px',
+            height: '40px',
+            pointerEvents: isOpen ? 'none' : 'all'
+          }}
```
tjfred35, line 94: We could remove this css if we went with a custom image that had an acceptable style and size

tjfred35: Like, say, the react-day-picker icon?

website/examples/date-picker-dialog.tsx
gpbl: That's fine for me this placeholder to… Yet, if you change the formatter in `formatDate` few lines above, this should be updated anyway.  I think i prefer to move these out of the function:

```ts
const formatString = 'P';
const placeholder = 'mm/dd/yyyy';

export default function Example() {
```

remove the custom `formatDate`, and always use `format` from date-fns.

website/examples/date-picker-dialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  const popper = usePopper(
+    popperReference.current,
+    popperElement,
+    {
+      placement: 'bottom-start'
+    }
+  );
+
+  const formatDate = (date: Date) => format(date, 'P');
```
gpbl, line 31: I think this wrapper adds clutter in the example code. We should always use `format` from date-fns. We could replace the calls to `formatDate` with `format(date, formatString)`, where `formatString` is declared at the top of the function (see my other comment).

website/examples/date-picker-dialog.tsx
gpbl: Wouldn't an emoji work :)

website/examples/date-picker-dialog.tsx
gpbl: no worries!

gpbl (approved):
Super cool I Followup issue: https://github.com/gpbl/react-day-picker/issues/1349

gpbl:
Wondering why the types fail here 🤔 

https://github.com/gpbl/react-day-picker/runs/4555019918

website/examples/date-picker-dialog.tsx
tjfred35: Hmm…yeah…hmm…mm/dd/yyyy is not very well localized. Maybe ‘P’ can be translated into some format string that is localized, but I guess I’ll have to think about it some more

tjfred35: I couldn't find any way to get a localized pattern -- I left the en-US-style pattern, with a comment to readers to localize as they see fit...

website/examples/date-picker-dialog.tsx
tjfred35: The img I had didn't work in sandpack, so I switched back to an emoji 🤷

website/examples/date-picker-dialog.tsx
tjfred35: Done

tjfred35:
> Super cool I Followup issue: #1349 
Fixed

tjfred35:
> Wondering why the types fail here 🤔
> 
> https://github.com/gpbl/react-day-picker/runs/4555019918

Fixed (it was because `initialFocusOnDay` needed to be changed to `initialFocus`)

gpbl (approved):

PR #1351, opened 2021-12-15
feat: support for putting initial focus on the day grid

tjfred35:
To support the use case where users will create a 
new DayPicker in a dialog, this property
allows us to put the focus on the target focus day of the day
grid, as defined by the business logic in the FocusProvider.

See https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html

### DIFF
--- a/packages/react-day-picker/src/components/Root/Root.tsx
+++ b/packages/react-day-picker/src/components/Root/Root.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 
 import { Month } from 'components/Month';
 import { useDayPicker } from 'contexts/DayPicker';
+import { useFocus } from 'contexts/Focus';
 import { useNavigation } from 'contexts/Navigation';
 
 /**
@@ -16,9 +17,12 @@ export function Root(): JSX.Element {
     style,
     styles,
     numberOfMonths,
-    showWeekNumber
+    showWeekNumber,
+    initialFocusOnDay
   } = useDayPicker();
 
+  const [, focusTarget, { focus }] = useFocus();
+  const [initialFocus, setInitialFocus] = React.useState(false);
   const { displayMonths } = useNavigation();
 
   const rootClassNames = [className ?? classNames.root];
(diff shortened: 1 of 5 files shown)


packages/react-day-picker/src/types/DayPicker.ts
```diff
@@ -229,6 +229,8 @@ export interface DayPickerProps {
   formatters?: Partial<Formatters>;
   // #endregion
 
+  initialFocusOnDay?: boolean;
```
gpbl, line 232: I'm not convinced by the name of this prop – i'd expect a "day" to "focus on", not a boolean. Maybe 

```suggestion
  /** 
  * When set, DayPicker will focus on today, or the first selected day if set.
  **/
  focus?: boolean;
```

(not sure the description makes sense, please help(

```ts
<DayPicker mode="single" focus /> // vs.
<DayPicker initialFocusOnDay mode="single" />
```

alternatives considered: `hasFocus`, `hasInitialFocus`

gpbl (approved):
Looking great! See my feedback. 

What do you think for `focus` vs. `initialFocusOnDay`. The rest of the changes would be appreciated (if makes sense), especially the API of `useFocus` - we may need to open a new issue to update it.

packages/react-day-picker/src/types/DayPicker.ts
tjfred35: How about `initialDayFocus` or `initialFocus`?

packages/react-day-picker/src/components/Root/Root.tsx
```diff
@@ -29,6 +33,13 @@ export function Root(): JSX.Element {
     rootClassNames.push(classNames.with_weeknumber);
   }
 
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
tjfred35: I tested quite a bit of this code I thought. Let me check the coverage.

packages/react-day-picker/src/types/DayPicker.ts
gpbl: @tjfred35 `initialFocus` sounds perfect, thanks for looking it up!

PR #1355, opened 2021-12-18
chore: update keyboard focus integration tests

gpbl:


packages/react-day-picker/test/po/po.ts
```diff
@@ -183,3 +183,12 @@ export function selectMonth(monthName: string) {
 export function selectYear(year: string | number) {
   userEvent.selectOptions(getYearDropdown(), String(year));
 }
+
+export function focusDaysGrid() {
```
gpbl, line 187: We should move these actions out of the po file… like… actions.ts….

website/test-integration/examples/keyboard-focus.test.tsx
```diff
@@ -21,57 +26,70 @@ const today = new Date(2022, 5, 10);
...
+              });
+            });
           });
-          describe('when the last focused day is no longer in the day grid', () => {
+          describe('when navigating to the next month', () => {
             beforeEach(() => {
               pressShiftTab(); // Back to next month button
               pressEnter(); // go to the next month
               pressTab(); // back to day grid
             });
```
gpbl, line 91: Tried to put this into a better action not sure...

PR #1360, opened 2021-12-19
chore: add tests for the ModifiersContext

gpbl:
Add tests for the Modifiers context. 

By doing that

- introduced a new `DayModifiers` type
- modified `InternalModifier` to enum
- moved out the logic from ModifiersContext so it can be tested


packages/react-day-picker/src/contexts/DayPicker/DayPickerContext.tsx
```diff
@@ -48,26 +48,28 @@ export interface DayPickerContextValue extends DayPickerProps {
...
 /**
- * The DayPicker Context shares the props passed to DayPicker within the
- * internal components. It is used to set the default values and perform
+ * The DayPicker Context shares the props passed to DayPicker within internal
+ * and custom components. It is used to set the default values and perform
  * one-time calculations required to render the days.
  *
- * Access this context from the [[useDayPicker]] hook when using custom
- * components.
+ * Developers may access to this context from the [[useDayPicker]] hook when
```
tjfred35, line 55: Either: “developers may get access to this” or “developers may access this”?

packages/react-day-picker/src/contexts/Modifiers/ModifiersContext.test.tsx
```diff
@@ -0,0 +1,15 @@
+describe('when passing a "selected" custom modifier', () => {
+  test('they should replace the internal "selected" modifiers');
```
tjfred35, line 2: Do we want to support that behavior? I kind of see “selected” as something you should never set separately from the selected prop

packages/react-day-picker/src/contexts/Modifiers/ModifiersContext.test.tsx
```diff
@@ -0,0 +1,15 @@
...
+});
+
+describe('when passing a "disabled" custom modifier', () => {
+  test('they should replace the internal "disabled" modifiers');
```
tjfred35, line 6: Or perhaps they should be additive with other internal logic related to other props like fromDate, toDate, min, max, etc. I don’t think “replace” is the right word

packages/react-day-picker/src/contexts/Modifiers/ModifiersContext.test.tsx
```diff
@@ -0,0 +1,15 @@
...
+});
+
+describe('when passing a "hidden" custom modifier', () => {
+  test('they should replace the internal "hidden" modifiers');
```
tjfred35, line 10: I don’t know of any internal hidden logic

packages/react-day-picker/src/contexts/Modifiers/ModifiersContext.test.tsx
```diff
@@ -0,0 +1,15 @@
...
+});
+
+describe('when passing a "today" custom modifier', () => {
+  test('they should replace the internal "today" modifiers');
```
tjfred35, line 14: Same as selected…why would we want to let a user specify one think for today and another for today modifier?

tjfred35 (commented):
I guess this is fine because I can’t see why users would want to overwrite the internal modifiers in the first place…a stricter API — albeit less flexible would be to have the internal modifiers overwrite the custom ones.

packages/react-day-picker/src/contexts/DayPicker/DayPickerContext.tsx
gpbl: Better! THanks1

packages/react-day-picker/src/types/DayPicker.ts
```diff
@@ -184,7 +180,7 @@ export interface DayPickerProps {
   /**
    * Add modifiers to the matching days.
    */
-  modifiers?: CustomModifiers;
+  modifiers?: DayModifiers;
```
gpbl, line 183: I've introduced a new type, `DayModifiers`, that accepts matcher as single value - as opposite of `CustomModifiers` that accept only _array_ of Matchers. 

`DayModifiers` only surfaces in the DayPicker props – in our code it is transformed to a `CustomModifiers` object (hence, only arrays).

packages/react-day-picker/src/types/Modifiers.ts
```diff
@@ -1,18 +1,29 @@
...
-export type InternalModifier =
-  | 'selected'
-  | 'disabled'
-  | 'hidden'
-  | 'today'
-  | 'range_start'
-  | 'range_end'
-  | 'range_middle';
+/** The modifiers that are used internally by DayPicker. */
+export enum InternalModifier {
```
gpbl, line 8: New `InternalModifier` enum: I actually added it to be able to comment these values in the docs.

packages/react-day-picker/src/contexts/Modifiers/ModifiersContext.tsx
```diff
@@ -16,74 +17,25 @@ export type ModifiersProviderProps = {
...
   const modifiers: Modifiers = {
-    selected: toMatcherArray(selected),
-    disabled: toMatcherArray(disabled),
-    hidden: toMatcherArray(hidden),
-    today: [today],
-    range_end: [],
-    range_middle: [],
-    range_start: []
+    ...customModifiers,
+    ...internalModifiers
```
gpbl, line 36: I've moved out the logic from `ModifiersProvider` so I could test it. Now it should be clear that we are _not_ overriding the internal modifier, as the first version of the test was saying...

packages/react-day-picker/src/contexts/DayPicker/defaultContextValues.ts
```diff
@@ -0,0 +1,61 @@
...
+import { IconRight } from 'components/IconRight';
+import { Row } from 'components/Row';
+import { WeekNumber } from 'components/WeekNumber';
+import { DayPickerContextValue } from 'contexts/DayPicker';
+
+import { defaultClassNames } from './defaultClassNames';
+import * as formatters from './formatters';
+import * as labels from './labels';
+
+export function getDefaultContextValues(): DayPickerContextValue {
```
gpbl, line 21: The test for the `ModifierContext` was easier to mock by putting the default values set by the DayPicker context into a separate function. This also allowed me to cleanup the DayPicker context.

packages/react-day-picker/src/contexts/DayPicker/DayPickerContext.tsx
```diff
@@ -99,38 +89,27 @@ export function DayPickerProvider(props: DayPickerProviderProps): JSX.Element {
 
     locale,
 
-    modifierClassNames: initialProps.modifierClassNames ?? {},
-    // TODO: Should this be initialProps.modifierPrefix ?? 'rdp-day_'?
```
gpbl, line 103: Nope, that was missing in the defaults! Looks like we are not testing it..

gpbl:
> I guess this is fine because I can’t see why users would want to overwrite the internal modifiers in the first place…a stricter API — albeit less flexible would be to have the internal modifiers overwrite the custom ones.

The test was added there only to see "how it worked", thanks for spotting it – now it should clear where are we overriding what. Much more work than I thought... 😅

Issue #1352, opened 2021-12-15
v8. Date range selection should match v7


tjfred35:
Current v8 behavior:

with `mode='range'`, on initial day selection, when user clicks a single day, we get 

```
  { from: day, to: day }
```

Desired behavior:

with `mode='range'`, on initial day selection, when user clicks a single day, we get 

```
  { from: day, to: undefined }
```

If user clicks the same day again, it fills in the to:

```
  { from: day, to: day }
```

This allows applications to infer that the range has been fully selected after two user clicks. Otherwise there’s no way to know after one click if the user had the intention of fully selecting a range whose start and end dates are the same, or if the user intends to make another click.

PR #1383, opened 2022-01-07
fix: Adjust returned range object to match v7 behaviour

donovanhutchinson-toast:
As noted in [issue 1352](https://github.com/gpbl/react-day-picker/issues/1352), this adjusts the returned object when using `mode='range'` and selecting dates. It should now when first selecting a date return:

```
{ from: day, to: undefined }
```

As opposed to the previous:

```
{ from: day, to: day }
```

It also adjusts the behaviour of the second click to return a range of 1 day.

Lastly, clicking on the later of two selected days in the range will return a range where the "from" is the day clicked, to enable starting a new range from the selected day.

The PR adds tests to cover each of the 9 return paths in `addToRange`.


gpbl:
Cool thanks! I was actually wondering what was the previous behavior :) 

Mind to update the test in `range.test.tsx` too?

donovanhutchinson-toast:
@gpbl Thanks. I'm seeing the fails locally in the website's tests. Had missed that step. It seems to have affected the min-max integration tests. I'll dig into that on Monday and see if I can sort it.

packages/react-day-picker/src/contexts/SelectRange/SelectRangeContext.tsx
```diff
@@ -118,63 +118,60 @@ export function SelectRangeProviderInternal({
...
-                max &&
-                selected.to &&
-                selected.from &&
-                isBefore(date, selected.from)
-              ) {
-                const diff = differenceInCalendarDays(selected.to, date);
-                if (diff >= max) {
-                  return true;
-                }
+      selected.to = !selected.to ? selected.from : selected.to;
```
donovanhutchinson-toast, line 121: I'm setting `selected.to` as the `from` date if it's undefined, while `to` is defined. This is as a result of changing the output of the range context to include `{ from: date, to: undefined }` on the first click.

tjfred35: That seems kind of weird to me. You are mutating shared state?

tjfred35: That seems like an unwanted side-effect

donovanhutchinson-toast: Thank you yes, I'll use a variable instead.

packages/react-day-picker/src/contexts/SelectRange/SelectRangeContext.tsx
```diff
@@ -118,63 +118,45 @@ export function SelectRangeProviderInternal({
...
-          ];
-        }
-        modifiers.range_end = [selected.to];
-      } else {
-        modifiers.range_end = [selected.from];
+            return false;
+          }
+        ];
       }
+      modifiers.range_end = [to];
```
tjfred35, line 159: There are I think no tests for this code that you changed...care to write a few?

donovanhutchinson-toast:
Note: I adjusted the way the `disabled` function works as it didn't seem to be working as expected here, but I missed a couple of situations. I'll add to the tests and fix it.

gpbl:
Thanks @donovanhutchinson-toast for your time on this. 

The PR is going a bit too far. We need to just change the lines of code resolving the issue, and the related tests (if any). Otherwise, please wait I complete the tests (see #1311) before continue working on this. Thanks again!

donovanhutchinson-toast:
No problem @gpbl - I've reverted the last two commits. The tests seem to be passing locally now having made a small change to the `SelectRangeContext`, setting a `to` using:

```
const to = selected.to ? selected.to : selected.from;
```

Perhaps [this commit](https://github.com/gpbl/react-day-picker/pull/1383/commits/161249087a9b72b565e7912286b1707f7ac59af1) could be useful at some point when writing tests, as I think there could be some issues with the existing Select Range Context functionality that the tests highlight, but if the tests don't align with what the API actually should do, that's totally understood.

gpbl:
> here could be some issues with the existing Select Range Context functionality that the tests highlight

Very likely! This is my fault for pushing not tested code. I will have time to look on this later, thanks for your patience @donovanhutchinson-toast

gpbl:
@donovanhutchinson-toast I could finally complete the tests for this parts, which unblocked #1408.  #1408 apply the same changes to `addToRange` found here. Thanks for your patience!

donovanhutchinson-toast:
Thanks!

PR #1388, opened 2022-01-27
fix: codesandbox style import

EthanStandel:
The docs currently say the v8 styles should be imported from /dist/style.css but the Code Sandbox examples in the website point to just /style.css and are subsequently all boken.


gpbl (approved):
Hey thanks a lot 👌🏽

Issue #1393, opened 2022-01-31
v8: Custom Caption component is not handled correctly


andreifloricel:
### Bug description

Custom Caption component is not handled correctly

### To reproduce 

Code sandbox: https://codesandbox.io/s/sandpack-project-forked-ykiib

#### Steps

Providing a custom component for [Caption](https://react-day-picker-next.netlify.app/api/interfaces/customcomponents#caption) has no effect.

Version: 8.0.0-beta.37

gpbl:
Thanks, what is the use case of customizing that component? I removed it because it is a complex component = but i forgot to remove it from the `CustomComponents` type. But I agree this should be part of another investigation, I'll restore that component.  Thanks for the PR too!

andreifloricel:
Thanks!
I customized the caption to provide navigation buttons both months and years.
The existing [`captionLayout="dropdown"`](https://react-day-picker-next.netlify.app/basics/navigation#using-a-drop-down-to-change-the-month) has the downside that `fromYear` & `toYear` are mandatory,limiting the navigation scope (which was a no-go in my use case)

![image](https://user-images.githubusercontent.com/5867129/152651558-31e43e41-328a-40f4-b13f-aea7f1c735a7.png)

gpbl:
@andreifloricel that looks neat, I'm interested in more solutions for the navigation. Is your requirement here to navigate quickly between years?

gpbl:
@andreifloricel would love to see how you implemented that navigation – it would help me a lot to see actual use cases for the Custom Caption component. Is it something that you can share, even just the code in a comment here?

andreifloricel:
@gpbl sure thing, here is a gist with the custom components: https://gist.github.com/andreifloricel/4f9bd525067c29cbf1575bb3ddc17f36

As it usually happens, it was done on the run and never refined :)

PR #1394, opened 2022-01-31
fix: restore custom caption component

andreifloricel:
Solves #1393 

I think this got accidentally lost when extracting the `Month` component out of `Root`
(see https://github.com/gpbl/react-day-picker/commit/fc9d913678bbdf3a07c170c4a6e37a3ba1686c3b#diff-562116226713def1cf68ac5328a7d4582aa5d3d431dd7ee5d77bd2802dccaedaL21)


gpbl:
Thanks I could add a test too!

Issue #1255, opened 2021-08-27
Unable to clear the DayPickerInput if an invalid date is submitted


lcl45:
**Describe the bug**
As said in the title it is not possible to update (or clear, in my case) the input's value if it contains an invalid date (a date that doesn't exists or if the user writes some letters) since it happens to be always undefined. For some other context [I also found this stack overflow user that seems to have the same problem as mine](https://stackoverflow.com/questions/62087670/value-is-reset-but-input-is-not).
Since I use the DayPickerInput in a form I really need the possibility to clear all the inputs.

**To Reproduce**
I don't have a specific code sandbox but I found this problem in a lot of code that I found online (e.g.: https://codesandbox.io/s/magical-darkness-kybqr?file=/src/App.js). The only thing to to is writing an invalid date and verify that the clear button doesn't clear the input.

**Expected behavior**
To be able to clear the content of the input.

pigpudle:
I vote for this too!

gpbl:
Closing as [we've released v8](https://github.com/gpbl/react-day-picker/discussions/1412): this issue belongs to a version which is not longer maintained, and it is likely fixed. In case, please open a new issue. Thanks!

PR #1395, opened 2022-02-02
allow clear date button to work when an invalid date is typed in

Go1dExperience:
#1255 
I can't explain too clearly about this, but if you do a `console.log(this.props.value)` on componentDidUpdate, you will see that the value never changes if the date is invalid, it will always be an empty string. This pull request force it to recognize the change by calling the onDayChange function with whatever typed. The 3 test cases removed are no longer necessary, because they are no longer called with `undefined`. The only thing I'm wondering is whether we need the typedValue anymore.

### DIFF
--- a/src/DayPickerInput.js
+++ b/src/DayPickerInput.js
@@ -422,16 +422,10 @@ export default class DayPickerInput extends React.Component {
       inputProps.onChange(e);
     }
     const { value } = e.target;
-    if (value.trim() === '') {
-      this.setState({ value, typedValue: '' });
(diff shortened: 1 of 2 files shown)


gpbl:
Hey thanks please update the description so we can reopen the PR.

Go1dExperience:
Hi @gpbl , due to some issue, I've created a new PR, with all the explanation in it. Please have a look there instead and tell me what you need: https://github.com/gpbl/react-day-picker/pull/1396.

PR #1396, opened 2022-02-06
Fix/1255 clear input when invalid date is typed

Go1dExperience:
#1255 

The above issue happen when the value of the Input is controlled by something other than the **Input** and the **DayPicker**. In this case, it's the reset button that changes the value to `""`. When we type invalid characters, the following line kicks in:
https://github.com/gpbl/react-day-picker/blob/aeb3e9e8f8aa4dc0cb8ed8400b0a388946e2104e/src/DayPickerInput.js#L431-L436
As you can see, the onChange being called with `undefined`, which sets the `dateStart` state of the App.js to be `undefined` everytime. The state is then passed down as value prop for the **DayPickerInput** component. So no matter what we type, the value prop is forever `undefined`. Then input the **DayPickerInput**, it's always an empty string. This part I don't know, but perhaps because of the defaultProps:
https://github.com/gpbl/react-day-picker/blob/aeb3e9e8f8aa4dc0cb8ed8400b0a388946e2104e/src/DayPickerInput.js#L141-L143
When the reset button is clicked, it sets the `dateStart` to empty string => value prop is empty string, but since it has always been an empty string, it doesn't register that the value prop has changed. And since the button doesn't trigger `handle` functions either, nothing happens to the input.
My PR does 2 things:

1. It forces the onChange to be called with whatever we type in, so the value prop will actually change as well. When the reset button is clicked, it will detect the new prop.
2. We have to update the typedValue as well, because otherwise value will be an empty string, but the typedValue will be the same, which prevents the input displaying the correct value due to this line:
https://github.com/gpbl/react-day-picker/blob/aeb3e9e8f8aa4dc0cb8ed8400b0a388946e2104e/src/DayPickerInput.js#L590

I have also removed 3 test cases, all of which expect the onChange to be called with `undefined`, but since we are calling it with whatever we type in, we don't really need those cases anymore. Or if you want I can update the test cases to be called with whatever typed in. Please let me know.

### DIFF
--- a/src/DayPickerInput.js
+++ b/src/DayPickerInput.js
@@ -196,6 +196,7 @@ export default class DayPickerInput extends React.Component {
         newState.value = formatDate(value, format, dayPickerProps.locale);
       } else {
         newState.value = value;
+        newState.typedValue = value;
       }
     }
 
@@ -423,7 +424,8 @@ export default class DayPickerInput extends React.Component {
     }
     const { value } = e.target;
     const day = parseDate(value, format, dayPickerProps.locale);
-    if (value.trim() === '' || !day) {
+    if (!day || value.trim() === '') {
+      // Day is invalid: we save the value in the typedValue state
       this.setState({ value, typedValue: value });
       if (onDayChange) onDayChange(value, {}, this);
       return;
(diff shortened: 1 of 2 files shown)


gpbl:

PR #1398, opened 2022-02-08
Setting date from the Navbar

dimitur2204:
## What does this pull request do
- Passes down a `setDate` prop to `Navbar` so that you are able to set whatever date you want and not only increment by month as was previously done.
- Binded the `showMonth` method to the Daypicker instance so that it doesn't lose context while being used from the Navbar. Shouldn't be a problem, because this method has currently been used only internally.


gpbl:

PR #1422, opened 2022-03-31
feat: add `weekStartsOn` prop

### LINKED ISSUE #1304: Add new weekStartsOn prop
Users may want to change the first day of the week.

- [ ] Add a new prop `weekStartsOn: 0 | 1 | 2 | 3 | 4 | 5 | 6` (default: `0`) to be passed to the `startOfWeek` implementations 
- [ ] Check if other parts should change according to this prop
- [ ] Add test 
- [ ] Add documentation

VladSez:
Fixes #1304 

This PR add's `weekStartsOn` prop to the `DayPicker` component and also some tests for this functionality. 
By default `weekStartsOn` is `0` (Sunday)

TODO:

- add docs

### DIFF
--- a/packages/react-day-picker/src/components/Root/Root.test.tsx
+++ b/packages/react-day-picker/src/components/Root/Root.test.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import { RenderResult, screen } from '@testing-library/react';
+import { RenderResult } from '@testing-library/react';
 import { addDays } from 'date-fns';
 import { DayPickerProps } from 'DayPicker';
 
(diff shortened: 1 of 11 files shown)


packages/react-day-picker/src/components/Root/Root.test.tsx
```diff
@@ -121,3 +121,31 @@ describe('when "initialFocus" is set', () => {
...
+
+describe('when using weekStartsOn prop', () => {
+  test('should start a week from Monday', () => {
+    const props: DayPickerProps = { weekStartsOn: 1 };
+
+    setup(props);
+
+    const firstDayOfTheWeek = screen
+      .getByRole('grid')
+      .getElementsByTagName('span')[1].textContent;
```
VladSez, line 133: not sure, if this is a good way to test this functionality...
Maybe you have some ideas, how can i rewrite the test?

gpbl:
@VladSez thanks look solid! ✨ Let me help you with the test and the docs!

VladSez:
@gpbl thank you! I also wanted to add docs and example to the website, but i don’t know how can i correctly create an example, because afaik examples are using latest ‘react-day-picker’ and my prop doesn’t exist in it :)

PR #1429, opened 2022-04-05
fix: toMonth to include the full month

mputilov:
If `toMonth` prop passed, we should limit date to the end of the month. Otherwise you cant disable dates in the same month as `toMonth` because it limits to the first day of month. Example: https://codesandbox.io/s/react-day-picker-base-forked-3b557v


gpbl:
@mputilov thanks! It makes sense! I would add more tests tho...

gpbl:
Thanks again @mputilov !

PR #1431, opened 2022-04-05
fix: isMatch condition for DateInterval

mputilov:
Fix matching condition to `||` for DateInterval for correct work

### DIFF
--- a/packages/react-day-picker/src/contexts/Modifiers/utils/isMatch.test.ts
+++ b/packages/react-day-picker/src/contexts/Modifiers/utils/isMatch.test.ts
@@ -27,6 +27,17 @@ describe('when matching an array of dates including the day', () => {
   });
 });
 
+describe('when matching date range', () => {
+  const matcher = {
+    from: testDay,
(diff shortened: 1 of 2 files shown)


gpbl:
@mputilov thanks! care to add some tests?

mputilov:
@gpbl i just try to implement the logic which works in v7. maybe it was wrong there and we should left it as it is right now and people just use array with before and after matcher. 
```
[
  {
    after: maxDate,
  },
  {
    before: minDate,
  },
]
```
instead of 
```
{
  after: maxDate,
  before: minDate,
} 
```

gpbl:
>  I just try to implement the logic which works in v7. maybe it was wrong there and we should left it as it is right now and people will use array with before and after matcher.

Yeah I agree it is a bit confusing. Let keep the behavior as it is, and see if we get more feedback. 

(The test of the DateRange was actually missing so thanks for adding it. I added it via 1f3b5604e165c01746ffba072ac1a75e346b40d4)

PR #1438, opened 2022-04-10
website: remove extraneous symbols from range-min-max example

mjomble:
The current code gives output like this:
> April 25th, 2022–$April 26th, 2022`

I removed the `$` and the `` ` ``


gpbl (approved):

gpbl:
Thanks @mjomble !

PR #1456, opened 2022-05-21
chore: set `exactOptionalPropertyTypes` in tsconfig

claviustales:
https://github.com/gpbl/react-day-picker/discussions/1455

### DIFF
--- a/packages/react-day-picker/src/components/Head/utils/getWeekdays.ts
+++ b/packages/react-day-picker/src/components/Head/utils/getWeekdays.ts
@@ -12,7 +12,10 @@ export function getWeekdays(
   /** The index of the first day of the week (0 - Sunday) */
   weekStartsOn?: 0 | 1 | 2 | 3 | 4 | 5 | 6
 ): Date[] {
-  const start = startOfWeek(new Date(), { locale, weekStartsOn });
+  const start = startOfWeek(new Date(), {
+    ...(locale && { locale }),
+    ...(weekStartsOn && { weekStartsOn })
+  });
   const days = [];
   for (let i = 0; i < 7; i++) {
     const day = addDays(start, i);
--- a/packages/react-day-picker/src/components/Navigation/Navigation.tsx
+++ b/packages/react-day-picker/src/components/Navigation/Navigation.tsx
@@ -11,9 +11,9 @@ export interface NavigationProps {
   /** The month where the caption is displayed. */
   displayMonth: Date;
   /** The previous month. */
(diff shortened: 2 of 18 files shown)


gpbl:
Cool thanks ! What about setting the  option in tsconfig.json too? I think it would make the compilation to fail, in case we have similar cases.

packages/react-day-picker/src/components/Head/utils/getWeekdays.ts
```diff
@@ -12,7 +12,10 @@ export function getWeekdays(
   /** The index of the first day of the week (0 - Sunday) */
   weekStartsOn?: 0 | 1 | 2 | 3 | 4 | 5 | 6
 ): Date[] {
-  const start = startOfWeek(new Date(), { locale, weekStartsOn });
+  const start = startOfWeek(new Date(), {
+    ...(locale && { locale }),
+    ...(weekStartsOn && { weekStartsOn })
+  });
```
gpbl, line 18: This change is weird :) I wanted to keep for `locale` and `weekStartsOn` the same type as the date-fns functions.

Are there reasons why you changed these props too? Was the compiler filing while setting the config?

packages/react-day-picker/src/components/Navigation/Navigation.tsx
```diff
@@ -11,9 +11,9 @@ export interface NavigationProps {
   /** The month where the caption is displayed. */
   displayMonth: Date;
   /** The previous month. */
-  previousMonth?: Date;
+  previousMonth?: Date | undefined;
   /** The next month. */
-  nextMonth?: Date;
+  nextMonth?: Date | undefined;
```
gpbl, line 16: OK this change make sense: I can understand we are passing `undefined` to `Navigation` and here it should fail.

gpbl (commented):
Thanks. There's something off by setting all of these to `undefined` – especially the event handlers and the `weeksStartOn`. 

While I can understand few, for most of them I don't see why the change is needed. I definitely need to investigate this more! Maybe there are some defaults to improve.

Thanks for the input!

packages/react-day-picker/src/components/Head/utils/getWeekdays.ts
claviustales: Yes. The `date-fns` `startOfWeek` function do not accept `locale` and `weekStartsOn` as `undefined`.

gpbl:
Closing this in favor of https://github.com/gpbl/react-day-picker/pull/1521

Issue #1472, opened 2022-06-21
Bug: CSS Module incompatible with 'pure' option for css-loader webpack plugin


amcurative:
### Bug description

CssSyntax error: Selector ":root" is not pure (pure selectors must contain at least one local class or id) (1:1)

```
  1 | :root {
    | ^
  2 |   --rdp-cell-size: 40px;
  3 |   --rdp-accent-color: #0000ff;
```

### To reproduce

Run webpack with css-loader where modules option is pure using CSS module from day-picker.

#### Steps

I encountered it when doing a next.js build but it can be reproduced simpler by

1. Create a project
2. Create a day-picker component, use CSS module (style.module.css) to style it
3. Modify webpack rules to include:
     {
        test: /\.css$/i,
        loader: "css-loader",
        options: {
          modules: {
            mode: "pure",
          },
        },
      },

5. Use webpack to package it and you should see the error

### Expected behavior

To not encounter this error and build successfully.

### Additional context

https://github.com/vercel/next.js/discussions/17089

This happens for next.js apps because they are defaulting to pure for this loader.

gpbl:
Not sure what "pure" mean – it is to avoid the global scope of the variables? In this case, replacing `:root` with `.rdp` would fix the issue?

andyschulzz:
@gpbl 
https://github.com/vercel/next.js/discussions/17089
we are running in the same issue, replacing :root with .rdp would fix this issue it seems.

andyschulzz:
I was able to fix this issue by adding the following to our next.config.js:

```
const regexEqual = (x, y) => {
    return (
        x instanceof RegExp &&
        y instanceof RegExp &&
        x.source === y.source &&
        x.global === y.global &&
        x.ignoreCase === y.ignoreCase &&
        x.multiline === y.multiline
    );
};

const loaderUtils = require('loader-utils');

/**
 * Generate context-aware class names when developing locally
 */
const localIdent = (loaderContext, localIdentName, localName, options) => {
    return (
        loaderUtils
            .interpolateName(
                loaderContext,
                `[folder]_[name]__${localName}`,
                options
            )
            // Webpack name interpolation returns `about_about.module__root` for
            // `.root {}` inside a file named `about/about.module.css`. Let's simplify
            // this.
            .replace(/\.module_/, '_')
            // Replace invalid symbols with underscores instead of escaping
            // https://mathiasbynens.be/notes/css-escapes#identifiers-strings
[...]

PR #1481, opened 2022-07-15
Use pure selector for root variables

andyschulzz:
fix for: https://github.com/gpbl/react-day-picker/issues/1472


gpbl:
Thanks @andyschulzz !

gpbl:
I had to roll it back as it doesn't work, the CSS variables are not working:

Screenshot 2022-07-15 at 17 42 59

PR #1484, opened 2022-07-17
Fix @reach/auto-id issue by removing the dependency

### LINKED ISSUE #1478: @reach/auto-id y React18
### Bug description

We have updated to React 18 and react-day-picker dependency @reach/auto-id is giving a lot of warnings. On npm install.

### To reproduce

Update to react 18 the main project and npm i.

### Expected behavior

It would be great to avoid all this noise. Will you update dependency or know any solution?

gpbl:
### Context 

We use the good [@reach/auto-id](https://www.npmjs.com/package/@reach/auto-id) package for assigning a unique id to the month captions. This package is creating warnings with DayPicker and React 18 (e.g. #1478), and with React 17 (https://github.com/reach/reach-ui/issues/921).

### Analysis

The [source](https://github.com/reach/reach-ui/blob/dev/packages/auto-id/src/index.tsx) of the package includes some clever code to support both React 17 and the native's `useId` from React 18. However, it doesn't play well in the context of a redistributed package like DayPicker.

### Solution

I copied in `src` the `useId` [source code](https://github.com/reach/reach-ui/blob/dev/packages/auto-id/src/index.tsx). Skipped [the lines of code](https://github.com/reach/reach-ui/blob/main/packages/auto-id/src/index.tsx#L90-L97) that would use React 18's `useId`. 

With this, DayPicker will fallback to the `useEffect` solution – even on React 18. We hopefully will find a better solution in the future.

Discarded options:

- use [accessible/use-id](https://www.npmjs.com/package/@accessible/use-id) - a fork of `auto-id`. Doesn't seem a much better solution at the burden of maintaining an extra dependency
- developing my own useId (no time for this, and I'd end up reinventing the wheel)
- make React 18+ mandatory to use the new [useId](https://reactjs.org/docs/hooks-reference.html#useid) hook - this would make lot of users unhappy as react 16 and 17 are still widely used
- keep the warning in React 18, but this wouldn't fix #1478

### DIFF
--- a/packages/react-day-picker/package.json
+++ b/packages/react-day-picker/package.json
@@ -32,9 +32,6 @@
     "src",
     "tsconfig.json"
   ],
-  "dependencies": {
-    "@reach/auto-id": "0.16.0"
-  },
   "devDependencies": {
     "@rollup/plugin-alias": "^3.1.9",
     "@rollup/plugin-commonjs": "^21.1.0",
--- a/packages/react-day-picker/src/components/Month/Month.tsx
+++ b/packages/react-day-picker/src/components/Month/Month.tsx
@@ -1,11 +1,10 @@
 import React from 'react';
 
-import { useId } from '@reach/auto-id';
-
 import { Caption } from 'components/Caption';
 import { Table } from 'components/Table';
 import { useDayPicker } from 'contexts/DayPicker';
 import { useNavigation } from 'contexts/Navigation';
+import { useId } from 'hooks/useId';
 
 /** The props for the [[Month]] component. */
 export interface MonthProps {
--- a/packages/react-day-picker/src/hooks/useId/index.ts
+++ b/packages/react-day-picker/src/hooks/useId/index.ts
@@ -1,0 +1,1 @@
+export * from './useId';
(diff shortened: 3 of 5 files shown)


dartess:
Oh my god, I spent several days in preparatory work to fix `@reach/auto-id`, I already set them up to run tests on multiple versions of React and provided support for react@18 in `@reach/utils` which is used by `@reach/auto-id`.
And now I happen to be here (most likely).
And now I don't have to worry about it.
Strange feeling. I just contributed to a package that I have never used and never will use.
Anyway, thanks for your work.

gpbl:
👋 @dartess thanks for writing here! Not sure I understand - did you commit to `@reach/auto-id` to fix the issue exposed here? 

I would love to use that package for react-day-picker – maybe with your update this will be fixed? Do you have a link to your work?

dartess:
@gpbl I just planned to do it.
But `@reach/auto-id` depends on package `@reach/utils`, which is also not compatible with React 18.
And for tests in React 18, the test environment was not configured in the repository. 

So, I first set up the tests: https://github.com/reach/reach-ui/pull/959
Then I worked on `@reach/utils`, write test and check compatible with React 18: https://github.com/reach/reach-ui/pull/960
After that, I planned to study the problems of `@reach/auto-id`.

If this is still relevant for your package, I may try continue this (It seems like it's not really needed).

Since we're talking about it here, you said this:
> However, it doesn't play well in the context of a redistributed package like DayPicker.

Can you elaborate on what you mean? I'll try to keep that in mind.

Issue #1491, opened 2022-07-29
Bug: Home and End keyboard buttons does not take weekStartsOn into consideration


apdrsn:
### Bug description

Thank you very much for this lovely day picker 👍 
In https://github.com/gpbl/react-day-picker/issues/1279 would it be possible to take `weekStartsOn` into consideration?
I am starting the week on Mondays. When I press `home` button it would take me to the Sunday before and `end` button will take me to Saturday. 

### To reproduce

[CodeSandbox](https://codesandbox.io/s/react-day-picker-8-forked-th7pb7)

#### Steps

1. Open CodeSandbox
2. Click on a date
3. Press `home` on keyboard
4. Selected date will be Sunday before
5. Press `end` on keyboard
6. Selected date will be Saturday after

### Expected behavior

When `weekStartsOn` is fx. Monday `home` button should take you to Monday.
When `weekStartsOn` is fx. Monday `end` button should take you to Sunday.

Again, thanks :)

PR #1492, opened 2022-07-29
Fix: home/end buttons behavior with startOfWeek

apdrsn:
### Context

I thought I might as well give it a go myself with this issue:
https://github.com/gpbl/react-day-picker/issues/1491

I did this in the blind as I could not trigger a `develop` script in react-day-picker package. Maybe docs need some update?

### Analysis

Forgot to take user prop weekStartsOn into consideration when focusing start or end of week using home and end keys

### Solution

I used `weekStartsOn` from the dayPicker context and used it as second parameter in date-fns `startOfWeek` and `endOfWeek`


gpbl (approved):
Well done, thanks!

PR #1493, opened 2022-08-03
Add prefix to auto generated ids

mihkeleidast:
### Context

- If some other dependency has a similar implementation of useId that only returns an incremented number, these IDs can conflict.
- Number-only ID-s are hard to read (mostly personal preference)

### Analysis

This will solve a potential conflict where another third-party dependency has a similar implementation of a useId hook. For example, some other dependency might use the one from reach-ui that this implementation is copied from.

Additionally, it makes the HTML in the DOM a bit easier to read, as `react-day-picker-1` is easier to scan for and find.

### Solution

Adding a prefix of `react-day-picker-` to the autogenerated ID-s will ensure the ID's generated from this hook are unique and will not conflict with other similar implementations.


gpbl (approved):
Thanks!

PR #1495, opened 2022-08-05
Add `addToRange` to the exported utilities

stopr29:
Hello,

### Context

I have been using v7 so far and I have updated to v8, but I have come across the problem that is also presented in this discussion: [v8 'DateUtils' is not exported from 'react-day-picker'](https://github.com/gpbl/react-day-picker/discussions/1489)

### Solution

I have looked over the project and found a function that does what the old `DateUtils.addDayToRange()` did, so I decided to export it.


gpbl:
Awesome, thanks for digging in!

PR #1510, opened 2022-08-17
fix: page keys not working when using setMonth

### LINKED ISSUE #1496: Bug: Page Up and Page Down key won't work once `setMonth` has been triggered
### Bug description

Page Up and Page Down will not work once you have used `setMonth` manually.

### To reproduce

[CodeSandbox](https://codesandbox.io/s/react-day-picker-8-forked-th7pb7)

#### Steps

1. Navigate months using page up and page down
2. Months will move accordingly
3. Click the button setting a month manually
4. Navigate months using page up and page down again
5. Months will not be updated visually only setting the value it seems

### Expected behavior

Setting months with page up/page down should be reflected visually even though you have set the month manually.

kimamula:
### Context

Fixes #1496 

### Analysis

react-day-picker should call `onMonthChange` every time when `goToMonth` is called. However, in `goToDate` in `NavigationContext`, `onMonthChange` is not called and this is the cause of #1496.

https://github.com/gpbl/react-day-picker/blob/fe80fbbb41ab23ccfb76bbd341e765627e91efd1/packages/react-day-picker/src/contexts/Navigation/NavigationContext.tsx#L61-L65

### Solution

I moved `onMonthChange` call to inside `goToMonth` to make sure `onMonthChange` is called every time when `goToMonth` is called.

### DIFF
--- a/packages/react-day-picker/src/components/CaptionDropdowns/CaptionDropdowns.tsx
+++ b/packages/react-day-picker/src/components/CaptionDropdowns/CaptionDropdowns.tsx
@@ -12,12 +12,11 @@ import { MonthChangeEventHandler } from 'types/EventHandlers';
  * Render a caption with the dropdowns to navigate between months and years.
  */
 export function CaptionDropdowns(props: CaptionProps): JSX.Element {
-  const { classNames, styles, onMonthChange, components } = useDayPicker();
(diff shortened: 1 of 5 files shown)


gpbl:
@kimamula amazing, thanks!

PR #1468, opened 2022-06-10
fix: prevent focus from moving beyond toDate and fromDate

kimamula:
Prevent focus from moving beyond toDate and fromDate to fix #1320

PR #1516, opened 2022-08-19
chore: Move moveFocus to a new function

### LINKED ISSUE #459: Disabled days can be selected
It is possible to "select" days that are disabled. People can click on disabled days, and the `selectedDays` array will get modified, with disabled days added to it, even though the `react-day-picker` will *not highlight* those days as selected.

### LINKED ISSUE #1320: Focus navigation breaks with disabled days
### Bug description

When navigating the calendar with the keyboard the focus breaks when the day to focus is disabled

### To reproduce 

[Code sandbox: https://codesandbox.io/s/react-day-picker-8-vl4ry](https://codesandbox.io/s/react-day-picker-8-forked-yk2ez?file=/src/App.tsx)

#### Steps

Focus the calendar and move with the arrows until a disabled day is met. The focus get lost.

### Expected behavior

The focus should get to the next (or previous?) focusable (not disabled) day. If there are no days to navigate to, the focus should stay in the last navigable day.

### Additional notes

Test should check all the keyboard combinations. Consider the specific use cases (e.g. navigate end of week when end of week is disabled).

gpbl:
Recently we merged #1468 to fix the focus when using `fromDate` or `toDate`. In this PR, I'm moving the function to get the next/previous focused day in a new file so that can be better tested and extended.

In the future, we may want to extend this new `getNextFocus` function to fix #1320.


gpbl:
cc @kimamula

PR #1524, opened 2022-08-26
fix: use parsed fromDate/toDate in useInput

hypergeometric:
### Context

`dayPickerProps` returned from `useInput` ignores `fromYear`/`toYear` and `fromMonth`/`toMonth` set on the options.

### Analysis

Only `option.fromDate`/`options.toDate` is included in `dayPickerProps`. This seems like an oversight and either the parsed `fromDate`/`toDate` should be used or all of the original from/to props should be added.

### Solution

Use the parsed `fromDate`/`toDate` instead of `options.fromDate`/`options.toDate`.


gpbl:
Thanks!

PR #1542, opened 2022-09-07
fix(docs): incorrect function matcher example

AntanasMisiunas:
### Context

Docs regarding `Matchers` have an example for `functionMatcher` which is incorrect.

### Analysis

Prop `day` is passed but is unused.

### Solution

Replace `new Day()` with `day`.


gpbl:
Thanks!

PR #1548, opened 2022-09-08
fix: CSS module doesn't include the root class

gpbl:


stetsche:
Thanks for addressing this so quickly!

PR #1549, opened 2022-09-09
fix: infinite recursion when focusing next days

### LINKED ISSUE #1543: Bug: "Maximum call stack size exceeded" when navigating past the allowable range
### Bug description

Using the keyboard to move outside of the range of enabled days results in an infinite loop as getNextDayFocus tries to find the next selectable day, eventually crashing with a stack error:

```
analytics:12 Uncaught RangeError: Maximum call stack size exceeded
    at isMatch (vendors-node_modules_react-day-picker_dist_index_esm_js-node_modules_react-popper_lib_esm_use-50aa90.chunk.js:9490:17)
    at vendors-node_modules_react-day-picker_dist_index_esm_js-node_modules_react-popper_lib_esm_use-50aa90.chunk.js:9533:13
    at Array.reduce (<anonymous>)
    at getActiveModifiers (vendors-node_modules_react-day-picker_dist_index_esm_js-node_modules_react-popper_lib_esm_use-50aa90.chunk.js:9531:51)
    at getNextFocus (vendors-node_modules_react-day-picker_dist_index_esm_js-node_modules_react-popper_lib_esm_use-50aa90.chunk.js:9601:31)
    at getNextFocus (vendors-node_modules_react-day-picker_dist_index_esm_js-node_modules_react-popper_lib_esm_use-50aa90.chunk.js:9604:20)
...
```

image

### To reproduce

https://codesandbox.io/s/react-daypicker-forked-5zxg66?file=/src/App.tsx

#### Steps

1. Visit the above sandbox, get keyboard focus into the day-picker with `tab`
2. Move past the current date with arrow keys. 
3. Notice that Safari javascript locks up entirely, or Chrome throws a stack error. 

### Expected behavior

Anything but an infinite loop. It's not entirely straightforward to fix though, given the `disabled` prop can be used to return arbitrary disabled/enabled dates. (Who knows, maybe every day is supposed to be disabled except for 2500-01-01, better keep trying future dates...)
Maybe getNextFocus should just bail if it gets more than 30 days ahead of the initial date?

eXamadeus:
### Context

Fixes #1543 

### Analysis

In #1543, it's pointed out that the `getNextFocus` function can infinitely recurse (or at least very profoundly recurse) if there is a complex or infinite list of non-focusable dates. There is an easy reproduction path in #1543.

### Solution

The solution is to limit the recursion. I chose 365 days arbitrarily. The value could likely be increased to some sane limit, but I'm not sure what that would be. 5 years? 10 years?

Assuming this is the chosen path forward, it should be added to the documentation somewhere.

### Open questions

- [ ] What value should `DEFAULT_RETRY_ATTEMPTS` be?
- [ ] Is `console.warn` appropriate?
- [ ] Is limiting the recursion the best approach?


packages/react-day-picker/src/contexts/Focus/utils/getNextFocus.tsx
```diff
@@ -59,7 +73,22 @@ export function getNextFocus(
     const isFocusable = !activeModifiers.disabled && !activeModifiers.hidden;
 
     if (!isFocusable) {
-      return getNextFocus(newFocusedDay, moveBy, direction, options, modifiers);
+      if (retryOptions.attempt >= retryOptions.attemptLimit) {
+        const origDayString = format(retryOptions.originalDay, 'P');
+        const warning = `Unable to select day within ${retryOptions.attemptLimit} days of ${origDayString}`;
+        // eslint-disable-next-line no-console
+        console.warn(warning);
```
eXamadeus, line 80: I'm not sure if this is a pattern you'd like to adopt @gpbl, but I figured _something_ should happen, since this just returns the last valid day: `retryOptions.originalDay`

eXamadeus: I was waffling between `console.debug` and `console.warn` but ended up with warn because it's more likely to be seen and could very well help people debug oddities in the future.

gpbl:
Thanks @eXamadeus for your PR! 
I am hijacking it, so you still appears between the contributors. Your solution is great I just needed to update the tests and align the code style. I realized there was another function that may cause the infinite recursion.

See the diff here: [`bb453d9` (#1549)](https://github.com/gpbl/react-day-picker/pull/1549/commits/bb453d994c56e2d5153d64fe923a360c6d8b39c8) - what do you think?

eXamadeus:
Sorry, it took me a while to reply @gpbl. This looks great! Thanks for cleaning it up.

PR #1571, opened 2022-10-02
fix: types for selection modes not being correctly set

gpbl:


packages/react-day-picker/src/types/DayPickerDefault.ts
```diff
@@ -1,10 +1,10 @@
 import { DayPickerProps } from 'DayPicker';
 
-import { DayPickerBase, DaySelectionMode } from './DayPickerBase';
+import { DayPickerBase } from './DayPickerBase';
 
 /** The props for the {@link DayPicker} component when using `mode="default"` or `undefined`. */
 export interface DayPickerDefaultProps extends DayPickerBase {
-  mode?: DaySelectionMode;
+  mode?: undefined | 'default';
```
VitaliyR, line 7: You already have `?` near mode so you don’t need to provide explicit `undefined`:

`mode?: 'default'`

Isn’t it?

gpbl: Actually there's a setting in tsconfig where this would be required (I couldn't find which right now), so I'd better keep it explicit.

gpbl: Found the option: https://www.typescriptlang.org/tsconfig/exactOptionalPropertyTypes.html

VitaliyR: image

FYI regarding `?` and unsetting it :)

VitaliyR:
Thanks for fixing this, I can now remove `as any` stubs to fix TS 👍

Issue #1567, opened 2022-09-27
Bug: Stuck in a focus loop trying to focus outside day


DanielJKelly:
### Bug description

This bug is related to focusing and outside days. It seems to occur in "range" mode when showing multiple months and the day that will be initially focused is part of the selected range and present on one calendar and also on the other calendar as an outside day. Both the "inside" day and the outside day have tabIndex="0" when inspected.

My initial thought is that outside days should not be considered focusable (or maybe at least have it be an option?), but perhaps I am missing a valid use case.

So perhaps this [line](https://github.com/gpbl/react-day-picker/blob/51a2ef79293b5145b1c38346cf73e17e9c414f13/packages/react-day-picker/src/contexts/Focus/utils/getInitialFocusTarget.tsx#L22) should be checking for `!activeModifiers.outside` as well? I haven't tried it out myself though, so it's possible the issue is due to something else.

### To reproduce

[Fork this CodeSandbox: https://codesandbox.io/s/react-day-picker-8-vl4ry with the code to reproduce the issue.](https://codesandbox.io/s/react-daypicker-forked-ocolmy?file=/src/App.tsx)

#### Steps

1. Focus the date picker
2. Press Tab until a day on the calendar is focused
3. Keep pressing Tab and you will be stuck cycling back between the day being focused and the right arrow being focused. You can sometimes see focus indicator on the outside day flash before focus is shifted to the "inside" day.

### Expected behavior

I would expect that the focus should continue on to the next tabbable element and that the button shown below the DayPicker can be focused.

### Screenshots

https://user-images.githubusercontent.com/22672008/192423154-d0ca4235-505c-4bac-a804-b0b1d797f3a6.mov

PR #1576, opened 2022-10-07
fix: disable tab for outside days (#1567)

DanielJKelly:
### Context

Please see https://github.com/gpbl/react-day-picker/issues/1567 for more context about the issue this PR is meant to fix.

### Analysis

The issue is that `const isFocusTarget = Boolean(focusContext.focusTarget &&
      isSameDay(focusContext.focusTarget, day))` will return true for multiple days if there is a day that is both "selected" and "outside".
      
 If a day is both outside and selected, the equivalent "inside" day must also be present on one of the displayed calendars, so this day will receive tabIndex=0 and there is no need for the outside version of the day to have tabIndex=0.

### Solution

Per the above analysis, my solution is to not consider a day with both the selected and outside active modifiers as a focus target, meaning the tabIndex will be set to -1.


DanielJKelly:
"If a day is both outside and selected, the equivalent "inside" day must also be present on one of the displayed calendars"

I just realized this is not true, whoops. Perhaps I could check if the day is within more than one displayed month 🤔

DanielJKelly:
> "If a day is both outside and selected, the equivalent "inside" day must also be present on one of the displayed calendars"
> 
> I just realized this is not true, whoops. Perhaps I could check if the day is within more than one displayed month 🤔
> 

I updated the PR to check if a day is both outside and displayed, which I think makes sense.

gpbl:
Hey @DanielJKelly thanks for your PR, I've been investigating #1567 too and this one seems a good solution.

> If a day is both outside and selected, the equivalent "inside" day must also be present on one of the displayed calendars,

I want to give a second look: I'm not 100% sure of this (yet) :)

packages/react-day-picker/src/hooks/useDayRender/useDayRender.tsx
```diff
@@ -99,8 +101,16 @@ export function useDayRender(
...
+    activeModifiers.outside && navigation.isDateDisplayed(day);
+
+  // If a day is both outside and displayed, the equivalent "inside" day must also be present on one
+  // of the displayed calendars, and we do not want to have multiple days with tabIndex="0" because
+  // it can break tabbed navigation (https://github.com/gpbl/react-day-picker/issues/1567)
   const isFocusTarget = Boolean(
-    focusContext.focusTarget && isSameDay(focusContext.focusTarget, day)
+    focusContext.focusTarget &&
+      isSameDay(focusContext.focusTarget, day) &&
+      !isOutsideAndDisplayed
```
gpbl, line 113: I think we don't need to check if the date is displayed or not. I'd just check the `activeModifiers.outside`.

```suggestion
      !activeModifiers.outside
```

gpbl:
@DanielJKelly Thanks for your investigation! I could review your PR and apply some changes:

- disabled tabIndex for outside days
- added an integration test for your use case

I think we should fix this in [getInitialFocusTarget](https://github.com/gpbl/react-day-picker/blob/f291749bdd68f5eef4956cea3c26d0586b176dbf/packages/react-day-picker/src/contexts/Focus/utils/getInitialFocusTarget.ts), but this function is not aware of the outside days yet.

What do you think? Am I missing something if we don't check "displayed days"?

DanielJKelly:
> @DanielJKelly Thanks for your investigation! I could review your PR and apply some changes:
> 
> * disabled tabIndex for outside days
> * added an integration test for your use case
> 
> I think we should fix this in [getInitialFocusTarget](https://github.com/gpbl/react-day-picker/blob/f291749bdd68f5eef4956cea3c26d0586b176dbf/packages/react-day-picker/src/contexts/Focus/utils/getInitialFocusTarget.ts), but this function is not aware of the outside days yet.
> 
> What do you think? Am I missing something if we don't check "displayed days"?

@gpbl Thanks for the feedback and making the changes. I think you are right that checking "displayed days" is unnecessary. I only added that extra check because I wasn't completely sure if there was some use case I was missing for having an outside day be focusable.

PR #1582, opened 2022-10-12
Update day button aria-label to improve screen reader experience

jordanoverbye:
### Context

Hello! Firstly just wanted to say thank you for such a great library. I am a developer working on an [open source design system](https://github.com/steelthreads/agds-next) which uses `react-day-picker` for our [date picker component](https://steelthreads.github.io/agds-next/packages/forms/date-picker). 

We've been doing some accessibility user testing sessions and a few of our screen reader users have said it would be nice to include the year in the label for each of the buttons inside the calendar. Without the year rendered in the label, if they forget what year they are in they need to jump to the top of the table to find that information.

### Solution

The solution is very straight forward, we are just updating the default aria label of the day button to include the full year.

```diff
- <button aria-label="21st November (Monday)" {...} />
+ <button aria-label="21st November 2022 (Monday)" {...} />
```

### DIFF
--- a/packages/react-day-picker/src/components/Table/__snapshots__/Table.test.tsx.snap
+++ b/packages/react-day-picker/src/components/Table/__snapshots__/Table.test.tsx.snap
@@ -146,7 +146,7 @@ exports[`should render correctly 1`] = `
         class="rdp-cell"
       >
         <div
-          aria-label="1st February (Saturday)"
+          aria-label="1st February 2020 (Saturday)"
           class="rdp-day"
         >
           1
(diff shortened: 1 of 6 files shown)


gpbl:
Hey @jordanoverbye thanks for you PR. We already planned updates to the aria label in #1550. In your case, you should use the [labels](https://react-day-picker.js.org/api/types/Labels) prop:

https://codesandbox.io/s/react-daypicker-forked-8sme19?file=/src/App.tsx

```tsx
import { DayLabel, DayPicker } from 'react-day-picker';
const labelDay: DayLabel = (day, activeModifiers, options): string => {
   return format(day, 'do MMMM yyyy (EEEE)', options);
 };
function MyApp() {
  return <DayPicker labels={{ labelDay }} />;
}
```

Issue #1613, opened 2022-12-05
Bug: onDayMouseEnter does not trigger when moving from disabled date to enabled date


norbertkeresztes:
### Bug description

The `onDayMouseEnter` event handler does not trigger when the user moves the mouse from a disabled date to a date that is enabled.

### To reproduce

Observe the counter above the calendar which shows how many times the `onDayMouseEnter` event handler triggers.

https://codesandbox.io/s/react-daypicker-forked-vvcnii

#### Steps

1. Navigate to December 2022.
2. Move the mouse over 4th December which is a Sunday and it is a disabled date.
3. Observe the current count after `onDayMouseEnter`.
4. Move the mouse over 5th December without going over any other enabled dates (i.e. directly from 4th to 5th). 5th of December is a Monday and it is an enabled date.
5. Observe that the counter did not change because the `onDayMouseEnter` event handler did not trigger.

### Expected behavior

The `onDayMouseEnter` should always trigger when the user moves the mouse over an enabled date.

[EDIT]
A bit more investigation revealed that it is in fact a React issue. This post explains that there are other issues in React with `onMouseEnter` (and `onMouseLeave`): https://lecstor.com/react-disabled-button-onmouseleave/. The pure HTML version works as expected:

```
<button disabled>Disabled</button><button onmouseenter="javascript:console.log(1)">Enabled</button>
```

Is this something that can be fixed with `react-day-picker`?

norbertkeresztes:
Created a PR to add support for `onPointerEnter` and `onPointerLeave`. These events are correctly triggered by React 17.0.2 and React 18.

gpbl:
Thanks I've merged your fix into v8.4.1

PR #1614, opened 2022-12-06
feat: add onDayPointerEnter, onDayPointerLeave props

norbertkeresztes:
feat: add support for `onPointerEnter` and `onPointerLeave` event handlers.

### Context

React doesn't call the `onMouseEnter` event handler when the user is moving the mouse over a non-disabled element from a disabled element, but it does call the `onPointerEnter` event handler.

### Analysis

https://github.com/gpbl/react-day-picker/issues/1613 explains the original issue with not being able to trigger an action when the user moves the mouse over a date inside the calendar.

### Solution

Adding the two event handlers doesn't solve the issue with `onMouseEnter`, but given that `onPointerEnter` does trigger correctly in React 17.0.2 and React 18.0.0, it can be used as a workaround.

### DIFF
--- a/packages/react-day-picker/src/hooks/useDayEventHandlers/useDayEventHandlers.tsx
+++ b/packages/react-day-picker/src/hooks/useDayEventHandlers/useDayEventHandlers.tsx
@@ -3,6 +3,7 @@ import {
   HTMLProps,
   KeyboardEventHandler,
   MouseEventHandler,
+  PointerEventHandler,
   TouchEventHandler
 } from 'react';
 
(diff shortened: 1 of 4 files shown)


packages/react-day-picker/src/hooks/useDayEventHandlers/useDayEventHandlers.tsx
```diff
@@ -115,6 +119,12 @@ export function useDayEventHandlers(
   const onMouseLeave: MouseEventHandler = (e) => {
     dayPicker.onDayMouseLeave?.(date, activeModifiers, e);
   };
+  const onPointerEnter: MouseEventHandler = (e) => {
```
xram, line 122: Could we use `PointerEventHandler` for this and `onPointerLeave`?

norbertkeresztes: Updated.

gpbl:
Thanks! Release coming soon...

PR #1658, opened 2023-01-23
chore: improve accessibility

gpbl:


packages/react-day-picker/src/contexts/DayPicker/formatters/formatWeekdayName.ts
```diff
@@ -8,5 +8,5 @@ export function formatWeekdayName(
   weekday: Date,
   options?: { locale?: Locale }
 ): string {
-  return format(weekday, 'cccccc', options);
+  return format(weekday, 'cccccc', options).toUpperCase();
```
hudochenkov, line 11: This change is a breaking change. After upgrade to the latest version I had to specify formatter for weekday name to get back to capitalized values (`Mo`, `Fr`). It is impossible to make capitalization with only CSS from uppercased text.

Kosai106: Just fell for the same issue and it has triggered so many visual regressions on our end because of it... Cc @gpbl

Issue #1663, opened 2023-01-25
feat: new dropdown-buttons caption layout


gpbl:
I want to know if it's possible to have month/year navigation and arrow month navigation at the same time
![Screenshot 2023-01-01 alle 22 51 38](https://user-images.githubusercontent.com/768952/210185667-2e5fd741-a8a2-4cb1-808a-ca28d837fe05.png)
![Screenshot 2023-01-01 alle 22 51 59](https://user-images.githubusercontent.com/768952/210185668-daebf29c-f31e-48d2-92d1-4aa3a2458540.png)

I have activated the dropdown with this captionLayout="dropdown" but I don't se arrows anymore

_Originally posted by @Piero87 in https://github.com/gpbl/react-day-picker/discussions/1630_

PR #1678, opened 2023-02-03
Added caption layout for both dropdowns and buttons

seanockert:
### Context

The `captionLayout` option currently accepts EITHER month/year dropdowns OR prev/next month navigation buttons, but not both. We need both navigation controls for our app.

This option [was available in v7](https://react-day-picker-v7.netlify.app/examples/elements-year-navigation/) but not in v8

Other users would also like to have this option: https://github.com/gpbl/react-day-picker/issues/1663

### Analysis

The simplest way is to add a new option for `captionLayout`. 

I considered passing an array eg. `captionLayout={['dropdowns', 'buttons']}` but this would be a breaking change and add complexity. I kept it as a string and added `both` option.

### Solution

- Added `captionLayout="both"` to show both month/year dropdowns and prev/next month buttons
- Added test files
- Updated docs
- It's a little awkward that we need to conditionally hide the captionLabel in `CaptionNavigation.tsx`. I can rewrite this component if you prefer

https://user-images.githubusercontent.com/574163/216491198-682a1d42-3ea7-47c0-a580-8fd05f795760.mov


packages/react-day-picker/src/components/Caption/Caption.tsx
```diff
@@ -11,15 +11,18 @@ export interface CaptionProps {
...
 
 /**
  * The layout of the caption:
  *
  * - `dropdown`: display dropdowns for choosing the month and the year.
  * - `buttons`: display previous month / next month buttons.
+ * - `both`: display both month / year dropdowns and previous month / next month buttons.
  */
-export type CaptionLayout = 'dropdown' | 'buttons';
+export type CaptionLayout = 'dropdown' | 'buttons' | 'both';
```
gpbl, line 25: Consider that we could have more caption layouts in the future, thus "both" would be meaningless.

I'd rename this to:

```suggestion
export type CaptionLayout = 'dropdown' | 'buttons' | 'dropdown-button';
```

gpbl (commented):
Hey thanks for your PR: I've been testing it and works great. 

I still see the issue with the width of the caption. If I propose some changes to this PR, would you have the time to review them? Thanks.

packages/react-day-picker/src/components/Caption/Caption.tsx
seanockert: Hey, no problem, I renamed it `dropdown-buttons`

gpbl (approved):

7PH:
Thank you guys for adding this feature! We are very interested to start using it, do you know when you plan on publishing the next version on npm @gpbl ?

PR #1696, opened 2023-02-16
docs: change example to use current year

andreasottosson:
This seems like a good idea 😊
Bumped both years + 1, perhaps this should even be set to 2021 - 2025 so this will be relevant for a while.

### Context

Explain the context of this Pull Request: why are you proposing this change? What is the issue? Include links to issues or discussions.

### Analysis

Describe in details why the current code is not working as expected, or why you need to improve it.

### Solution

Explain the reasoning behind the changes implemented in this PR.


gpbl:
Thanks! Good idea.

Issue #1688, opened 2023-02-11
Bug: gridcell ARIA role incorrectly added to Day button element


gpbl:
### Discussed in https://github.com/gpbl/react-day-picker/discussions/1674

<div type='discussions-op-text'>

<sup>Originally posted by **rihardsgravis** February  2, 2023</sup>
### Bug description

The ARIA `gridcell` role is added to the Day button, but that is an incorrect role for the button since it does not represent a cell in a grid – it is in fact a child of a cell. Accessibility tests using the component are failing.

### Expected behaviour

The role for the button could be `button` or omitted at all since it's a HTM  button component and the role for it is clear.
</div>

jacobsickels:
Waiting on a fix for this!

gpbl:
Reason this is taking so long is because it is a breaking change: changing the HTML structure will likely break tests selectors and styles.

I could finally make some progress with this in PR #2089 and will cut a mayor release for this fix. More details coming soon... thanks a lot for your patience!

PR #1708, opened 2023-03-02
a11y: fix wrong role for day button

Kosai106:
### Context

Fixes https://github.com/gpbl/react-day-picker/issues/1688 by changing the button role to the correct `button` role rather than `gridcell`.

Also adds `aria-hidden="true"` on the hidden gridcell div.

### Analysis

See GitHub issue: https://github.com/gpbl/react-day-picker/issues/1688

### Solution

See "Context" section.


gpbl:
Hey thanks a lot for your contribution. Before changing the role, I'd like to add more accessibility tests - so please have patience before we merge this. 

BTW, do you know some best practice to test the screen readers?

Kosai106:
@gpbl No worries - Sadly I don't know of a way to automate screen reader tests.
We use [`axe-playwright`](https://github.com/abhinaba-ghosh/axe-playwright) to do automated accessibility tests, which I suppose could help you too.

gpbl:
OK @Kosai106 I could add the aXe tests for website. Now that the role is `button`, I need to fix the other failures first :)

packages/react-day-picker/src/hooks/useDayRender/useDayRender.tsx
```diff
@@ -107,7 +107,7 @@ export function useDayRender(
   const buttonProps = {
     ...divProps,
     disabled: activeModifiers.disabled,
-    role: 'gridcell',
+    role: 'button',
```
GeorgeTaveras1231, line 110: I believe this will break accessibility in some ways because `button` elements are not supposed to be children of `grid > rowgroup`. I would recommend testing any changes with a screen reader.

I've found that the structure that makes the most sense is to add a `div` with the role of `gridcell` and the `button` inside the `gridcell`. Additionally, using the `aria-pressed` role for the button when it is selected (since `aria-selected` is not valid attribute for `button` elements, and simply having `aria-selected` in the `gridcell` did not work in VoiceOver. (as in, it didn't announce that the day was selected).

So

```html
<table role="grid">
<!-- OMITTED HEAD -->
    <tbody role="rowgroup">
      <tr>
        <td role="presentation">
          <div role="gridcell">
            <button name="day" aria-label="April 1st, 2022" aria-pressed="false"  tabindex="0" type="button">1</button>
           </div>
        </td>
        <!-- !! SELECTED STATE -->
        <td role="presentation">
          <div role="gridcell" aria-selected="true">
            <button name="day" aria-label="April 1st, 2022" aria-pressed="true"  tabindex="0" type="button">1</button>
           </div>
        </td>
      </tr>
   </tbody>
</table>
```

GeorgeTaveras1231: You can possibly consider removing the `div` and changing `role="presentation"` to `role="gridcell"` in the `td` element. But please test in a screen reader.

GeorgeTaveras1231: 🤔 I have doubts about this approach. From my research it seems `gridcell` elements are grouping elements (except when the content is editable [think spreadsheet, for example]). In which case, clicking on the cells is just used to focus on the cell to allow the user to type. The way in which the `DayPicker` works, each Day behaves as a toggle button. Clicking on it, toggles the selected/unselected state. A `gridcell` by itself wont be able to communicate the sort of interactions each Day supports, as well as a toggle button can. 

I also looked at the ARIA role hierarchy to support this point, where you can see interactive elements generally stem from the abstract `widget` role. There you'll see the `button` role, whereas the `gridcell` role inherits from `section`, which is a `structure` role, and is mainly for grouping content.

Resources: 

- [interactive `gridcell`'](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/gridcell_role#interactive_grids_and_treegrids)
- [ARIA role hierarchy](https://www.w3.org/TR/wai-aria-1.0/rdf_model.svg)

GeorgeTaveras1231: Additionally you can consider making each day a `checkbox` or `radio` depending on the `mode` (`single` -> `radio`; `multiple` -> `checkbox`) 💡

Issue #1583, opened 2022-10-12
types: investigate type error with mode prop


gpbl:
Hi! I'm not opening new issue because I think it can be somehow related with this one 🙂 

So, after updating to 8.3.1 this issue appeared:
![Screenshot 2022-10-05 at 12 00 00](https://user-images.githubusercontent.com/106966230/194036711-c4b3b6fb-c0bc-4b0c-b3d9-951cf83767bf.png)

`mode` and `selectedDate` comes from props. Their types are defined as follows:
```
mode?: 'single' | 'multiple' | 'range';
selectedDate: DatesType
```
(and `DatesType: Date | Date[] | DateRange | undefined`, where `DateRange` is imported from `react-day-picker)

At this point I ran out of ideas how to resolve this issue. The only solution that worked, was something like this:
```
{ mode === 'single' && <DayPicker mode='single' selected={selectedDate as Date | undefined} {...otherProps} /> }
{ mode === 'multiple' && <DayPicker mode='multiple' selected={selectedDate as Date[] | undefined} {...otherProps} /> }
{ mode === 'range' && <DayPicker mode='range' selected={selectedDate as DateRange | undefined} {...otherProps} /> }
```
but to be hones it doesn't look too clean.

Maybe you would know what is the issue here and if there is a better way to resolve it than the one I showed above?

Thanks in advance 😃

_Originally posted by @bartlomiej-bykowy in https://github.com/gpbl/react-day-picker/discussions/1570#discussioncomment-3803763_

gpbl:
The prop types have been rewritten in the next major version: please check them out here http://react-day-picker.js.org/next

PR #1718, opened 2023-03-08
types: DayPickerProps to use generic constraints

ongyuxing:
### Context

Fix issue #1583, where passing a union type to the `mode` prop of `DayPicker` results in a TypeScript warning that the type of `mode` is incorrect.

### Analysis

TypeScript does not support merging property types from multiple interfaces in a union type.

### Solution

I updated the `DayPickerProps` type definition to include a generic constraint. Additionally, I provided a default type for the generic parameter to avoid breaking changes.


gpbl:
@ongyuxing this seems the solution we were looking for. Thanks for the help! I'll check the tests which have too much logs to be properly debugged...

ongyuxing:
I tried to solve the issue that `onSelect` parameter inference is incorrect when `mode` is a union type by converting `onSelect` to an intersection type.
But it requires using a utility types library called [type-fest](https://github.com/sindresorhus/type-fest) that includes a [UnionToIntersection](https://github.com/sindresorhus/type-fest/blob/bed43c1e26d9f96b949e0747d7dbabe8c643cefc/source/union-to-intersection.d.ts) type.
Would you recommend using this library and type for my purposes?

gpbl:
> Would you recommend using this library and type for my purposes?

@ongyuxing I'm interested why type-fest would be necessary? The update you propose seems to work. (The failing test is unrelated to this change.)


Also could we add an example file where we see this change working as intended? Like in website/examples.

Thanks!

ongyuxing:
@gpbl When the `mode` type is a union of 'single' | 'multiple', the `onSelect` type will also be inferred as a union of SelectSingleEventHandler | SelectMultipleEventHandler. This will result in a TypeScript error when passing in `onSelect`.

To fix this error, the type of `onSelect` needs to be converted to SelectSingleEventHandler & SelectMultipleEventHandler, allowing TypeScript to correctly merge the types of parameters and return values. For example, the first parameter will be inferred as a union of Date | Date[].

This is an example:
```typescript
import type {
  DayPickerBase,
  DayPickerDefaultProps,
  DayPickerMultipleProps,
  DayPickerRangeProps,
  DayPickerSingleProps,
  DaySelectionMode,
  SelectMultipleEventHandler,
  SelectRangeEventHandler,
  SelectSingleEventHandler,
} from 'react-day-picker'
import type { UnionToIntersection } from 'type-fest'

interface DayPickerProps<T extends DaySelectionMode = DaySelectionMode>
  extends DayPickerBase {
  mode?: T
  selected?: T extends 'single'
    ? DayPickerSingleProps['selected']
    : T extends 'multiple'
    ? DayPickerMultipleProps['selected']
    : T extends 'range'
    ? DayPickerRangeProps['selected']
    : DayPickerDefaultProps['selected']
  onSelect?: SelectEventHandler<T>
  required?: T extends 'single' ? DayPickerSingleProps['required'] : never
  min?: T extends 'multiple'
    ? DayPickerMultipleProps['min']
    : T extends 'range'
    ? DayPickerRangeProps['min']
    : never
  max?: T extends 'multiple'
    ? DayPickerMultipleProps['max']
    : T extends 'range'
    ? DayPickerRangeProps['max']
    : never
}

type SelectEventHandler<T extends DaySelectionMode = DaySelectionMode> =
  T extends 'single'
    ? SelectSingleEventHandler
    : T extends 'multiple'
    ? SelectMultipleEventHandler
    : T extends 'range'
    ? SelectRangeEventHandler
    : never

declare function DayPicker<T extends DaySelectionMode = 'default'>(
  props: Omit<DayPickerProps<T>, 'onSelect'> & {
    // Directly changing onSelect in DayPickerProps is not recommended as it will break internal references.
    onSelect: UnionToIntersection<SelectEventHandler>
  }
): JSX.Element

const DatePicker = ({ index }: { index: number }) => {
  const modes = ['single', 'multiple', 'range'] as const
  const values = [
    new Date(),
    [new Date()],
    { from: new Date(), to: new Date() },
  ]

  return (
    <DayPicker
      mode={modes[index]}
      selected={values[index]}
      onSelect={(value) => {}}
    />
  )
}
```

ongyuxing:
This is the running result without any modification:
image
This is the running result without modifying the type of onSelect:
image
This is the running result after the modification:
image

gpbl:
@ongyuxing I see it now – thanks for the clear explanation! So yes, feel free to add the `type-fest` dependency. This is a great change, looking forward to release it 👍

ongyuxing:
@gpbl It is done.🎉

gpbl:
Thanks @ongyuxing . I'm seeing interesting types errors... I am looking at them.

gpbl:
Sorry for hi-jacking your PR, I couldn't get actions working on your fork... I added a test and removed the need of type-fest in [`6571670` (#1718)](https://github.com/gpbl/react-day-picker/pull/1718/commits/6571670bc8e4e373d4a91074f613a680b64e65bb).

Could we start from this change? What is missing in the test to require the type-fest utility?

ongyuxing:
@gpbl The argument for `onSelect` is currently being inferred as `any` type. You need to add "strict: true" to the "compilerOptions" in the tsconfig.json for website. Then TypeScript will report an error.

gpbl:
> @gpbl The argument for `onSelect` is currently being inferred as `any` type. You need to add "strict: true" to the "compilerOptions" in the tsconfig.json for website. Then TypeScript will report an error.

Let investigate this in another PR. I think it will fail for other cases too, not related to this change. I remember we tried to set strict mode and it wasn't working well. I cannot find the PR right now.

gpbl:
OK I want to give it a second try :) 
I'm investigating these errors in the test: https://github.com/gpbl/react-day-picker/actions/runs/4387214125/jobs/7824904268#step:6:20

gpbl:
If I make the `onSelect` optional, without strict mode the only thing that is failing is:

```tsx
export default function Example(props: DayPickerProps) {
  return <DayPicker {...props} />;
}
```

Screenshot 2023-03-18 at 07 59 10

I don't think we can make much progress from here. In the upcoming version we should rethink the component prop considering this case.

ongyuxing:
@gpbl `DayPickerProps` doesn't match the actual `DayPicker` props, but changing it will break internal references. A less elegant solution is adding a standalone DayPickerProps type for internal use.

> I don't think we can make much progress from here. In the upcoming version we should rethink the component prop considering this case.

You're right! If the `onSelect` type is modified like this:

```typescript
type SelectEventHandler = <
  T extends DaySelectionMode = DaySelectionMode
>(
  values: {
    day?: T extends 'single' ? Date : never;
    days?: T extends 'multiple' ? Date[] : never;
    range?: T extends 'range' ? DateRange : never;
    selectedDay: Date;
    activeModifiers: ActiveModifiers;
  },
  e: React.MouseEvent
) => void;
```

It will make writing type definitions much easier.

ongyuxing:
Just as an experiment, I added an InternalDayPickerProps type.

samuelg0rd0n:
Any updates on this please?

gpbl:
@samuelg0rd0n we plan to implement this in the next mayor version (still WIP), as it may be a breaking change.

AntoineKM:
Any news here ?

XXXVII:
Any updates on this one?

PR #1730, opened 2023-03-15
feat: add custom ID to grid elements

GeorgeTaveras1231:
### Context

This allows creating combobox elements that are properly linked via aria-controls

Resources:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role

### Analysis

Currently, there is no `id` set on the grid element, and there isn't a way to set it.

### Solution

Leverage existing pattern for adding IDs to `caption` and `aria-describedby` property.

I left a comment because to a reader of the code, it may seem like the ID serves no purpose (since it isn't used internally). 

**nitpick**: I don't like that we unconditionally add an ID here, even though it is probably harmless, its not needed by most consumers, and has the side-effect of taking a naming in the global id namespace. Interested in feedback if there is a simple/elegant path for conditionally setting this id.


GeorgeTaveras1231:
@gpbl any interest in merging this or improving it?

gpbl:
Thanks @GeorgeTaveras1231 yes interested to merge it! Thanks for your contribution!

Could you share some more complete code showing how to use this id? The one I find in the comment is not clear:

```tsx
<input role="combobox" aria-haspopup="grid" aria-controls="daypicker-id-0-grid" />
```

DayPicker [can render multiple months](https://react-day-picker.js.org/basics/customization#multiple-months) (grids). Why should an input control the first grid and not the whole calendar – what about the other grids? 

Would something like this work better?

```tsx
<input role="combobox" aria-haspopup="grid" aria-controls="testId" />
<DayPicker id={testId} />
```

GeorgeTaveras1231:
> Thanks @GeorgeTaveras1231 yes interested to merge it! Thanks for your contribution!
> 
> Could you share some more complete code showing how to use this id? The one I find in the comment is not clear:
> 
> ```tsx
> <input role="combobox" aria-haspopup="grid" aria-controls="daypicker-id-0-grid" />
> ```
> 
> DayPicker [can render multiple months](https://react-day-picker.js.org/basics/customization#multiple-months) (grids). Why should an input control the first grid and not the whole calendar – what about the other grids?
> 
> Would something like this work better?
> 
> ```tsx
> <input role="combobox" aria-haspopup="grid" aria-controls="testId" />
> <DayPicker id={testId} />
> ```
@gpbl 

I think your example may have an error because the `testId` is passed as a string to `aria-controls` and as a variable to `DayPicker`.

And yea, well, the `input` element's `aria-controls` has to equal a space delimited list of all the grid elements. In the case where multiple months are rendered, the caller can explicitly pass 2 or more IDs following the naming convention implemented by the current pattern.


Examples: Single month
```jsx
<>
  <input role="combobox" aria-haspopup="grid" aria-controls="daypicker-id-0-grid" />
  <DayPicker id="daypicker-id" />
</>
```

Examples: Multiple months
```jsx
<>
  <input role="combobox" aria-haspopup="grid" aria-controls="daypicker-id-0-grid daypicker-id-1-grid" />
  <DayPicker id="daypicker-id" numberOfMonths={2} />
</>
```

GeorgeTaveras1231:
@gpbl just a status update on this. Was my explanation clear and sufficient? Is there anything missing/unclear in the example?

packages/react-day-picker/src/components/Table/Table.tsx
```diff
@@ -42,6 +44,7 @@ export function Table(props: TableProps): JSX.Element {
   const FooterComponent = components?.Footer ?? Footer;
   return (
     <table
+      id={props.id}
```
gpbl, line 47: I don't think we need a prop here. Since we have the `props.displayMonth`, we can use it to calculate the unique id string, which could be `id + year + month`:

```suggestion
      id={id ? `{id}-${props.displayMonth.fullYear}-${props.displayMonth.month}` : undefined|
```

Where `id` comes from the `useDayPicker` hook:

```diff
-    ISOWeek
+    ISOWeek,
+    id
  } = useDayPicker();
```

gpbl (changes requested):
Thanks @GeorgeTaveras1231 ! I understand the use case now. I'm OK adding an `id` attribute to the rendered grid. 

I think your implementation could be simplified if we use year and month to create the unique id, as opposed to the use of `displayIndex`.

PS. could you `yarn lint --fix` before pushing? Thanks!

packages/react-day-picker/src/components/Table/Table.tsx
GeorgeTaveras1231: @gpbl If I'm understanding this correctly, this means that the id will continue to change for the displayed displayed tables depending on the displayed months. This is probably not a good idea because it means that we have to manage the displayed month state outside of the datepicker to be able to use the ids anywhere else. The current implementation is good because the displayed months are irrelevant, we just need the ids of the displayed tables.

gpbl: Oh sure, it makes sense. Thanks for the heads up.

gpbl (changes requested):
@GeorgeTaveras1231 would you address my changes so we can move forward? Thanks a lot for your help!

GeorgeTaveras1231:
@gpbl I've made the updates you've requested. Let me know if there's anything else I should do.

PR #1769, opened 2023-04-22
Add undefined to optional properties

fedemp:
### Context

Allow undefined for optional properties when exactOptionalPropertyTypes is enabled

See http://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes

### Analysis

I'm using the [strictest tsconfig.json](https://github.com/tsconfig/bases/blob/main/bases/strictest.json) and I'm not allow to do

```tsx
<DayPicker month={selectedDate ?? undefined}/>
```


gpbl:
Thanks @fedemp, this is great! All tests are passing!

Could you share some more code examples to reproduce this issue? Your use case only seems to consider the `month` prop. 

I'd like to better test the impact of this change when the TS strict mode is set: how?

We've been trying this change some time ago (see https://github.com/gpbl/react-day-picker/pull/1456) but it got difficult for reasons I don't remember.

fedemp:
Hello, @gpbl
Even though these issues look similar, goals are different.

From what I see in the discussion, those changes were more about migrating `react-day-picker` to support `exactOptionalPropertyTypes` *as a project* but you (plural) had trouble consuming `date-fns`, which doesn't accept `undefined` for most of its methods. IMO in this case it's up to `date-fns` to update their typings cause I'm sure `undefined` is a valid value.

My proposal is more about `react-day-picker` as a dependency of other projects which set `exactOptionalPropertyTypes`. Those other projects should be allowed to pass `undefined` values cause your project can handle them fine.

You know the old saying _"be conservative in what you send, be liberal in what you accept from others"_.

gpbl:
Closing this as v9, currently under development, will address this issue,

Issue #1507, opened 2022-08-16
docs: example with time picker


gpbl:
It is a frequent request to see how DayPicker should work with time. We should add an example to the docs...

### Discussed in https://github.com/gpbl/react-day-picker/discussions/1474

<div type='discussions-op-text'>

<sup>Originally posted by **marcostaborda** June 23, 2022</sup>
How to use `react-day-picker` with a time picker?

Does anyone have used `react-day-picker` with time selector?</div>

Erik-McKelvey:
I can work on this

Erik-McKelvey:
I could show an example of integrating it with this time picker https://github.com/wojtekmaj/react-time-picker?

PR #1772, opened 2023-04-23
docs: Time Picker Example

Erik-McKelvey:
### Context

Added an example using a time picker to the docs as outline in issue #1507

Here is a code sandbox with my solution: https://codesandbox.io/s/react-daypicker-forked-h6oqo8?file=/src/App.tsx

### Analysis

As discussed, it is a frequently asked question on how to use the DayPicker package with a time selection

### Solution

Added a simple example of using an `<input type="time">` and parsing the input to add the hours and minutes to the same Date React state as the DayPicker.


gpbl:
Cool! Thanks a lot!

PR #1787, opened 2023-05-10
feat(dropdown): Expose the raw dropdown options

oof2win2:
### Context

This PR attempts to solve an issue when you want to use your own implementation of select with the dropdown selectors, where you would want to be able to access the raw values that should be chosen.

### Analysis

The current code does not work as it only provides the already pre-made selectors as a `ReactNode` prop in `children`

### Solution

I have only provided the existing data that has worked before. Also had to change the typing of the formatter from `ReactNode` to `string`, however, it might make sense to create a separate object which returns only strings, such as `stringFormatters` on the `DayPickerContextValue` type.


oof2win2:
Changed it to use `stringFormatters` param instead so it doesn't conflict with any current implementations.

oof2win2:
Hi, any chance to get this merged? Would find it relatively useful for adding custom month or year selectors

.gitpod.yml
gpbl, line 1: This file should be removed from the PR.

packages/react-day-picker/src/components/Dropdown/Dropdown.tsx
```diff
@@ -3,13 +3,19 @@ import React from 'react';
 import { IconDropdown } from 'components/IconDropdown';
 import { useDayPicker } from 'contexts/DayPicker';
 
+type DropdownOption = {
+  name: string;
```
gpbl, line 7: - This type should be exported and have some comment like the interface below.
- What is `name` used for? Maybe you mean label?

```suggestion
  label: string;
```

packages/react-day-picker/src/components/Dropdown/Dropdown.tsx
```diff
@@ -3,13 +3,19 @@ import React from 'react';
...
+  value: string | number;
```
gpbl, line 8: Should be optional / undefined?

```suggestion
  value?: string | number | undefined;
```

packages/react-day-picker/src/components/YearsDropdown/YearsDropdown.tsx
```diff
@@ -64,6 +65,10 @@ export function YearsDropdown(props: YearsDropdownProps): JSX.Element {
       onChange={handleChange}
       value={displayMonth.getFullYear()}
       caption={formatYearCaption(displayMonth, { locale })}
+      options={years.map((m) => ({
+        value: m.getMonth(),
+        name: stringFormatYearCaption(m, { locale })
```
gpbl, line 70: Why can't we use `formatters.formatMonthCaption`? Or do you meant to add `formatMonthOption` or something?

packages/react-day-picker/src/types/Formatters.ts
```diff
@@ -22,8 +27,29 @@ export type Formatters = {
   formatWeekdayName: DateFormatter;
 };
 
+/** Represent a map of formatters used to render localized content. */
+export type StringFormatters = {
```
gpbl, line 31: What is different from `Formatters` above?

gpbl (commented):
Thanks @oof2win2, I could review your PR.

I understand your proposal here, but not the context. Could you give us some more details? What is the feature for? Some example usage?

Is passing `components={{ Dropdown: CustomDropdown }}` not working for you?

packages/react-day-picker/src/components/Dropdown/Dropdown.tsx
oof2win2: Yeah, it was meant to be the label of the value

packages/react-day-picker/src/components/YearsDropdown/YearsDropdown.tsx
oof2win2: The issue is that `formatMonthCaption` has the return type of a JSX element. I created this because I did not want to introduce a breaking change. Changing it to a `string` return type would however be trivial, as it already returns a string - just has the wrong type information for this use. What option do you think is better?

packages/react-day-picker/src/types/Formatters.ts
oof2win2: Same thing as above - the issue is that the current formatters return a React node instead of a string. Technically speaking, the type information there is wrong as the formatters return strings - not React nodes. It hopefully shouldn't conflict if it were to be changed to a `string` return type, however, I did this just in case.

packages/react-day-picker/src/components/YearsDropdown/YearsDropdown.tsx
gpbl: > The issue is that `formatMonthCaption` has the return type of a React node. I created this because I did not want to introduce a breaking change.

Oh I see it now, thanks for the explanation. There are many ways for skipping a breaking change (e.g. we could add a new `formatDropdownOption: () => string` function to the formatters).

gpbl:
> The main issue with the use of passing `components={{ Dropdown: CustomDropdown }}` is that `children` is already pre-defined to be an array of HTML `<option>` components, so it can't be replaced with some other component

Interesting. Let me give a look thanks!

gpbl:
> The main issue with the use of passing `components={{ Dropdown: CustomDropdown }}` is that `children` is already pre-defined to be an array of HTML `<option>` components, so it can't be replaced with some other component

Could you please elaborate more on this? Do you have en example of code where you are blocked? 

I think I understand the issue you want to solve, but the proposed solution seems overkill to me (e.g. we won't add a `stringFormatters` prop). If we could have a better context we could find a less invasive approach.

oof2win2:
I'm using [RadixUI Select components](https://www.radix-ui.com/docs/primitives/components/select) instead of the native select. The issue is that I need to use their SelectItem for it to work, which I am not able to do with the way the current Dropdown props are set.

gpbl:
> I'm using [RadixUI Select components](https://www.radix-ui.com/docs/primitives/components/select) instead of the native select. The issue is that I need to use their SelectItem for it to work, which I am not able to do with the way the current Dropdown props are set.

I think adding a custom component to replace the option element would work better. Trying to implement a solution here:

https://github.com/gpbl/react-day-picker/pull/1800

gpbl:
> I'm using [RadixUI Select components](https://www.radix-ui.com/docs/primitives/components/select) instead of the native select. The issue is that I need to use their SelectItem for it to work, which I am not able to do with the way the current Dropdown props are set.

@oof2win2 I suggest you to use [a custom Caption](https://react-day-picker.js.org/guides/custom-components#custom-caption) instead. With the [DayPicker hooks](https://react-day-picker.js.org/guides/custom-components#daypicker-hooks) you can access to the DayPicker props and navigate the calendar:

https://codesandbox.io/s/react-daypicker-forked-iplbfb?file=/src/App.tsx

There you could replace the select with the one from RadixUI.

I agree this part should be better documented...

oof2win2:
Oh neat, thanks!

PR #1794, opened 2023-05-26
Only announce month when the month changes

GeorgeTaveras1231:
__NOTE: Implemented this quick and dirty to gather feedback. Tests and types will be addressed separately.__

### Context/Analysis

aria-live as it is implemented currently causes usability issues for this component.

You cannot render it on a page without it interfering what the screen reader user will hear. `aria-live` forces the screen reader to be high-jacked by the content on this element regardless of the content order. This is a bad user experience. Instead, the sreen-reader should wait until the screen reader cursor reached the datepicker to announce this content. Additionally, announce it when the user interacts with the next and previous butons.

This causes accessibility issues for a number of use-cases where we defer the mounting of the datepicker.

1. Datepicker is in a modal
2. Datepicker is in a dropdown as part of a combobox
3. Virtualization -- when unmounting components that are out of the view to reduce DOM overhead.
.

### Solution

Only apply the `aria-live` property for a second after the month has changed.

### DIFF
--- a/packages/react-day-picker/src/components/CaptionLabel/CaptionLabel.tsx
+++ b/packages/react-day-picker/src/components/CaptionLabel/CaptionLabel.tsx
@@ -16,13 +16,14 @@ export function CaptionLabel(props: CaptionLabelProps): JSX.Element {
     locale,
     classNames,
     styles,
-    formatters: { formatCaption }
+    formatters: { formatCaption },
+    shouldAnnounceMonthChange
   } = useDayPicker();
   return (
     <div
       className={classNames.caption_label}
       style={styles.caption_label}
-      aria-live="polite"
+      aria-live={shouldAnnounceMonthChange ? 'polite' : undefined}
       role="presentation"
       id={props.id}
     >
--- a/packages/react-day-picker/src/contexts/DayPicker/DayPickerContext.tsx
+++ b/packages/react-day-picker/src/contexts/DayPicker/DayPickerContext.tsx
@@ -44,6 +44,7 @@ export interface DayPickerContextValue extends DayPickerBase {
   numberOfMonths: number;
   styles: Styles;
   today: Date;
+  shouldAnnounceMonthChange: boolean;
 }
 
 /**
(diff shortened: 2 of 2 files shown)


gpbl:
@GeorgeTaveras1231 thanks for the help! I understand from your issue that screen readers read the `aria-live` region when they should not. 

I don't understand how to replicate the issue.

Could you first send us an example of this behavior by reproducing it in a Codesandbox? https://codesandbox.io/s/react-daypicker-vl4ry

Instead of changing the aria-live attribute, should having an `aria-hidden` container work to not have the screen reader reading the aria-live region>

E.g.


```tsx
function App() {
	return (
		<div aria-hidden>
			<DayPicker />
		</div>
	)
}
```

GeorgeTaveras1231:
@gpbl I'm glad you challenged me to reproduce this. This forced me to find the root of the issue. If you want, you can look at my sandbox where I was able to implement a [perfectly working modal datepicker](https://codesandbox.io/s/react-daypicker-forked-l0bq3r?file=/src/App.tsx:141-212) 😅 

The use case I was running into was when placing the date picker in a modal. When I would open the modal, instead of Voice Over announcing that I was in a modal, it would just say the date ("April 2023"). 

The issue is a design issue and an implementation issue. In our implementation, we don't have a design for a close button for the modal, which is the source of the issue. The appropriate solution is to force focus on an element in the modal (possibly the close button). This forces voice-over to have something to announce that takes precedence over the `polite` live alerts. 

I'm happy to close this PR in favor of this solution which is more elegant. 

However, I will take this opportunity to raise a discussion I posted before which ties into this, which is to improve focus management capabilities of the datepicker. There are cases where the datepicker (or its container) are animated into place, or for other reasons, we may only use CSS to hide and show it, in which case, we need to be able to imperatively move focus to it. https://github.com/gpbl/react-day-picker/discussions/1735

gpbl:
Thanks @GeorgeTaveras1231 for the update. Glad it worked for you!

PR #1843, opened 2023-07-20
chore: add dates as an option to formatWeekNumber

amakaa:
### Context

Proposing this because I want to be able to recalculate the week number based on the date

### Analysis

When using `formatWeekNumber`, I only have access to the `locale` and would like to be able to format the week based on the date as well

### Solution

I have added `dates` to the object passed to `formatWeekNumber`

### DIFF
--- a/src/components/WeekNumber/WeekNumber.tsx
+++ b/src/components/WeekNumber/WeekNumber.tsx
@@ -39,7 +39,7 @@ export function WeekNumber(props: WeekNumberProps): JSX.Element {
     );
   }
 
-  const label = labelWeekNumber(Number(weekNumber), { locale });
+  const label = labelWeekNumber(Number(weekNumber), { locale, dates });
 
   const handleClick: React.MouseEventHandler = function (e) {
     onWeekNumberClick(weekNumber, dates, e);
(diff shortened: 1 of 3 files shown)


gpbl:
@amakaa thanks for your PR!  I'm closing this because you don't need to change the code of the formatter to get the dates belonging to the week. Formatters are only to format a value and they shouldn't perform other calculations.

Instead, you want to use a [custom component](https://react-day-picker.js.org/guides/custom-components) replacing the default `WeekNumber` component.

See for example: https://codesandbox.io/s/rough-violet-h96s4n?file=/App.tsx 

```tsx

function CustomWeekNumber(props: WeekNumberProps) {
  return (
    <span title={`1st week day is ${props.dates[0].toLocaleDateString()}`}>
      <WeekNumber {...props} number={100} />
    </span>
  );
}
  return (
    <DayPicker
      showWeekNumber
      components={{
        WeekNumber: CustomWeekNumber
      }}
    />
  );
}
```

- I set the week number to be always `100`
- I access to the dates of that week via `props.dates`.

Also, maybe you don't need to change the week number. Please check this out: https://react-day-picker.js.org/basics/customization#showing-the-week-numbers

PR #1876, opened 2023-08-11
fix: range not selected when missing the `from` value

Sam-Apostel:
### Context
A user might want to filter all items before a specific date and modify the `selected` state outside `react-day-picker`. 
While this works perfectly when only `from` is set. But when only `to` is set, the calendar doesn't reflect this state.
The next time a user selects a date, the state is affected like when no values were set at all.
Discussion: https://github.com/gpbl/react-day-picker/discussions/1875

### Analysis
`selected` on a range picker accepts a DateRange where `from` is undefined but `to` is a Date. 
Since this works in one direction and the types allow this state, this component should reflect the other direction too.

### Solution

To prevent current behaviour changes, All of the logic additions only take effect when `to` is set and `from` is not.
From within this component, this is an unreachable state, but when the state is controlled externally, it is now reflected in the internal state.

All other code changes and restructuring were made to benefit typescript.
No test were changed, few were added for specific state.

### DIFF
--- a/src/contexts/Modifiers/utils/isDateInRange.ts
+++ b/src/contexts/Modifiers/utils/isDateInRange.ts
@@ -5,21 +5,21 @@ import { DateRange } from 'types/Matchers';
 /** Return `true` whether `date` is inside `range`. */
 export function isDateInRange(date: Date, range: DateRange): boolean {
   let { from, to } = range;
-  if (!from) {
-    return false;
+  if (from && to) {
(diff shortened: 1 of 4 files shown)


gpbl (approved):

gpbl:
This fix has been released as https://github.com/gpbl/react-day-picker/releases/tag/v8.8.1. (Not sure why the release doesn't appear here)

PR #1877, opened 2023-08-13
Add flag to disable selecting days range with disabled days in between.

trabeast:
### Context

Added flag to disable selecting range of dates with disabled dates in between.

### Solution

During `getInternalModifers` execution, if the `disableNonConsecutiveDates` is set true, this will create an array of `DateBefore` and `DateAfter` to be added to `disabled`. 

```typescript
export function createDisabledNonConsecutiveDates(
  disabled: Matcher[],
  from: Date
): (DateBefore | DateAfter)[] {
  const disabledNonConsecutiveDates: (DateBefore | DateAfter)[] = [];

  let earliest: Date | undefined;
  let latest: Date | undefined;

  disabled.filter(isDateRange).forEach((range) => {
    // We only need to add the latest 'from' date in the range.
    // As when we disable it with DateBefore object, others will be included.
    if (range.from && range.from < from) {
      (!latest || isAfter(range.from, latest)) && (latest = range.from);
    }
    // We only need to add the earliest 'to' date in the range.
    // As when we disable it with DateAfter object, others will be included.
    if (range.to && range.to > from) {
      (!earliest || isBefore(range.to, earliest)) && (earliest = range.to);
    }
  });

  latest && disabledNonConsecutiveDates.push({ before: latest });
  earliest && disabledNonConsecutiveDates.push({ after: earliest });

  return disabledNonConsecutiveDates;
}
```

### DIFF
--- a/src/contexts/Modifiers/utils/createDisabledNonConsecutiveDates.test.ts
+++ b/src/contexts/Modifiers/utils/createDisabledNonConsecutiveDates.test.ts
@@ -1,0 +1,51 @@
+import { createDisabledNonConsecutiveDates } from './createDisabledNonConsecutiveDates';
+import { addDays } from 'date-fns';
+import { isDateAfterType, isDateBeforeType } from '../../../types/Matchers';
+
+describe('when today is before disabled dates', () => {
+  const disabledNonConsecutiveDates = createDisabledNonConsecutiveDates(
+    [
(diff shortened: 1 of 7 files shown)


trabeast:
To have option of prohibiting selecting date ranges that contains disabled dates.

PR #1878, opened 2023-08-13
feat: add flag to disable selecting days range with disabled days in between.

trabeast:
### Context

Allow option for users to disable selecting range of dates with disabled dates in between.

### Solution

With `limitSelectRangeWithoutDisabledDays` property set to true we will compute for an array of `DateBefore` and `DateAfter` to be inserted to the final  `InternalModifiers.disabled` based from the `disabled` and `selected.from` properties.

```typescript
/** Return a list of {@link DateBefore} and {@link DateAfter} that does not contain disabled days. */
export function createDateRangeWithoutDisabledDaysBasedFromStartDate(
  disabled: Matcher[],
  startDate: Date
): (DateBefore | DateAfter)[] {
  const disabledNonConsecutiveDates: (DateBefore | DateAfter)[] = [];

  let earliest: Date | undefined;
  let latest: Date | undefined;

  disabled.filter(isDateRange).forEach((range) => {
    // We only need to add the latest 'from' date in the range.
    // As when we disable it with DateBefore object, others will be included.
    if (range.from && range.from < startDate) {
      (!latest || isAfter(range.from, latest)) && (latest = range.from);
    }
    // We only need to add the earliest 'to' date in the range.
    // As when we disable it with DateAfter object, others will be included.
    if (range.to && range.to > startDate) {
      (!earliest || isBefore(range.to, earliest)) && (earliest = range.to);
    }
  });

  latest && disabledNonConsecutiveDates.push({ before: latest });
  earliest && disabledNonConsecutiveDates.push({ after: earliest });

  return disabledNonConsecutiveDates;
}
```

### DIFF
--- a/src/contexts/Modifiers/utils/createDateRangeWithoutDisabledDaysBasedFromStartDate.test.ts
+++ b/src/contexts/Modifiers/utils/createDateRangeWithoutDisabledDaysBasedFromStartDate.test.ts
@@ -1,15 +1,16 @@
-import { createDisabledNonConsecutiveDates } from './createDisabledNonConsecutiveDates';
+import { createDateRangeWithoutDisabledDaysBasedFromStartDate } from './createDateRangeWithoutDisabledDaysBasedFromStartDate';
 import { addDays } from 'date-fns';
 import { isDateAfterType, isDateBeforeType } from 'types/Matchers';
 
 describe('when today is before disabled dates', () => {
-  const disabledNonConsecutiveDates = createDisabledNonConsecutiveDates(
-    [
-      { from: addDays(new Date(), 3), to: addDays(new Date(), 8) },
-      { from: addDays(new Date(), 12), to: addDays(new Date(), 17) }
-    ],
-    new Date()
-  );
+  const disabledNonConsecutiveDates =
+    createDateRangeWithoutDisabledDaysBasedFromStartDate(
+      [
+        { from: addDays(new Date(), 3), to: addDays(new Date(), 8) },
+        { from: addDays(new Date(), 12), to: addDays(new Date(), 17) }
+      ],
+      new Date()
+    );
 
   test('should only create one disabled matcher', () => {
     expect(disabledNonConsecutiveDates).toHaveLength(1);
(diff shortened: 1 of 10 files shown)


gpbl:
Hi @trabeast, thanks for your contribution! 

Could you clarify better the problem you are trying to solve here? What is happening / what is expected instead? It would help reviewing your changes.

ThibautCuchet:
Hi @gpbl and @trabeast , I have the same issue in a project.
Currently, if you have a DayPicker in mode `range` with disabled dates you cannot click on them but they can be included in a range.
I think that @trabeast is trying to add a new parameter to prevent user to have a disabled date in the range.
Here is a short video showing that case.
https://github.com/gpbl/react-day-picker/assets/19268304/76664bfe-1bf8-454d-b550-cf2a0e956892

trabeast:
Hi @ThibautCuchet thanks for the explanation.

@gpbl So yes, it is a feature request to basically disable user selecting a date range where there exists a disabled date in between. How the logic works is that when you select  a start date, it will disable other dates that are before or after disabled days. Only consecutive non-disabled days can be selected from your start date.

Example:

Days: 1, 2, 3, 7, 8, 9 are disabled.

- When user selects a date across 4 and 6, the dates after 9 and before 1 (previous month) will get disabled.
- When user selects a date after the 9th (example 10), all dates 9 and before will get disabled.

This is useful for use case where you need the user to select a range for reservations for example.

gpbl:
@ThibautCuchet @trabeast thanks for the help!

This example should be a proof-of-concept of the behavior we want to implement. Am I correct?

https://codesandbox.io/s/recursing-bouman-r7sz4n?file=/src/App.tsx 

1. select the first range day
2. select the last range day _including the disabled date_

the range is reset to the clicked date, without including the disabled day.

https://github.com/gpbl/react-day-picker/assets/120693/7a990678-4ec0-4665-8b70-f652eaf3b107

ThibautCuchet:
It's a really good implementation. I will use it until the package has a param for it.

gpbl:
Thanks! Follow https://github.com/gpbl/react-day-picker/issues/1885 for updates about this bug.

PR #1890, opened 2023-08-25
feat: add formatClassNames function to format CSS class names

tonyspb:
### Context
DayPicker accepts class names for styling, but class names composition doesn't work well with TailwindCSS.

### Analysis
Class names are concatenated as is. No options to modify composed selected day, range start day, etc. 

### Solution
DayPicker accepts `formatClassNames/formatModifiersClassNames` functions to format corresponding class names. Defaults are identity functions.


gpbl:
@tonyspb, thanks for your contribution! 

I'm not familiar with tailwind, so I'm not sure about the problem this PR should solve. Could you provide an example? Thanks!

PR #1917, opened 2023-09-29
feat: Introduced custom components "Months"

pawelgoc:
### Context

As developer I might want to wrap `<DayPicker />` and use [children API](https://react.dev/reference/react/Children) to map over months. For instance I would want to extend pagination to behave like carousel - this seems good use case for using `Children.map`. 
To achieve it we need to be able to host collection of  `<Month />` elements. That's why replacing hardcoded months container by public customisable `<Months />` open additional possibility how library can be consumed. 

### Analysis

```
const AnyComponent = ({children}) => {
  return <p>Count: {Children.count(children)}</p>
}
---
<AnyComponent>
  <DayPicker numberOfMonths={4} />
</AnyComponent/>

--- output
<p>Count: 1</p>

--- with custom Months
const CustomMonths = ({children}) => (<Months><AnyComponent>{children}</AnyComponent></Months>)

<DayPicker numberOfMonths={4} components={Months: CustomMonths}>

--- output
  <p>Count: 4</p>
```

Current implementation will always return only one children because in **Root** component **DIV** element is wrapper on top of months. 

### Solution

With custom `<Months />` component we can have direct access to array children or to do any other customisation. It means library is more flexible without exposing core functionality.
We already can find dedicated `classes and styles` for `months` to allow customisation, that's makes me think that custom component `<Months />` also make sense.

Other approach I've tried was to compose `DayPicker` form existing components like:

```
const DayPicker = () => {
  <Providers>
    <Root>
       <CustomMonths>{months.map(() => <Month />)}</CustomMonths>
    </Root>
  </Providers>
}
``` 
but it is also impossible because `<Month />` is private (I believe for good reasons).

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "@testing-library/react": "^14.0.0",
     "@testing-library/user-event": "^14.5.1",
     "@types/jest": "^29.5.5",
-    "@types/node": "^18.17.19",
+    "@types/node": "^20.6.5",
     "@types/react": "^18.2.22",
     "@types/react-dom": "^18.2.7",
     "@types/testing-library__jest-dom": "^5.14.9",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -39,8 +39,8 @@ importers:
         specifier: ^29.5.5
         version: 29.5.5
       '@types/node':
-        specifier: ^18.17.19
-        version: 18.17.19
+        specifier: ^20.6.5
+        version: 20.6.5
       '@types/react':
         specifier: ^18.2.22
         version: 18.2.22
@@ -85,7 +85,7 @@ importers:
         version: 5.11.1(eslint@8.50.0)(typescript@4.9.5)
       jest:
         specifier: ^29.7.0
-        version: 29.7.0(@types/node@18.17.19)(ts-node@10.9.1)
+        version: 29.7.0(@types/node@20.6.5)(ts-node@10.9.1)
       jest-environment-jsdom:
         specifier: ^29.7.0
         version: 29.7.0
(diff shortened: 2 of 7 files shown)


gpbl:
Interesting, thanks for your contribution! Looking at it.

pawelgoc:
> Interesting, thanks for your contribution! Looking at it.

Hi @gpbl , do you have any feedback regarding this PR?
Curious if you are considering to merge it and eventually when (not sure If I should wait or rather rely on fork)?

gpbl:
Thanks @pawelgoc I wanted to update a bit the code to match our code style. Thanks for you contribution, coming soon on npm!

gpbl:
(PS. if you are trying to add animation to the calendar - it is something we plan to add in v9. Would love to see your implementation)

pawelgoc:
> (PS. if you are trying to add animation to the calendar - it is something we plan to add in v9. Would love to see your implementation)

I doubt our impl. will fit to library as it would require additional dependency.

Idea is to utilise pagination build in **DayPicker** + any library that serve **Carousel** functionality.
On DayPicker prop numberOfMonths is set to `n + 2` where `n` is amount of months that will be visible in Carousel and remaining 2 months are just left and right offset. Having offsets will allow to animate transition between months and is done by Carousel.

Of course there is needed additional sync between both components to resolve what is current month and slideIndex.
There was also challange how to prevent focus on day that is in offset month which can be triggered by keyboard navigation (library API not allow to control / prevent `focus()` and default behaviour of focus is to scroll to element which destroy calculated positions by Carousel)

**Simplified example:**

```
const CustomMonths = ({children, props}) => {
    return (
      <Months>
        <Carousel {...props} >
          {children}
        </Carousel>
      </Months>
    );
}

const Component = ({...props}) => (<DayPicker
        {...props}
        numberOfMonths={numberOfMonthsPerPage + 2}
        month={currentMonth}
        components={{ ...props.components, Months: CustomMonths }}
  />)
```

**How it looks:**

day-picker-horizontal

PR #1947, opened 2023-10-18
Fixing incompatible usage of PropsWithChildren so that typing works with React 17

binhpv:
### Context

The newest version of react-day-picker is not compatible with @types/react 17.

### Analysis

In the last version of the library some props were declare with PropsWithChildren. This interface originally in @types/react 17 is:

```
    type PropsWithChildren<P> = P & { children?: ReactNode | undefined };
```

but in @types/react version 18 it has changed to

```
    type PropsWithChildren<P = unknown> = P & { children?: ReactNode | undefined };
```

Please note the new default of unknown. This mean new typing relies on @types/react@18 doesn't have to provide the typing when using PropsWithChildren, however it is a must with @types/react@17. The usage of PropsWithChildren in react-day-picker without the argument for the generic type make it incompatible with projects using React 17.

### Solution

When using PropsWithChildren, always include argument for the generic typing so that it's also compatible with React 17 codebase. 
In this PR I also change some places where the prop is declare as equal to PropsWithChildren (e.g. MonthsProps) to explicitly type as { children: ReactNode }. This ensure that children is required, and there is no need to rely on PropsWithChildren since there is no additional prop.

### DIFF
--- a/src/components/Months/Months.tsx
+++ b/src/components/Months/Months.tsx
@@ -1,9 +1,9 @@
-import { PropsWithChildren } from 'react';
+import { ReactNode } from 'react';
 
 import { useDayPicker } from 'contexts/DayPicker';
 
 /** The props for the {@link Months} component. */
-export type MonthsProps = PropsWithChildren;
+export type MonthsProps = { children: ReactNode };
 
 /**
  * Render the wrapper for the mont grids.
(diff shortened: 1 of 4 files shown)


gpbl:
Thanks so much @binhpv! I completely missed that. Publishing soon on npm...

binhpv:
Thanks @gpbl  for your quick follow up. Cheers!

PR #1964, opened 2023-11-02
fix(types): cannot find module types/Matchers

jerodfritz:
…ponding type declarations.

### Context

when using this within a nextjs app the error Type error: Cannot find module 'types/Matchers' or its corresponding type declarations.

### Analysis

Something is causing a conflict with the tsconfig path matching

### Solution

Include the Matchers similar to Modifiers to prevent reliance on types/

### DIFF
--- a/src/types/EventHandlers.ts
+++ b/src/types/EventHandlers.ts
@@ -6,7 +6,7 @@ import {
   TouchEvent
 } from 'react';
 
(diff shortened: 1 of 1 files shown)


gpbl:
Hey thanks for this! I must have missed it out.

gpbl (approved):

jerodfritz:
Would you be able to release this on npm?

PR #1969, opened 2023-11-08
fix(docs): remove anchor from the Readme image

neicore:
### Context

In the current documentation site, the header image looks like a component that seems clickable but opens the same site in the new tab when clicked which leads to a confusing user experience.

### Solution

I have removed the surrounding anchor tag around the picture element. I understand that it made sense to have it in the readme because it was in Github but since the same readme is parsed and used in the landing page I think the other links in the readme should be enough to help a visitor in Github to get to the website.


gpbl:
Thanks!

PR #1978, opened 2023-11-26
docs: matcher api reference page uses incorrect object syntax

### LINKED ISSUE #1976: Fix: Docs. Type alias: Matcher api reference page uses incorrect ";" in the examples
### Docs issue description

In the code snippets provided in the documentation [Type alias: Matcher](https://react-day-picker.js.org/api/types/Matcher) we encounter Date objects which end with `;`, which is an incorrect syntax. 

### Method of fixing
Just remove all instances of `;` in the objects matchers such as `{ after: new Date(2019, 1, 2); };`

janaiscoding:
### Context

Closes #1976 
I would like to improve the website docs as per the contributing guide's website proofreading line. 

### Analysis

It's important that the documentation is consistent and correct, which is why I would like to add these small fixes to the typos

### Solution

Removed the incorrect `;` from the following lines:  
`const arrayMatcher: Matcher = [new Date(2019, 1, 2);, new Date(2019, 1, 4)];`
`const afterMatcher: DateAfter = { after: new Date(2019, 1, 2); };`
`const beforeMatcher: DateBefore = { before: new Date(2019, 1, 2); };`
`const rangeMatcher: DateRange = {
   from: new Date(2019, 1, 2);,
   to: new Date(2019, 1, 5)
};`
on the [Type alias: Matcher](https://react-day-picker.js.org/api/types/Matcher) API reference documentation

### DIFF
--- a/src/types/Matchers.ts
+++ b/src/types/Matchers.ts
@@ -16,12 +16,12 @@
  * const dateMatcher: Matcher = new Date();
  *
  * // will match the days in the array
- * const arrayMatcher: Matcher = [new Date(2019, 1, 2);, new Date(2019, 1, 4)];
+ * const arrayMatcher: Matcher = [new Date(2019, 1, 2), new Date(2019, 1, 4)];
  *
  * // will match days after the 2nd of February 2019
- * const afterMatcher: DateAfter = { after: new Date(2019, 1, 2); };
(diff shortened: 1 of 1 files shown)


gpbl:
Awesome, thanks!

PR #1984, opened 2023-11-30
docs: fix typo

toby-brilliant:
### Context

There are only 7 days in a week, so either the actual range is 0-6 or Sunday is both day 0 and 7

### Analysis

Trying this out, I don't believe days higher than 6 do anything. Updating the comment to reflect the valid range.

### Solution

Update the comment


gpbl:
Thanks!

PR #1987, opened 2023-12-04
docs: adjust example wording: "the today's date"

toby-brilliant:
### Context
image

### Analysis

None

### Solution

"the today's date" is a bit awkward, so update it to just say "today's date".


gpbl:
@toby-brilliant thanks!

PR #2002, opened 2023-12-22
Update date-fns to 3.0.5

martinez-hugo:
### Context

Needs to update date-fns

### Analysis

Just a dependency update

### Solution

Editing package.json


gpbl:
Hey @martinez-hugo thanks for your PR. As you have seen, there were some other changes to do to make the package compliant to date-fns v3. I've fixed these in https://github.com/gpbl/react-day-picker/pull/2004. Thanks again!

PR #2254, opened 2024-07-07
docs: update selection-modes.mdx

xardit:
Wrong example on the documentation


gpbl:
Thanks! I always wonder - should this be called `daysOfWeek` or `dayOfWeek`? 🤔

PR #2298, opened 2024-07-23
fix: typo in css rdp-hidden

AlessioDP:
## What's Changed

Renamed css variable `end-range_start-color` to `--rdp-range_start-color`. This error is preventing project build.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

PR #2307, opened 2024-07-24
fix: typo in css rdp-hidden module css

ayuhito:
## What's Changed

Follow up to #2298. I use CSS modules, so my CSS preprocessor (PostCSS) refuses to build in this situation.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/style.module.css
+++ b/src/style.module.css
@@ -262,7 +262,7 @@
 
 .hidden {
   visibility: hidden;
-  color: var(end-range_start-color);
+  color: var(--rdp-range_start-color);
 }
 
 .range_start {


gpbl:
Something did not work in the build here. Thanks.

ayuhito:
Can you clarify what that something is?

gpbl:
> Can you clarify what that something is?

Sorry, just a rushed note to myself :) The issue here is that when we change `style.css`, the sibling `style.module.css` is not updated during the build step.

Thanks for the PR will be published as soon as possible.

PR #2343, opened 2024-08-04
fix: update calendar state when `startMonth` or `endMonth` change

gpbl:


gpbl:
Please note this has been reverted in v9.0.9.

PR #2363, opened 2024-08-12
Export ChevronProps

rishabh-ink:
## What's Changed

Exporting the type of `Chevron`’s `props` as `ChevronProps`. All other custom components do export their types. It was missing on `Chevron`.

## Type of Change

- [x] Bug fix

### DIFF
--- a/src/components/Chevron.tsx
+++ b/src/components/Chevron.tsx
@@ -38,3 +38,5 @@ export function Chevron(props: {
     </svg>
   );
 }
+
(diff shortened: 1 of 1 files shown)


rishabh-ink:
@gpbl In case you missed the notification, friendly reminder to take a look. Thank you!

gpbl (approved):

gpbl:
Thanks @rishabh-ink for the heads up and the change (I am actually having a break after the release so I could be a bit slower here. 🧘🏽 )

Issue #2393, opened 2024-08-23
MonthsDropdown and YearsDropdown classes should be pass to the related Dropdowns


hsnaydd:
Classes `UI.MonthsDropdown` and `UI.YearsDropdown` do not pass to the related components. We do not have any data to differentiate between the two dropdowns.

PR #2394, opened 2024-08-23
fix: pass month and years dropdown classes to the related components

hsnaydd:
Fixes: #2393

## What's Changed

This pull request passes the `MonthsDropdown` and `YearsDropdown` classes to the related components. With these classes, users will be able to differentiate between these two dropdowns.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update


gpbl:
I really missed it. Thanks!

PR #2437, opened 2024-09-08
feat: add PreviousMonthButton and NextMonthButton to custom components

gpbl:


GAUKjpd:
@gpbl Will this be released with 9.1.0 ?

The docs says it's available now https://daypicker.dev/api/type-aliases/CustomComponents, but it's currently not in 9.0.9

PR #2441, opened 2024-09-09
fix(types): add missing ClassNames types for dropdowns

hsnaydd:
## What's Changed

the `classNames` prop in `DropdownProps` does not contain `UI.MonthsDropdown` and `UI.YearsDropdown`.   These classes are necessary because the user may need them. This pr allows the user to use them.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/components/Dropdown.tsx
+++ b/src/components/Dropdown.tsx
@@ -27,7 +27,12 @@ export function Dropdown(
     components: Pick<CustomComponents, "Select" | "Option" | "Chevron">;
     classNames: Pick<
       ClassNames,
-      UI.DropdownRoot | UI.Dropdown | UI.CaptionLabel | UI.Chevron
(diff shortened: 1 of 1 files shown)


gpbl:
Thanks! Wondering how we could write a test for this... so it doesn't happen again 🤔

gpbl:
> In my opinion, there is no benefit in restricting users from accessing all the classes here. All the classes are already being passed, so what is the risk in allowing the use of all classes? Users can already access these classes through the useDatePicker hook. I believe we should set the classNames type to ClassNames instead of selecting only a few. This way, you won’t need to worry about testing this. 🙂

Yes you are absolutely right, that was an overzealous typing from my side.

PR #2449, opened 2024-09-12
fix: use correct month to disable Chevron of NextMonthButton

qgadrian:
## What's Changed

Inside the `NextMonthButton` component, the `disabled` prop is now using the correct _next month_ reference for the `Chevron` component.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update


gpbl:
Thanks!

PR #2523, opened 2024-10-08
Explicitly provide resolved locale for all functions which need it

### LINKED ISSUE #2509: Inconsistent weekday headers and grid dates when using date-fns with default locale
## Code 

```tsx
import { enAU } from 'date-fns/locale';
import * as dateFns from 'date-fns';

dateFns.setDefaultOptions({ locale: enAU });

export function App() {
  return <DayPicker dateLib={dateFns} mode='single' />;
}
```

## Expected Behavior

I would expect the weekday column headers to correctly match up with the dates in the grid as pictured below:

> Note: Friday 4th October 2024 is a correct date and appears in the two grids

locale explicitly set to `enAU` (`locale={enAU}`):
CleanShot 2024-10-04 at 14 48 48

locale explicitly set to `enUS` (`locale={enUS}`):
CleanShot 2024-10-04 at 14 50 18

## Actual Behavior

When `dateFns` is supplied as the `dateLib` and has been setup with the default locale of `enAU`, the weekday headers render as `enUS`, but the grid renders as `enAU`, resulting in incorrect weekdays lining up with dates:

> Note: Friday 5th October 2024 is **not** a correct date

CleanShot 2024-10-04 at 14 47 35

daveallie:
## What's Changed

`locale` is optionally user defined as a prop to DayPicker and should fallback to en-US (the default for DayPicker) when none is supplied by the user. There were several places where this default was not respected, and instead `props.locale` was used (which can be undefined).

This PR addresses the above problem by explicitly requiring `locale` as a non-optional argument everywhere it's used, similar to how `dateLib` is passed around.

This resolves #2509 (related discussion #2511) in where if the client application used date-fns and had a locale set through `dateFns.setDefaultOptions`, then the internal calls to date-fns within DayPicker would use this client defined default (instead of the en-US default) in the cases where `props.locale` was being used (namely `useCalendar`).

Once we cross the threshold into DayPicker, we should always be using a resolved locale object to ensure consistency of rendered dates across components.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Additional Notes

Fixes #2509
Related #2511

### DIFF
--- a/src/DayPicker.tsx
+++ b/src/DayPicker.tsx
@@ -103,7 +103,7 @@ export function DayPicker(props: DayPickerProps) {
     formatYearDropdown
   } = formatters;
 
-  const calendar = useCalendar(props, dateLib);
+  const calendar = useCalendar(props, dateLib, locale);
 
   const {
     days,
@@ -128,7 +128,8 @@ export function DayPicker(props: DayPickerProps) {
     calendar,
     getModifiers,
     isSelected ?? (() => false),
-    dateLib
+    dateLib,
+    locale
   );
 
   const {
--- a/src/helpers/getDates.test.ts
+++ b/src/helpers/getDates.test.ts
@@ -1,5 +1,4 @@
-import { dateLib } from "../lib";
-
+import { dateLib, defaultLocale as locale } from "..";
 import { getDates } from "./getDates";
 
 describe("when the first month and the last month are the same", () => {
@@ -13,7 +12,8 @@ describe("when the first month and the last month are the same", () => {
           {
             fixedWeeks: false
           },
-          dateLib
+          dateLib,
+          locale
         );
         expect(dates).toHaveLength(42);
         expect(dates[0]).toEqual(new Date(2023, 10, 26));
@@ -28,7 +28,8 @@ describe("when the first month and the last month are the same", () => {
           {
             fixedWeeks: true
           },
-          dateLib
+          dateLib,
+          locale
         );
         expect(dates).toHaveLength(42);
         expect(dates[0]).toEqual(new Date(2023, 10, 26));
@@ -46,7 +47,8 @@ describe("when the first month and the last month are the same", () => {
           {
             fixedWeeks: false
           },
-          dateLib
+          dateLib,
+          locale
         );
         expect(dates).toHaveLength(35);
         expect(dates[0]).toEqual(new Date(2023, 3, 30));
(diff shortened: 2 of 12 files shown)


gpbl:
@daveallie, thanks for your PR! I see your point.

I need to investigate this issue further. I assumed date-fns' `setDefaultOptions` worked per package and it surprised me it "leaked" into DayPicker. 

I planned to use `setDefaultOptions` in DayPicker too, but if it changes the behavior of any date-fns used in the application, it might not be the right solution. We may need to add locale as you did here.

daveallie:
yep, unfortunately it's not like Axios where you can create an instance and then configure it with all your relevant defaults, `setDefaultOptions` applies to the whole library (although I suspect if you have multiple versions of date-fns available to your application, you'd get some interesting behaviour depending on where you called it from). 

the options as I see it are:
1. Be diligent with ensuring `locale` is defined everywhere you call a date-fns function
2. Re-export each date-fns method from your own module with a required `locale` prop to ensure usage
3. Build a module / class which takes `locale` at initialisation and sets up functions using [date-fns' FP currying](https://date-fns.org/v4.1.0/docs/FP-Guide#usage) (or just have that module/class internally call the date-fns methods with the locale)

imo 3 makes the most sense as dateLib is already being passed around. I would consider defining a specific class to wrap all date-fns calls along with the locale and replace all the dateLib passing with an instance of this new class.

e.g.

```ts
class DateUtils {
  constructor(readonly locale: Locale) {}

  format(date: Date, formatStr: string, options: FormatOptions = {}) {
    return format(date, formatStr, { ...options, locale: this.locale });
  }

  // ...
}

const dateLib = new DateUtils(locale);
dateLib.format(new Date(), "PPPP");
```

gpbl:
I really like the idea of a `DateLib` class initialized with options, instead of passing around all these `locale` - that has been proved as a bad idea. Not sure if we should merge this and undo all your work here @daveallie.

daveallie:
I don't mind at all if the work so far gets undone, the fix isn't urgent as the issue requires a few different stars to align (apparently just being Australian is enough 😅). I should have some time this weekend to implement it if you'd like to leave it in my hands, otherwise if you'd like to take this over, then you're more than welcome to

daveallie:
Diving into this, it's a fair bit more involved than I initially suspected, opening a new PR in case we want to merge this instead.

gpbl:
Superseded by https://github.com/gpbl/react-day-picker/pull/2550

PR #2535, opened 2024-10-11
build: update exports in package.json for better compatibility

### LINKED ISSUE #2533: build: Verify exports in package.json
We are receiving some reports (e.g. https://github.com/gpbl/react-day-picker/discussions/2521), about exports not being correctly interpreted by some bundler:

https://arethetypeswrong.github.io/?p=react-day-picker%409.1.3

### Problems

- [ ] [Resolution failed](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/NoResolution.md)
Import failed to resolve to type declarations or JavaScript files.
- [ ] [ESM (dynamic import only)](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/CJSResolvesToESM.md)
A require call resolved to an ESM JavaScript file, which is an error in Node and some bundlers. CommonJS consumers will need to use a dynamic import.
- [ ] [Used fallback condition](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FallbackCondition.md)
Import resolved to types through a conditional package.json export, but only after failing to resolve through an earlier condition. This behavior is a [TypeScript bug](https://github.com/microsoft/TypeScript/issues/50762). It may misrepresent the runtime behavior of this import and should not be relied upon.
- [ ] [Internal resolution error](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/InternalResolutionError.md)
Import found in a type declaration file failed to resolve. Either this indicates that runtime resolution errors will occur, or (more likely) the types misrepresent the contents of the JavaScript files.
Expand the raw details below to see the imports that failed to resolve.

gpbl:
Fix #2533.


Adnan-Sait:
@gpbl, when will this change be released?
Our Jest tests are failing since we upgraded from version 7 to 9.

```
SyntaxError: Cannot use import statement outside a module

import { DayPicker, DayPickerProps } from 'react-day-picker';
```

gpbl:
@Adnan-Sait, this has been published in v9.1.4. Could you check if it resolves your issue? Thanks!

Goodoose:
> @Adnan-Sait, this has been published in v9.1.4. Could you check if it resolves your issue? Thanks!

It still doesn't work as well

PR #2537, opened 2024-10-11
perf: improve performance when selecting long ranges

### LINKED ISSUE #2531: Performance issue when selecting a long date range
When using the `react-day-picker` component to select a date range, if the selected date range is too long, the UI takes too long to respond after selecting a new date.

How long the date range needs to be to cause this issue depends on the computing power of the device rendering the UI.

In the current project I work on, we use the `react-day-picker` together with a date input where users can type any date, and we were able to easily test a long date range and found the performance issue.
The screen recording below shows the response time of the UI:

https://github.com/user-attachments/assets/0660659b-5abd-493b-b283-bca3b6674901

The date range tested above can be rare to be ever used in a real application, but I tested a much shorter date range with CPU slowdown of 6x, the response delay is already noticeable as showed in the screen recording below:

https://github.com/user-attachments/assets/85124535-4d23-4cc3-84b1-b4a5b52310b7

## Code 

```tsx
import { useState } from 'react';

export function Start() {
  const [range, setRange] = useState({ from: new Date('0001-01-01'), to: new Date() });

  return (
    <DayPicker mode="range" selected={range} onSelect={setRange}  />
  );
}
```

## Expected Behavior

The [UI should respond in less than 100 ms](https://web.dev/articles/rail) when selecting a new date no matter how long the date range is.

## Actual Behavior

The UI takes a noticeable amount of time to respond after selecting a new date if the selected date range is too long.

## Screenshots

Besides the screen recordings above, I ran the Chrome profiler to debug the issue, the screenshots below show the issue in the profiler:

Flame chart zoom out

By zooming in we can see the following:

Flame chart zoom in

## Notes

By investigating the source code with the information from the profiler, the cause of the issue seems to be this [logic in the useRange hook](https://github.com/gpbl/react-day-picker/blob/e190c1123becef90a3ca24577c5d55b5e29be038/src/selection/useRange.tsx#L48).

It's iterating per day within the selected date range, which causes the issue when the date range is too long.

I thought about how to improve this logic and fix the performance issue, I can gladly open a PR with the necessary changes, please let me know if I can go ahead and try to fix it.
(I wasn't sure if I should open the PR right away or report the issue first, so I decided to report the issue first)

rodgobbi:
## What's Changed

Update the `select` function in the `src/selection/useRange.tsx` hook to improve the performance.

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Additional Notes

Fixes #2531


rodgobbi:
Hi @gpbl

I created a draft PR for now because I'd like to discuss with you what I found and align with you on how to implement the rest.

There's a straightforward improvement we can do with these [changes](https://github.com/gpbl/react-day-picker/pull/2537/commits/3c3f05b60e3938db77a0f9a3d1c3c32084c5e574).
I hoisted the check for `excludeDisabled && disabled` to the upper `if`, which prevents the `while` loop from running for all use cases where `excludeDisabled` is not enabled or there's no `disabled` dates.

But we still have the performance issue when `excludeDisabled && disabled` is `true`.
I checked the type of the `disabled` prop, which is the `Matcher` type, and it has many possible formats, which makes it a bit difficult to fix the performance issue. 😅 
I expect though that it is possible to fix the issue for all possible `Matcher` formats besides the callback one which is `(date: Date) => boolean`.
I think we can handle most of the `Matcher` formats by having a unique check for each format that is the most performant for that specific format.
E.g.: if `disabled` contains a single `Date`, we could just check if that `Date` is within the selected `DateRange` using the ideal function from `date-fns` for this use case.
But if the `Matcher` is a `(date: Date) => boolean` function, I think we cannot do anything besides iterating every date within the selected `DateRange`, as we need to call the `Matcher` with all the iterated dates.

Do you think it's worth adding this complexity to solve this issue?

If yes, I can work on it and show it to you once it's ready.
I could also do it in a separate PR, as these [changes](https://github.com/gpbl/react-day-picker/pull/2537/commits/3c3f05b60e3938db77a0f9a3d1c3c32084c5e574) I mentioned could be merged separately.

gpbl:
Nice work, @rodgobbi! I believe the change is indeed a step forward. However, [this line](https://github.com/rodgobbi/react-day-picker/blob/issue2531/src/selection/useRange.tsx#L48) is where most of the impact occurs. Specifically, it's looping from `newRange.from` to the existing `newRange.to`, which could be quite a big range.

I suspect that using `months[months.length-1].month` (`months` is coming from `useCalendar`) instead of `newRange.to`, might be enough to reduce the loop's size.

How could we test this behavior in Jest? It would simplify debugging.

gpbl:
PS A suggestion from ChatGPT  to improve the loop:

> You can improve the performance by moving the expensive differenceInCalendarDays calculation outside of the loop, checking once before entering the loop. Additionally, reducing the complexity of dateMatchModifiers if possible can further improve performance. Here’s an optimized version:

```tsx
if (newRange?.from && newRange.to) {
  let newDate = newRange.from;
  const totalDays = dateLib.differenceInCalendarDays(newRange.to, newDate);

  for (let i = 0; i < totalDays; i++) {
    newDate = dateLib.addDays(newDate, 1);
    
    if (excludeDisabled && disabled && dateMatchModifiers(newDate, disabled, dateLib)) {
      newRange.from = triggerDate;
      newRange.to = undefined;
      break;
    }
  }
}
```

rodgobbi:
@gpbl 
Thanks for the suggestion, I missed hoisting the `differenceInCalendarDays` call, which is indeed better.

Just one important detail, we cannot rely only on the visible `months` returned by the `useCalendar` hook for checking if there is any disabled date within the selected `DateRange`, because [the selected DateRange can be longer than the visible months in the calendar](https://github.com/gpbl/react-day-picker/issues/2531#issuecomment-2406190134).

I'll work on what I have in mind during the upcoming week and if it works, we can review it when it's ready.

About
> How could we test this behavior in Jest? The number of calls of... what? It would simplify debugging.

IMO testing how many times something was called doesn't provide that much value, as different devices take different amounts of time to run the same amount of iterations.
It's also difficult to assert performance, even more difficult using only Jest.
For such a test I imagine we could assert how long it took for the component to rerender.
The problem is that every different machine running this test will have a different result, which can cause the test to be very flaky.
It seems like it's a case where the cost of testing is not worth it, but let me know if you'd like to try implementing a test for it.

gpbl:
> Just one important detail, we cannot rely only on the visible `months` returned by the `useCalendar` hook for checking if there is any disabled date within the selected `DateRange`, because [the selected DateRange can be longer than the visible months in the calendar](https://github.com/gpbl/react-day-picker/issues/2531#issuecomment-2406190134).

This is true! The reason we do that is to reset the range to start from the clicked date. 

🤔 We do have a more performant way to know if a day is disabled, via [`useDayPicker().getModifiers()`](https://daypicker.dev/api/functions/useDayPicker). This function returns the modifiers for a given day and should be more performant than `dateMatchModifiers`.

> The problem is that every different machine running this test will have a different result, which can cause the test to be very flaky.

This could be my chance to finally add the long-awaited performance tests to DayPicker 👯

rodgobbi:
@gpbl 
Great work!
I see that you already set up the performance tests, that was quick 👏 

Two points about it:
1 - Currently the `performance` CI check is failing in this PR, and it doesn't look like it's related to the `lighthouse` test.

2 - In theory the `RangeLong.tsx` example should not be passing the `performance` test, was it passing when you added the `performance` CI check for it?
If yes, we will need to use a much longer date range in the test case or try to limit the performance of the process running the test.
The second option looks better if feasible, as it would make the CI check less flaky.

gpbl:
@rodgobbi Cool! I was excited to add the performance tests, and I played around with Lighthouse over the weekend.

I’m learning how to set up Lighthouse with GitHub correctly—sorry for the confusion with the PR. I’ve just updated it to the main branch.

I still need to write the documentation for this new `performance-tests` package. It should also work via the CLI and speed up the testing a little bit:
```
pnpm install
pnpm --filter performance-tests run capture
```

For now, I’d use these performance tests as a report here on GitHub. The job runs when a PR is labeled with `performance`, and the report can be downloaded as an artifact. Ideally, I’d like to see the report as a comment in the PR.

Screenshot 2024-10-14 at 8 43 26 AM

rodgobbi:
@gpbl I finished the first iteration of what I had in mind, I added the [`src/utils/rangeMatchModifiers.ts`](https://github.com/gpbl/react-day-picker/pull/2537/files#diff-6d788dbe590eabcb6b3700e4c54f5651428d296c95a7de41016646771cfb4f97) utils function.

I tested the "Interaction to next paint" with the [`examples/RangeLongExcludeDisabled.tsx`](https://github.com/gpbl/react-day-picker/pull/2537/files#diff-4ba97b740786a6f67d6c60902bc9009ec27df5f67b79e93bdba518baa04889e2) component with 20x CPU slowdown and this is the difference:

| Current implementation  | `rangeMatchModifiers` implementation |
| ------------- | ------------- |
| Current version - day picker |  Optmized version - day picker   |

Important note, ~ 300 ms of response time is the default expected time for the component in general when using 20x slowdown (on my machine):
Range test - day picker

@gpbl Let me know what you think and if you want to integrate those changes, I pushed the first iteration I worked on to showcase it, so I still need to test better, polish the code, and implement automated tests.

gpbl (commented):
Well done, @rodgobbi! I can see the performance improvements. Give me some time to review your changes. Thanks!

src/utils/rangeMatchModifiers.ts
```diff
@@ -0,0 +1,123 @@
...
+export function rangeMatchModifiers(
+  range: DateRange,
+  matchers: Matcher | Matcher[],
+  dateLib: DateLib = defaultDateLib
+): boolean {
+  // TODO maybe add a console.warning if this function took too long to finish
+  const matchersArr = !Array.isArray(matchers) ? [matchers] : matchers;
+
+  // function matchers needs to verified against every day in the date range,
+  // because of that it's the least performant one and should be deferred to be the last evaluated
```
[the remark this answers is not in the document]
rodgobbi: If I understood the suggestion correctly, I think this is the case with the current implementation, and the `dateMatchModifiers` utils function is already the generic function that validates any specific `Date` against a list of `Matcher`s, regardless of the `Date` coming from a date `Range`.
This is easier to understand, has less complexity, and works, but hits a performance bottleneck that the `rangeMatchModifiers` solves.
Having a logic that specifically handles the `Range` data structure allows us to optimize for that.

The only issue is that I don't see any other possible approach when a `Matcher` is of type `(date: Date) => boolean`, we need to iterate all dates within the date `Range` passing the dates to the `Matcher` function.

gpbl:
OK, I had a chance to dig deeper into your PR, @rodgobbi! 

The performance improvement is noticeable, and it's definitely a step in the right direction. However, the part about matching ranges instead of single days is still a bit unclear to me, and I'm not sure I fully understand what the bottleneck was there.

I'll try adopting some of your changes in a different branch to see how it goes.

rodgobbi:
@gpbl  thanks for following up!
Great, lemme know what you think after using this solution more.
I also found another issue while testing the `react-day-picker`, I'll open another issue later to discuss, which can be tackled in parallel.

> However, the part about matching ranges instead of single days is still a bit unclear to me, and I'm not sure I fully understand what the bottleneck was there.

Could you elaborate more about your concerns and doubts so I can explain better the solution?

If I understood correctly your concern, the difference and also the benefit of creating a custom function for matching `Matcher`s against a date `Range` is that for most types of `Matcher`s we can have a specialized logic that requires way less computation than iterating all days in the date `Range` and calling `dateMatchModifiers` with every iterated day date.

E.g.: if the day picker has the date 31/12/1999 disabled, and we try to select the date range from 01/01/0001 to 01/01/2000, the previous implementation will call `dateMatchModifiers` approximately 729.635 times (1999 years * 365 days) until the logic finds that one of the days in the date `Range` is disabled so the selected date `Range` is invalid.
With the `rangeMatchModifiers` implementation, we call `rangeIncludesDate` only once for the same use case, and `rangeIncludesDate` internally calls `differenceInCalendarDays` and `isSameDay` a fixed amount of times no matter the duration of the date `Range` thus the computational complexity is fixed no matter how long the date `Range` passed to `rangeMatchModifiers` is.

The example above helps illustrate that the computation grows with the duration of the date `Range` if we only use single days for matching against `Matcher`s, which can eventually hit a bottleneck taking too long to iterate all days in a date `Range`. 
However, I could not find a better solution when the `Matcher` is a function, because the `Matcher` needs to be tested against all single days within the date `Range`.

src/selection/useRange.tsx
```diff
@@ -43,20 +43,11 @@ export function useRange<T extends DayPickerProps>(
...
-          disabled &&
-          dateMatchModifiers(newDate, disabled, dateLib)
-        ) {
-          // if a disabled days is found, the range is reset
-          newRange.from = triggerDate;
-          newRange.to = undefined;
-          break;
-        }
+    if (excludeDisabled && disabled && newRange?.from && newRange.to) {
+      if (rangeMatchModifiers(newRange, disabled, dateLib)) {
```
gpbl, line 47: `rangeMatchModifiers` also checks `newRange.from` and `newRange.to`. We could force the function to work only with `Required<DateRange>`, and remove the same checks in `rangeMatchModifiers`.

src/utils/rangeMatchModifiers.ts
```diff
@@ -0,0 +1,123 @@
...
+ * const matcher2: DateRange = {
+ *  from: new Date(2022, 5, 1),
+ *  to: new Date(2022, 5, 23)
+ * }
+ * const dateMatchModifiers(date, [matcher1, matcher2]); // true, since matcher1 is in the date.
+ * ```
+ *
+ * @group Utilities
+ */
+export function rangeMatchModifiers(
```
gpbl, line 34: I think the function should be renamed;

```suggestion
export function rangeContainsModifiers(
```

src/utils/rangeMatchModifiers.ts
```diff
@@ -0,0 +1,123 @@
...
+  range: DateRange,
```
gpbl, line 35: I think this should help to remove the `range.from`, `range.to` checks inside this function:

```suggestion
  range: Required<DateRange>,
```

If `<Required>` utils does not work as expected, let be even more explicit:

```suggestion
  range: { from: Date, to: Date },
```

src/utils/rangeMatchModifiers.ts
```diff
@@ -0,0 +1,123 @@
...
+
+      // iterate at maximum one week or the total days if the range is shorter than one week
+      const totalDaysLimit = Math.min(totalDays, 6);
+      for (let i = 0; i <= totalDaysLimit; i++) {
+        if (dayOfWeekArr.includes(date.getDay())) {
+          return true;
+        }
+        date = dateLib.addDays(date, 1);
+      }
+      return false;
```
gpbl, line 88: This code, to me, should be moved into a new `rangeContainsDayOfWeeks` utility, so that we can unit test it separately:

```suggestion
      const dayOfWeekArr = !Array.isArray(matcher.dayOfWeek)
        ? [matcher.dayOfWeek]
        : matcher.dayOfWeek;
      return rangeContainsDayOfWeeks(dayOfWeekArr);
```

src/utils/rangeMatchModifiers.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (
+      isDateInterval(matcher) ||
+      isDateAfterType(matcher) ||
+      isDateBeforeType(matcher)
+    ) {
+      return (
+        dateMatchModifiers(range.from, matcher, dateLib) ||
+        dateMatchModifiers(range.to, matcher, dateLib)
+      );
+    }
```
gpbl, line 99: The above is difficult to follow to me :) Could we split it in single `if`s and add some comments about:

```suggestion
    if (isDateInterval(matcher)) {
      return (
        dateMatchModifiers(range.from, matcher, dateLib) ||
        dateMatchModifiers(range.to, matcher, dateLib)
      );
    }
    if (isDateAfterType(matcher)) {
      return (
        dateMatchModifiers(range.from, matcher, dateLib) ||
        dateMatchModifiers(range.to, matcher, dateLib)
      );
    }
    if (isDateBeforeType(matcher)) {
      return (
        dateMatchModifiers(range.from, matcher, dateLib) ||
        dateMatchModifiers(range.to, matcher, dateLib)
      );
    }
```

src/utils/rangeMatchModifiers.ts
```diff
@@ -0,0 +1,123 @@
...
+  });
+
+  if (nonFunctionMatchersResult) {
+    return true;
+  }
+
+  const functionMatchers = matchersArr.filter(
+    (matcher) => typeof matcher === "function"
+  );
+  if (functionMatchers.length && range.to && range.from) {
```
gpbl, line 110: According to the proposed type in `range` argument, we already know `range.to` and `range.from` are defined:

```suggestion
  if (functionMatchers.length) {
```

gpbl (changes requested):
@rodgobbi Ok, I've added more comments! Thanks for your patience. 

I think we are on the right track (really liking this approach), but we need to refine the code and add some unit tests.

Let me know if my comments make sense. If you don't have time, I can take it from here.

I've already improved range performance in v9.2.1 via https://github.com/gpbl/react-day-picker/pull/2560, so we can focus more effectively on this.

rodgobbi:
@gpbl 
No problem!
It's good to be thorough and I appreciate the care you put into maintaining the lib.

Thanks for offering the help, I'll try to handle everything now that we are aligned, I'll work on it this week and let you know once it's ready for review.
FYI, I'll leave some replies to your comments.

Also, one last thing that I need feedback: I left this [code comment](https://github.com/gpbl/react-day-picker/pull/2537/files#diff-6d788dbe590eabcb6b3700e4c54f5651428d296c95a7de41016646771cfb4f97R39) for discussion.
Because the performance bottleneck can still happen for `Matcher` functions, do you think it would be useful to leave a console warning when we detect that [this while loop](https://github.com/gpbl/react-day-picker/pull/2537/files#diff-6d788dbe590eabcb6b3700e4c54f5651428d296c95a7de41016646771cfb4f97R110) took too long to complete?
This would help in debugging if someone ever faces this issue, we could instruct them to try to avoid `Matcher` functions and use other `Matcher` types if possible.
Detecting it would be easy, we could measure if, for example, the while loop took more than 100ms to complete.

src/utils/rangeMatchModifiers.ts
rodgobbi: FYI this function follows the same pattern as `dateMatchModifiers`, it felt more appropriate to follow the same naming because it's the same idea but applied for date ranges.
I'll proceed with applying the suggestion in any case, if you change your mind lemme know.

src/utils/rangeMatchModifiers.ts
rodgobbi: Thanks for the info 👍 
I thought about doing the same but preferred to get feedback about what pattern to follow for the codebase.
Do you think we should extract this type to the general `types` file in the lib?
Name it something like `FullDateRange` or `CompleteDateRange`.
This could make the typing exported by the lib confusing, so this type could be used only internally.

gpbl: I'd prefer keep it explicit in the argument... but I can apply these style changes once merged - do not worry! Thanks for asking.

rodgobbi:
@gpbl
I got in the flow today after work and finished the changes, the PR is ready for review 👌 
I implemented many tests for everything I added, and thanks to them I was able to surface some bugs in the implementation of the `rangeContainsModifiers` function.

gpbl (approved):

gpbl:
Looking great, @rodgobbi! Thanks for your effort and patience in addressing my feedback.

Your contribution is the first significant one in a while. I’m happy to see this performance bottleneck so elegantly resolved. 🤟

rodgobbi:
@gpbl 
Thanks for the feedback and encouragement!
I'm happy to be able to contribute back 🙏 

As a follow-up, I found a minor bug with the day picker when using the `startMonth` and `endMonth` props.
Do you prefer that I open an issue first, reporting the bug, or is it better to open a PR directly?
(I'd explain the issue in the PR description)

gpbl:
> @gpbl Thanks for the feedback and encouragement! I'm happy to be able to contribute back 🙏
> 
> As a follow-up, I found a minor bug with the day picker when using the `startMonth` and `endMonth` props. Do you prefer that I open an issue first, reporting the bug, or is it better to open a PR directly? (I'd explain the issue in the PR description)

Sure, open the PR - do not worry about creating an issue if it is a quick fix. Thanks!

PR #2538, opened 2024-10-12
Replace dateLib object with DateLib class

### LINKED ISSUE #2509: Inconsistent weekday headers and grid dates when using date-fns with default locale

daveallie:
## What's Changed

- `dateLib` was an object containing a subset of date-fns functions, it is now a full class, which takes that subset and locale/formatting options as constructor arguments.
  - Although `DateLib` can take an undefined Locale, functions within `DateLib` which different return values based on locale, will require the DateLib class to have been initialised with a locale in order to be called.
    - This should allow for flexibility, while also enforcing the use of locale when it's needed.
  - Other formatting options like `weekStartsOn`, `firstWeekContainsDate`, `useAdditionalWeekYearTokens`, `useAdditionalDayOfYearTokens` are all passed to `DateLib` on initialization and explicitly not allowed to be passed in directly when calling those functions, instead we use the values that DateLib was initialized with
    - This effectively creates an instance of date-fns with a fixed set of formatting options
- I have tried to keep the APIs unchanged, across externally facing functions, using the optional `dateLib` as it was used before
  - `dateLib.options` exposes the formatting options it was initialized with, allowing for them to be passed down to user-provided label/format functions.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Additional Notes

Fixes #2509
Related #2523
Related #2511


gpbl (commented):
Great work here, @daveallie! I’ll need some time to review it. Thanks...

src/lib/dateLib.ts
```diff
@@ -173,3 +178,206 @@ export const dateLib = {
...
 };
+
+/**
+ * The date library used by DayPicker. It's a subset of the date-fns functions
+ * with an explicit locale / formatting options.
+ *
+ * @private
+ * @internal
+ */
+export class DateLib {
```
gpbl, line 189: Would this work somehow?

```suggestion
export class DateLib extends DefaultDateLib {
```

src/lib/dateLib.ts
```diff
@@ -173,3 +178,206 @@ export const dateLib = {
...
+   */
+  constructor(
+    readonly options: Options = {},
+    dateLibBaseOverrides: Partial<DateLibBase> = {}
+  ) {
+    this.dateLibBase = {
+      ...defaultDateLibBase,
+      ...dateLibBaseOverrides
+    };
+  }
```
gpbl, line 204: Is the `dateLibBase` property necessary? What about a `DateLib` class to extends a `DefaultDateLib`: Then, in the constructor:

```ts
if ('addWeeks' in dateLibBaseOverrides) this.addWeeks = dateLibBaseOverrides.addWeeks;
```

There might be reasons why we’re not doing this, but I can’t identify any just by looking at the code.

src/lib/dateLib.ts
```diff
@@ -173,3 +178,206 @@ export const dateLib = {
...
+  /** Formats the given date using the specified format string. */
+  format = (date: DateArg<Date> & {}, formatStr: string) => {
+    if (!this.locale) {
+      throw new Error(
+        "DateLib needs to be initialized with locale to use the `format` function."
+      );
+    }
+
+    return format(date, formatStr, this.options);
+  };
```
gpbl, line 348: The assumption here is that `this.options` includes `this.locale`, right? If so, why is `this.locale` needed separately? Couldn't we just:

```suggestion
    if (!this.options.locale) {
      throw new Error(
        "DateLib needs to be initialized with locale to use the `format` function."
      );
    }

    return format(date, formatStr, this.options);
  };
```

P.S. I love how well the code is documented. Thanks!

src/lib/dateLib.ts
```diff
@@ -173,3 +178,206 @@ export const dateLib = {
...
+
+  /** Returns the week number for the given date. */
+  getWeek = (date: DateArg<Date> & {}) => {
+    if (!this.locale) {
+      throw new Error(
+        "DateLib needs to be initialized with locale to use the `getWeek` function."
+      );
+    }
+
```
gpbl, line 357: Also, can’t we rely on a default `this.options.locale` to avoid these checks? 🤔

gpbl (commented):
I find your solution very interesting, and there's definitely something to learn from your changes. I'm not very up-to-date with OOP in JavaScript, so here are some initial thoughts on `DateLib`.

The thing is, we should also test it, and I'm trying to find a way to simplify the code.

src/types/props.ts
```diff
@@ -437,7 +437,7 @@ export interface PropsBase {
    * @since 9.0.0
    * @experimental
    */
-  dateLib?: Partial<DateLib> | undefined;
+  dateLib?: Partial<DateLibBase> | undefined;
```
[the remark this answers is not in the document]
daveallie: yeah fair, I can leave the name unchanged for exporting, and adjust the internal name from `DateLib` to `DateLibImpl` or something

src/lib/dateLib.ts
daveallie: Default locale is what caused the original bug, by making locale optional on deeper calls in the library, we're allowing callers to forget to provide the known locale. For this implementation, I took the approach of still allowing locale optionally, but ensuring it's provided to use any function which is dependent on locale. This should help catch issues where locale is forgotten about in DateLib construction, but required when rendering.

Ultimately DateLib construction only happens inside `react-day-picker`, so the only time this is hit will be if we've written a bug, it's a footgun prevention mechanism

src/lib/dateLib.ts
daveallie: yeah absolutely! I think what you're suggesting is more clear, I was just using `this.locale` as shorthand, but it's more indirection.

src/lib/dateLib.ts
```diff
@@ -173,3 +178,206 @@ export const dateLib = {
...
+      ...dateLibBaseOverrides
+    };
+  }
+
+  /**
+   * Creates a new instance of the date library with the default locale.
+   *
+   * @param options The formatting options for the date library.
+   */
+  static fromOptionsDefaultLocale(
```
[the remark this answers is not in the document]
daveallie: yep, let me take another look at my approach, will attempt to rework this part

src/lib/dateLib.ts
daveallie: currently I've done it this way, as for `endOfWeek`, `format`, `getWeek`, and `startOfWeek` (locale required functions), I still want access to the original function with the correct name, but the interface from the class to not include the locale and have it pass in the internally known locale to the original function. I'll play around with another way to achieve the same thing a little cleaner.

daveallie:
Thanks for the review @gpbl, I'm not an expert with OOP in JS either tbh, but it's pretty powerful if done right and this seems like a good use for it (wrapping up context and logic into a single instance). I had to wrap my head around the default args in the renderers which were a bit squirrelly, but this approach functions fine and provides a little more robust-ness for future expansion and prevention of issues in the future.

I'll throw up another revision in another few days addressing your feedback, if you spot anything else in there you'd like changed, please call it out. 🫡

gpbl:
@daveallie, I had the chance to review the `DateLib` class in https://github.com/gpbl/react-day-picker/pull/2550.

You can check out the code here: [dateLib.ts](https://github.com/gpbl/react-day-picker/blob/gpbl/datelib/src/lib/dateLib.ts).

I still need to add tests, particularly for the weekday names that you met before, but this version is closer to the concept I had in mind.

If you have some time, I’d appreciate a quick review!

gpbl:
PS. I tried configuring my PR to merge my changes into this one for an easier diff, but for some reason, I couldn't get it to work.

gpbl:
@daveallie thanks so much for your extensive work here! I've merged #2550 which includes [the test](https://github.com/gpbl/react-day-picker/blob/8e396186e27b64452660003be6011055881994c3/examples/Testcase2511.tsx) for #2511.

daveallie:
Thanks for picking this up and following it through @gpbl, appreciate it 🤝

PR #2547, opened 2024-10-15
build: improve css exports in package.json

karlshea:
# Pull Request Template

## What's Changed

_Briefly describe the changes in this pull request._

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Tips for a good PR

Thanks

## Additional Notes

Test PR updating stylesheet exports to fix SASS file resolution using sass-loader and webpack.


karlshea:
Is the build-website fail after you pushed up the change? 

It's crazy that it's impossible to find decent docs about how module resolution is supposed to work. sass-loader says it uses webpack's resolution and they have docs here: https://webpack.js.org/configuration/resolve/

According to those what you originally had should have worked. I'll open an issue with sass-loader and see if they have any input.

karlshea:
Discussion link here: https://github.com/webpack/webpack/discussions/18866

gpbl:
@karlshea yes I've patched your PR with some more changes. 

> Oh I see you fixed the import. So I guess it could go either way, the change I just pushed to add style/import/require or we could revert it back to just listing out default/types without the nested keys. Either seems to be working when I run my build.

I won't touch it anymore :D thanks for your help here hopefully will help your work.

BTW the TypeScript exports are now:

Screenshot 2024-10-16 at 11 26 40 AM

karlshea:
Thanks for taking a look! Also would like to say thanks for the package, I spent the last several days moving a big project from moment/react-datetime to date-fns/react-day-picker and it's been _very_ nice to work with!

gpbl:
(will publish the fix later in the day...)

karlshea:
The only thing I ran into is actually #2549 and then saw there was an open issue for it already, but it's not a huge blocker or anything. Everything else just worked how I would have hoped!

PR #2555, opened 2024-10-23
build(style): support importing styles in Sass

GeorgeTaveras1231:
# Pull Request Template

## What's Changed

Allows consumers to import base CSS using SASS. This gives consumers more flexibility to manipulate the base CSS. For example, forcing the styles under a CSS layer

```scss
@layer third-party-libs {
  @import 'pkg:react-day-picker/styles.module';
}
```

OR, forcing the styles to be scoped under a selector

```scss
.my-daypicker {
  @import 'pkg:react-day-picker/styles.module';
}
```

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Tips for a good PR

Thanks

## Additional Notes

Add any extra comments or questions here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,9 @@
       "style": {
         "default": "./src/style.css"
       },
+      "sass": {
+        "default": "./src/style.css"
+      },
       "import": {
         "types": "./src/style.module.css.d.ts",
         "default": "./src/style.css"
@@ -67,6 +70,9 @@
     },
     "./style.module.css": {
       "style": {
+        "default": "./src/style.module.css"
+      },
+      "sass": {
         "default": "./src/style.module.css"
       },
       "import": {
@@ -82,6 +88,9 @@
       "style": {
         "default": "./src/style.css"
       },
+      "sass": {
+        "default": "./src/style.css"
+      },
       "import": {
         "types": "./src/style.module.css.d.ts",
         "default": "./src/style.css"
@@ -93,6 +102,9 @@
     },
     "./src/style.module.css": {
       "style": {
+        "default": "./src/style.module.css"
+      },
+      "sass": {
         "default": "./src/style.module.css"
       },
       "import": {
@@ -108,6 +120,9 @@
       "style": {
         "default": "./src/style.css"
       },
+      "sass": {
+        "default": "./src/style.css"
+      },
       "import": {
         "types": "./src/style.module.css.d.ts",
         "default": "./src/style.css"
@@ -119,6 +134,9 @@
     },
     "./dist/style.module.css": {
       "style": {
+        "default": "./src/style.module.css"
+      },
+      "sass": {
         "default": "./src/style.module.css"
       },
       "import": {
--- a/package.json
+++ b/package.json
@@ -52,11 +52,11 @@
         "default": "./dist/cjs/locale.js"
       }
     },
+    "./style": {
+      "sass": "./src/style.css"
+    },
     "./style.css": {
       "style": {
-        "default": "./src/style.css"
-      },
-      "sass": {
         "default": "./src/style.css"
       },
       "import": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -52,6 +52,9 @@
         "default": "./dist/cjs/locale.js"
       }
     },
+    "./style": {
```
gpbl, line 55: This doesn't sound correct to me: does it mean only Sass can import from `react-day-picker/style`? Why not add a `sass` key to the existing `style.css`? 

Or did you mean:

```suggestion
    "./style.scss": {
```

package.json
```diff
@@ -65,6 +68,9 @@
         "default": "./src/style.css"
       }
     },
+    "./style.module": {
+      "sass": "./src/style.module.css"
```
gpbl, line 72: Question: when should Sass users import this CSS module?

gpbl (commented):
Thanks @GeorgeTaveras1231 for your PR!

I'm not a Sass expert, so I'm not sure why the current package.json isn't working with Sass. Could you provide more details?

GeorgeTaveras1231:
@gpbl Sorry if the PR lacked context. Here are some more details:

The issue with the current package.json exports is that Sass requires the extension to be omitted when importing plain CSS - this is necessary for it to differentiate between runtime and build-time imports. See their docs on [how `@import` works](https://sass-lang.com/documentation/at-rules/import/#plain-css-imports).

I will answer the other questions inline 😌

package.json
GeorgeTaveras1231: @gpbl 

> does it mean only Sass can import from react-day-picker/style? 

it does mean that at the moment, only Sass can import `react-day-picker/style`. Maybe its a bit odd, but you aren't limited from re-purposing this export in the future. 

> Why not add a sass key to the existing style.css?

This was actually how I first attempted to fix my issue, but like I shared in [my other comment](https://github.com/gpbl/react-day-picker/pull/2555#issuecomment-2437640335), Sass requires that the import URL not end with `.css` when it is meant to be a build-time import. 

> ::about your suggestion::

I'm not sure if that would work. I prefer the extensionless approach because it more closely resembles how a .css file would be treated when using Sass. Adding the `.scss` could work but could be misleading. 🤔

package.json
GeorgeTaveras1231: This is actually the export I'm using in my project, and the PR description has some examples.

The strongest example is probably the `@layer` one. 

In my project, I'm using [CSS Cascade Layers](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Cascade_layers) to manage the order of CSS rules written by my team, and dozens of other teams.

Since we use CSS modules and often re-use components, and override styles, we need some way to manage (other than with CSS specificity) the order in which CSS is applied to components. Because we can't predict which order the CSS will load in. 

We've created a dedicated `third-party` Layer which is where we put CSS that comes from third party libraries, which allows us to cleanly re-use third-party CSS, and override it without fighting with its selector's specificity.

I showed the example of how this is being imported in my project


In `MyDatepicker.module.scss`
```scss
@layer third-party {
  @import 'pkg:react-day-picker/styles.module';
}

/* Here we can include any overrides */
.day_button {
  /* Omitted */
}
```

I then import this module in my implementation of the `DayPicker` component.

```tsx
import styles from './MyDatepicker.module.scss';

const MyDatepicker = () => {
  return <DayPicker {/* omitted */} classNames={styles} />
}
```

gpbl:
@GeorgeTaveras1231 it makes sense to me! Thanks for your insight here (and for saving me from reading Sass manuals :)

gpbl (approved):

GeorgeTaveras1231:
Thanks @gpbl! I will be on the lookout for the next release 😌

PR #2570, opened 2024-11-02
chore: refactor and improve docs for range utilities

gpbl:
Some small updates after #2537

cc @rodgobbi


rodgobbi:
@gpbl  FYI, for the `areRangesOverlapping` function name I followed the pattern from [areIntervalsOverlapping](https://date-fns.org/v4.1.0/docs/areIntervalsOverlapping) from `date-fns`, as it's the same logic and I tried to reuse an established pattern.

Thanks for the clean-up 👍

gpbl:
> FYI, for the areRangesOverlapping function name I followed the pattern from [areIntervalsOverlapping](https://date-fns.org/v4.1.0/docs/areIntervalsOverlapping) from date-fns, as it's the same logic and I tried to reuse an established pattern.

That’s interesting! Thanks for the heads up. I’ll note this. In a future version, we might consider using the same terminology as date-fns.

PR #2578, opened 2024-11-10
fix: disable outside days outside the before and end month range

rodgobbi:
### Current behavior
When using both `showOutsideDays` prop equal to `true` together with a `startMonth` date, the days before the `startMonth` month are not disabled in the calendar and can be clicked and selected.

Screen recording of the issue using the `StartEndMonthsShowOutsideDays` example in this MR, the month March of 2024 is the only enabled, and the 10th day is disabled for showing how a disabled date behaves in the calendar:

https://github.com/user-attachments/assets/cab9d471-02b2-45b2-81f0-514d5d6bdb8f

### Expected behavior
Days before the `startMonth` month or after the `endMonth` month should be disabled in the calendar.

Screen recording of the fix using the `StartEndMonthsShowOutsideDays` example in this MR:

https://github.com/user-attachments/assets/9487942d-58e7-4ad3-8ded-373ac1d52397

## What's Changed

Add logic to flag the days before the `startMonth` month or after the `endMonth` month as disabled in the `src/useGetModifiers.tsx`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Additional Notes

@gpbl This PR starts as a draft because I need to discuss with you what should be the expected behavior and I'll suggest a behavior change based on what I discovered when fixing the issue.
I'll follow-up in the PR comments.

### DIFF
--- a/src/useGetModifiers.tsx
+++ b/src/useGetModifiers.tsx
@@ -17,9 +17,24 @@ export function useGetModifiers(
   props: DayPickerProps,
   dateLib: DateLib
 ) {
-  const { disabled, hidden, modifiers, showOutsideDays, today } = props;
+  const {
+    disabled,
+    hidden,
+    modifiers,
+    showOutsideDays,
+    today,
+    startMonth,
+    endMonth
+  } = props;
 
-  const { isSameDay, isSameMonth } = dateLib;
+  const {
+    isSameDay,
+    isSameMonth,
+    startOfMonth,
+    isBefore,
+    endOfMonth,
+    isAfter
+  } = dateLib;
 
   const internalModifiersMap: Record<DayFlag, CalendarDay[]> = {
     [DayFlag.focused]: [],
@@ -36,9 +51,18 @@ export function useGetModifiers(
 
     const isOutside = Boolean(displayMonth && !isSameMonth(date, displayMonth));
 
-    const isDisabled = Boolean(
-      disabled && dateMatchModifiers(date, disabled, dateLib)
+    const isBeforeStartMonth = Boolean(
+      startMonth && isBefore(date, startOfMonth(startMonth))
     );
+
+    const isAfterEndMonth = Boolean(
+      endMonth && isAfter(date, endOfMonth(endMonth))
+    );
+
+    const isDisabled =
+      Boolean(disabled && dateMatchModifiers(date, disabled, dateLib)) ||
+      isBeforeStartMonth ||
+      isAfterEndMonth;
 
     const isHidden =
       Boolean(hidden && dateMatchModifiers(date, hidden, dateLib)) ||
--- a/src/useGetModifiers.tsx
+++ b/src/useGetModifiers.tsx
@@ -59,13 +59,14 @@ export function useGetModifiers(
       endMonth && isAfter(date, endOfMonth(endMonth))
     );
 
-    const isDisabled =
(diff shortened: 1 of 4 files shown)


rodgobbi:
@gpbl 
While working to fix the issue, I discovered different behaviors for the days outside the `startMonth` and `endMonth` months.
More precisely, the days before the `startMonth` month are displayed in the calendar when `showOutsideDays` is equal to `true`, but the days after the `endMonth` month are not displayed in the calendar.

This behavior seems inconsistent, IMO it would feel more consistent if the day picker always showed the days or always hid the days, both the days before the `startMonth` month and the ones after the `endMonth`.
What do you think?
If it makes sense to have the "outside days" behavior the same, what would you prefer, to hide them or show them disabled?

src/useGetModifiers.tsx
```diff
@@ -41,10 +56,19 @@ export function useGetModifiers(
...
-      disabled && dateMatchModifiers(date, disabled, dateLib)
+    const isBeforeStartMonth = Boolean(
+      startMonth && isBefore(date, startOfMonth(startMonth))
+    );
+
+    const isAfterEndMonth = Boolean(
+      endMonth && isAfter(date, endOfMonth(endMonth))
     );
 
+    const isDisabled =
```
gpbl, line 67: Why disabling these days, instead of actually _hiding_ them ? (In the line below, we calculate the `isHidden` for that.)

gpbl (commented):
Hey, thanks for opening this PR. I missed that in our test cases. 

Question: why _disabling_ these days instead of _hiding_ them?

> More precisely, the days before the `startMonth` month are displayed in the calendar when `showOutsideDays` is equal to `true`, but the days after the `endMonth` month are not displayed in the calendar.


I believe this is because dates are listed from past to future, so the calendar just stops there, at the `endMonth`. But it is a good question; maybe there's a bug hidden there 👀

src/useGetModifiers.tsx
rodgobbi: For me it makes sense to hide, I left this implementation in the draft because I first started with it, and then understanding better the behavior of the calendar I discovered that we have different behaviors for days before the `startMonth` month and the ones after the `endMonth`, so I wanted to check with you first what would the preferred approach if we should commit to the same behavior for both sides of the `startMonth` and `endMonth` range.

With your feedback I have enough info to finish this PR 👍 
I'll update the logic and add some tests to mark the PR as ready.

gpbl:
@rodgobbi, I added the tests for `useGetModifiers` in https://github.com/gpbl/react-day-picker/pull/2586. Hopefully, I haven’t undone any of your work. Please sync this branch with `main`, before doing more changes. Thanks a lot for your help!

rodgobbi:
@gpbl 
The PR is ready for review.
The result after the changes:

https://github.com/user-attachments/assets/f4d9989a-776b-4588-9616-a8b302feccb9

I also updated the `src/useGetModifiers.test.tsx` test and made the assertions more restrictive because different props passed to the function can have different behaviors for certain modifiers, e.g. the change in this PR flags the outside days before the `startMonth` and after the `endMonth` as `hidden`, because of that it makes sense to assert every modifier for a given day just to be safe.

gpbl (approved):
Looking great! Love how you resolved the inconsistency in the `endMonth` part too. Thanks!

PR #2595, opened 2024-11-20
allow to use react-day-picker v7 with any 15+ react version

Knyazik01:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
   },
   "homepage": "https://react-day-picker.js.org",
   "peerDependencies": {
-    "react": "~0.13.x || ~0.14.x || ^15.0.0 || ^16.0.0 || ^17.0.0"
+    "react": "~0.13.x || ~0.14.x || >=15.0.0"
   },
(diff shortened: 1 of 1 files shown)


gpbl:
@Knyazik01, what about forking the v7 branch instead? I’m really short on time to deal or support with such an old version of this project.

Knyazik01:
@gpbl 
I've used v7.4.9 with react 18 (installed with legacy peer deps) and all works fine.

Migration from v7 to v8 has a lot braking changes.
So, I decided to request this change (update peer deps field) to allow use this version without legacy flag.

gpbl:
> @gpbl 
> I've used v7.4.9 with react 18 (installed with legacy peer deps) and all works fine.

Have you tried using `overrides` in package.json until you can upgrade to v9?

https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides

PR #2614, opened 2024-12-05
Fix: enable all years available in the year select dropdown

### LINKED ISSUE #2549: Using `startMonth` and `endMonth` can make dropdown unusable
When using `captionLayout="dropdown"`, months and years outside of the range are disabled. This prevents the user from using the dropdowns to select a valid month and year, if the currently selected month or year does not exist in the target month or year.

## Code 

```tsx

export function Start() {
  const from = new Date("2024-10-01");
  const to = new Date("2025-02-01");
  return (
    <DayPicker startMonth={from} endMonth={to} captionLayout="dropdown"  />
  );
}
```

## Expected Behavior

I should be able to select a date in the following year using only the dropdowns.

Ideally, if I select "2025" while in "November 2024", it should automatically change the month accordingly.

## Actual Behavior

It is impossible to select a different year if the current month is not available in that year.

## Screenshots

Screen Recording 2024-10-17 at 17 43 34

rodgobbi:
## What's Changed

Enable all years available in the year select dropdown. If after selecting the year, the selected month is before the `startMonth` or after the `endMonth`, the day picker selects the closest valid month.

Fix #2549 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Demo

https://github.com/user-attachments/assets/d335b0cc-a2f6-4cd5-aff5-6368d8c8b2d5

### DIFF
--- a/examples/Dropdown.test.tsx
+++ b/examples/Dropdown.test.tsx
@@ -23,43 +23,65 @@ test("should display the year dropdown", () => {
   expect(yearDropdown()).toBeInTheDocument();
 });
 
-test("should disable the months out of range", () => {
-  expect(
-    within(monthDropdown()).getByRole("option", { name: "January" })
-  ).toBeDisabled();
+test("should disable the months before startMonth", () => {
+  const disablesMonth = [
+    "January",
+    "February",
+    "March",
+    "April",
+    "May",
+    "June"
+  ];
+  for (const month of disablesMonth) {
+    expect(
+      within(monthDropdown()).getByRole("option", { name: month })
+    ).toBeDisabled();
+  }
+});
+
+test("should disable the months after endMonth", async () => {
+  await user.selectOptions(yearDropdown(), "2025");
+  const disablesMonth = ["November", "December"];
+  for (const month of disablesMonth) {
+    expect(
+      within(monthDropdown()).getByRole("option", { name: month })
+    ).toBeDisabled();
+  }
 });
 
 describe("when choosing a month", () => {
-  const monthName = "December";
-  beforeEach(async () => {
+  test("should display the month", async () => {
+    const monthName = "December";
     await user.selectOptions(monthDropdown(), monthName);
-  });
-  test("should display the month", () => {
     expect(grid()).toHaveAccessibleName(`${monthName} 2024`);
-  });
-  test("should disable the years out of range", () => {
-    expect(
-      within(yearDropdown()).getByRole("option", { name: "2025" })
-    ).toBeDisabled();
   });
 });
 
 describe("when choosing a year", () => {
-  const year = "2025";
-  beforeEach(async () => {
+  test("should display the year", async () => {
+    const year = "2025";
     await user.selectOptions(yearDropdown(), year);
-  });
-  test("should display the year", () => {
     expect(grid()).toHaveAccessibleName(`July ${year}`);
+  });
+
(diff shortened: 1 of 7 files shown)


rodgobbi:
@gpbl I found other potential improvements to the behavior of the day picker with the dropdown layout, which I'd like to share with you to discuss if they make sense and if they'd improve the day picker.
If you'd like to hear my suggestions, where do you prefer for me to post them to start a follow-up discussion?

gpbl (commented):
Hey @rodgobbi, thanks a lot for your PR.

I’ve been trying solution with the test case for https://github.com/gpbl/react-day-picker/issues/2549, but it doesn’t work as expected.

Here’s the code:

```tsx
import { DayPicker } from “react-day-picker”;

export function Dropdown() {
  const from = new Date(2024, 9);
  const to = new Date(2025, 1);
  return <DayPicker startMonth={from} endMonth={to} captionLayout=“dropdown” />;
}
```

This does render the list of disabled months, but January 2025 and February 2025 are not navigable via the dropdown menu.

https://github.com/user-attachments/assets/e7006448-934f-430e-a229-06dab66693a0

What I would try:

- Never disable the select options.
- When a select option changes, call `useCalendar().goToMonth`:

```tsx
calendar.goToMonth(
	new Date(yearDropdownValue, monthDropdownValue)
);
```

[goToMonth() function](https://github.com/gpbl/react-day-picker/blob/683370461495525ab184d0189666c0da258f7b39/src/useCalendar.ts#L136-L151) already navigates to the correct month if the selected option is not available. 

...if I haven’t overlooked anything, this solution could resolve the issue.

What do you think?

gpbl:
> where do you prefer for me to post them to start a follow-up discussion?

Feel free to open a new issue! Thanks for asking!

rodgobbi:
FYI @gpbl, this PR is already doing what you suggested 😅, I've added a screen recording to the PR to help visualize.   
`goToMonth` was already called by the change handler of the select dropdown and it handles trying to select an invalid date, which is great.  
I made all year options enabled by default, but left the month options disabled if they are invalid for the selected year, would you like to enable them also?   
IMO though, disabling or hiding the invalid months make sense if they are not allowed to be selected.

gpbl:
> FYI @gpbl, this PR is already doing what you suggested 😅, I've added a screen recording to the PR to help visualize.

Sorry, my bad! Something was off when checking out the PR. I do see it works now! Let me give the code changes a review. 👀 

---

I do still see an issue with the code below, where some months are not displayed. I think this is another bug as it happens also on the main branch.

```tsx
export function Dropdown() {
  const from = new Date(2024, 9);
  const to = new Date(2025, 1);
  return <DayPicker startMonth={from} endMonth={to} captionLayout="dropdown" />;
}
```

Screenshot 2024-12-07 at 5 00 54 PM


I can confirm that the issue reported in #2549, that we can replicate using [this code](https://github.com/gpbl/react-day-picker/pull/2614#pullrequestreview-2483169513) is still not fixed. There's something wrong in the calculations of the months:

gpbl (approved):

gpbl:
Thanks again for your work @rodgobbi !

rodgobbi:
Happy to help @gpbl 
Thanks for following up and merging, because I could not reproduce the issue you shared in your screenshot, the `Dropdown` example when I test in my machine works as expected.

PR #2620, opened 2024-12-10
refactor(jest): switch to @swc/jest for faster transforms

thevuong:
Remove ts-jest and configure @swc/jest for JavaScript/TypeScript transformations. This change improves the speed of transforming source files before running tests and supports modern React syntax with automatic runtime.


gpbl:
@thevuong woot thanks so much! Long time due!

PR #2658, opened 2025-01-07
feat: ethiopic calendar

gpbl:
This PR introduces support for the Ethiopic calendar in the DayPicker. 

The PR is still a work in progress. Switch to this branch to start applying changes to the source.

## How It Works 

The [Ethiopic calendar](https://en.wikipedia.org/wiki/Ethiopian_calendar) is rendered importing it from `react-day-picker/ethiopic`, as we do with the [Persian calendar](https://daypicker.dev/docs/localization#persian-calendar).

```tsx
import { DayPicker } from 'react-day-picker/ethiopic';

<DayPicker />
```

- [ ] Understand the correct name, Ethiopian or Ethiopic?

### Ethiopic Module

* [x] `src/ethiopic/index.tsx`: render DayPicker using the Ethiopic calendar, including the overridden locale and numeral system props.
* [ ] `src/ethiopic/lib`: module to contain the `date-fns` functions to override for date calculations and formatting (see below).

### Example Component, and Docs

* [x] `examples/Ethiopic.tsx`: the example components 
* [ ] `examples/Ethiopic.test.tsx`: test file for the new Ethiopic calendar example.
* [ ] `website/docs/docs/localization.mdx`: add Ethiopic Calendar paragraph.
* [ ] `website/components/playground`: add Ethiopic Calendar option.

> See [Working With the Source Code](https://daypicker.dev/development/contributing#working-with-the-source-code) to start the examples app and reach http://localhost:5173/?example=Ethiopic to see it running.

### Numeral System and Locale

* [ ] Understand which Locale should be used. Are they included in date-fns already?
* [ ] Understand which numeral systems we need to support.
* [ ] `src/types/props.ts`, `src/types/shared.ts` include the Ethiopic numeral system.
* [ ] Update `DateLib` to work with the new numeral system (e.g. `getDigitMap`)

### Ethiopic DateLib

Having the ethiopic date library compatible with the [DateLib](http://daypicker.dev/api/classes/DateLib) overrides protocol is the most challenging part of this project. The DateLib is initialized [here](https://github.com/gpbl/react-day-picker/blob/57662036f21ef4e054284b952e67ba4580ba5da2/src/ethiopic/index.tsx#L70) with the ethnic overrides, and should work out-of-the-box once we provide the correct .

In absence of a package like `date-fns-ethiopic`, we can initially include the ethiopic DateLib override's functions in the DayPicker package. Note that not all the functions may be require an ethiopic version.

- [ ] complete the missing functions (see list below) and their tests.

<details>
<summary>List of DateLib functions</summary>

- [ ] `addDays`
- [ ] `addMonths`
- [ ] `addWeeks`
- [ ] `addYears`
- [ ] `differenceInCalendarDays`
- [ ] `differenceInCalendarMonths`
- [ ] `eachMonthOfInterval`
- [ ] `endOfBroadcastWeek`
- [ ] `endOfISOWeek`
- [ ] `endOfMonth`
- [ ] `endOfWeek`
- [ ] `endOfYear`
- [ ] `format`
- [ ] `formatNumber`
- [ ] `getISOWeek`
- [ ] `getMonth`
- [ ] `getWeek`
- [ ] `getYear`
- [ ] `isAfter`
- [ ] `isBefore`
- [ ] `isSameDay`
- [ ] `isSameMonth`
- [ ] `isSameYear`
- [ ] `max`
- [ ] `min`
- [ ] `newDate`
- [ ] `setMonth`
- [ ] `setYear`
- [ ] `startOfBroadcastWeek`
- [ ] `startOfDay`
- [ ] `startOfISOWeek`
- [ ] `startOfMonth`
- [ ] `startOfWeek`
- [ ] `startOfYear`
- [ ] `today`

</details>

### Configuration and Build

* [x] `package.json`: includes the Ethiopic calendar module.
* [x] `ethiopic.js`: added module export for the Ethiopic calendar.
* [ ] Verify the type export with https://arethetypeswrong.github.io

### DIFF
--- a/ethiopic.js
+++ b/ethiopic.js
@@ -1,0 +1,4 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
+/* eslint-disable no-undef */
+const ethiopic = require("./dist/cjs/ethiopic/index.js");
+module.exports = ethiopic;
--- a/ethiopic.d.ts
+++ b/ethiopic.d.ts
@@ -1,0 +1,1 @@
+export * from "./dist/cjs/ethiopic/index.d.ts";
--- a/examples/Ethiopic.test.tsx
+++ b/examples/Ethiopic.test.tsx
@@ -1,0 +1,16 @@
+import React from "react";
+
+import { render } from "@/test/render";
+
+import { Ethiopic } from "./Ethiopic.jsx";
+
+const today = new Date(2021, 10, 25);
+
+beforeAll(() => jest.setSystemTime(today));
+afterAll(() => jest.useRealTimers());
+
+beforeEach(() => {
+  render(<Ethiopic />);
+});
+
+test.todo("should render the Ethiopic calendar");
--- a/examples/Ethiopic.tsx
+++ b/examples/Ethiopic.tsx
@@ -1,0 +1,7 @@
+import React from "react";
+
+import { DayPicker } from "react-day-picker/ethiopic";
+
+export function Ethiopic() {
+  return <DayPicker />;
+}
--- a/examples/index.ts
+++ b/examples/index.ts
@@ -18,6 +18,7 @@ export * from "./DisableNavigation";
 export * from "./Dropdown";
 export * from "./DropdownMonths";
 export * from "./DropdownMultipleMonths";
+export * from "./Ethiopic";
 export * from "./Fixedweeks";
 export * from "./FocusRecursive";
 export * from "./Footer";
--- a/package.json
+++ b/package.json
@@ -52,6 +52,16 @@
         "default": "./dist/cjs/persian.js"
       }
     },
+    "./ethiopic": {
+      "import": {
+        "types": "./dist/esm/ethiopic/index.d.ts",
+        "default": "./dist/esm/ethiopic/index.js"
+      },
+      "require": {
+        "types": "./dist/cjs/ethiopic/index.d.ts",
+        "default": "./dist/cjs/ethiopic/index.js"
+      }
+    },
     "./locale": {
       "import": {
         "types": "./dist/esm/locale.d.ts",
--- a/src/ethiopic/index.tsx
+++ b/src/ethiopic/index.tsx
@@ -1,0 +1,71 @@
+import React from "react";
+
+import type { Locale } from "date-fns";
+
+import {
+  DateLib,
+  DateLibOptions,
+  DayPicker as DayPickerComponent
+} from "../index.js";
+import type { DayPickerProps } from "../types/props.js";
+
+import * as ethiopicDateLib from "./lib/index.js";
+
+/**
+ * Render the Persian Calendar.
+ *
+ * @see https://daypicker.dev/docs/localization#persian-calendar
+ */
+export function DayPicker(
+  props: DayPickerProps & {
+    /**
+     * The locale to use in the calendar.
+     *
+     * @default `am-ET`
+     */
+    locale?: Locale;
+    /**
+     * The numeral system to use when formatting dates.
+     *
+     * - `latn`: Latin (Western Arabic)
+     * - `arab`: Arabic-Indic
+     * - `arabext`: Eastern Arabic-Indic (Persian)
+     * - `deva`: Devanagari
+     * - `ethio`: Ethiopic
+     * - `beng`: Bengali
+     * - `guru`: Gurmukhi
+     * - `gujr`: Gujarati
+     * - `orya`: Oriya
+     * - `tamldec`: Tamil
+     * - `telu`: Telugu
+     * - `knda`: Kannada
+     * - `mlym`: Malayalam
+     *
+     * @defaultValue `ethio` Eastern Arabic-Indic (Persian)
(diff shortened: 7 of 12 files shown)


ethiopic.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
+/* eslint-disable no-undef */
+const ethiopic = require("./dist/cjs/ethiopic/index.js");
```
rodgobbi, line 3: Related to the todo "Understand the correct name, Ethiopian or Ethiopic?".
From what I searched online:
> ... “Ethiopic” can be used more broadly to describe anything related to Ethiopia, though it is less common in this usage compared to “Ethiopian".

IMO even if "Ethiopic" is technically more correct, using "Ethiopian" seems better because it results in better discoverability, due to it being more used.
The discoverability is also better even if someone searches for "Ethiopia", because the search would still match with the "Ethiopian" word.

temesgen-mulugeta:
- Thinking about the name, I think "Ethiopic" is better. It pays homage to the Geez language, which is the basis of the calendar.

- When it comes to the numerals. There are specific geez numerals, so I can also work on that and add them to the utility functions, and we can add the support for it. 

-  I have implemented most of the date-fns functions that need to be overridden in my repo. But I haven't written tests for it, and I can also assist with that. From my experience, overriding the months and years functions is enough, with the exception of getWeek to get the week count right. 
https://github.com/temesgen-mulugeta/shadcn-ethiopian-date-picker/blob/main/src/lib/EthiopianDateLib.tsx

ethiopic.js
gpbl: Edit: @temesgen-mulugeta agrees with Ethiopic:

> Thinking about the name, I think "Ethiopic" is better. It pays homage to the Geez language, which is the basis of the calendar.

gpbl:
> When it comes to the numerals. There are specific geez numerals, so I can also work on that and add them to the utility functions, and we can add the support for it.

Interesting. Which numerals should we support? 

In my implementation, I could rely on `Intl` native object to switch to other numerals:

https://github.com/gpbl/react-day-picker/blob/1f070c3c7cefe474303e8c1d67ef532b07e080f0/src/classes/DateLib.ts#L129-L144

Are all the numerals used in the Ethiopic calendar included in the `Intl` object? In MacOS, they list two calendars: Ethiopic and Ethiopic (Alete Alem). Is Geʽez another one? 🤔

temesgen-mulugeta:
> Interesting. Which numerals should we support?
> 

The only numeral we should consider is Geez. The numerals look like '፩', '፪', '፫', '፬'. But I don't think we have support for it on Intl. But the numerals shouldn't be the primary focus. Since most of us use Western Arabic numerals.

ethiopic.js
rodgobbi: Thanks for the insights!
That is a good point to follow existing conventions, like Mac OS, and also ask other people who have more context.

gpbl:
I made some progress without knowing much about the Ethiopic calendar.

I scaffolded utilities and functions with the help of Copilot, so they are not guaranteed to work. The next step is to write tests for each of them and then fix the functions to pass the tests.

- [ethiopic/utils](https://github.com/gpbl/react-day-picker/tree/ethiopic-calendar/src/ethiopic/utils) to convert dates from/to Gregorian.
- [ethiopic/lib](https://github.com/gpbl/react-day-picker/tree/ethiopic-calendar/src/ethiopic/lib): date manipulation and formatting functions compatible with DayPicker's DateLib.

I wonder work @temesgen-mulugeta's work on MUI (or even the [ethiopian date](https://www.npmjs.com/package/ethiopian-date) package could be useful here.

Still missing the formatting function. I wonder if we could use the `Intl.DateTimeFormat` API to format Ethiopic dates?

temesgen-mulugeta:
I’ve been quite busy over the past week. I didn’t want to open a PR with an incomplete task, but I’ve now created one to assist you.
I’ve implemented utility functions for converting between Ethiopian and Gregorian dates. I’ve used these functions in over 10 projects, though, to be honest, I originally inherited them from my boss 😁—he’s been using them for over 12 years. However, they are only tested in real-life. I haven't written tests for them.
I’ve also implemented the necessary date manipulation functions in ethiopic/lib, but only the ones that were required. The PR is still incomplete since it needs tests, but it should provide a solid starting point.
Let me know your thoughts! 🚀

temesgen-mulugeta:
> Still missing the formatting function. I wonder if we could use the `Intl.DateTimeFormat` API to format Ethiopic dates?

As far as I know Intl doesn't provide date formatting for Ethiopian dates. So what I did was create a date formatting function for the formattings you use on the package.

gpbl:
Looks like the `toEthiopicDate` is not working anymore. Is there a library we can check to understand how the conversion should work?

gpbl:
I finally managed to complete the work by @temesgen-mulugeta and get the Ethiopic calendar working. I think we should release this and let users test it, since this PR has been open for a while.

Screen Shot 2025-09-14 at 13 56 47

Issue #2671, opened 2025-01-22
Setting the display month to a value outside of the selectable range causes the DayButton component to not render


anthony-bonta-gaf-energy:
## To reproduce

1. Set both startMonth and endMonth to a range value
2. Set the month prop to a value outside of the startMonth and endMonth
3. Render the component

If you’re unable to easily reproduce this bug, consider opening a [support thread](https://github.com/gpbl/react-day-picker/discussions) to discuss it further.

```tsx
import { useMemo, useState } from 'react';
import { addYears, startOfYear, endOfYear } from 'date-fns';
import { DayPicker } from 'react-day-picker';

export function App() {
  const today = useMemo(() => new Date(), []);
  const oldDays = useMemo(() => addYears(today, -10), [today]);
  const [month, setMonth] = useState(oldDays);
  const yearsFuture = endOfYear(addYears(today, 2));
  const yearsAgo = startOfYear(addYears(today, -2));

  return (
    <DayPicker
      captionLayout='dropdown'
      endMonth={yearsFuture}
      month={month}
      onMonthChange={setMonth}
      mode='single'
      hideNavigation
      required={false}
      selected={oldDays}
      startMonth={yearsAgo}
      style={{ backgroundColor: 'white' }}
    />
  );
}
```

CodeSandbox: https://playcode.io/2231858

## Actual Behavior

The table cells render for the month and the actual months in the drop down are correctly disabled.  However, the buttons in each cell do not render at all.

Note that you can still work around this by selecting a year in the range.

## Expected Behavior

I would expect a set of buttons to be rendered even if they're outside of the range.  The use case for this is a range of +-1 year and the user selected a value over a year ago and we want to reflect that in the selection and make them update it.

## Screenshots

![Image](https://github.com/user-attachments/assets/6b030529-72cf-4c07-88ef-54c54f8ee995)

rodgobbi:
Hi @anthony-bonta-gaf-energy 
Even if the day buttons would be displayed outside the valid range, they would be displayed disabled, besides all the month options being disabled, which can feel confusing for a user, and in the behavior you described it's expected that the user should select an year that is already selected by default in the dropdown to enable the calendar for selection, therefore the steps to put the `DayPicker` in a valid month is very hidden.
Is there a specific reason for this use case?
Wouldn't it be better to programmatically put the `DayPicker` in a state where the user can start selecting valid dates?

gpbl:
I agree that this is an unexpected state of the calendar.

[...]

PR #2672, opened 2025-01-25
fix: display calendar in a valid month when month prop is invalid

rodgobbi:
## What's Changed

Passing a `month` prop that is outside the `startMonth` and `endMonth` range breaks the calendar.
The PR fixes this edge case by always displaying a valid month as suggested in https://github.com/gpbl/react-day-picker/issues/2671#issuecomment-2613980971.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/useCalendar.ts
+++ b/src/useCalendar.ts
@@ -96,7 +96,8 @@ export function useCalendar(
   const initialMonth = getInitialMonth(props, dateLib);
   const [firstMonth, setFirstMonth] = useControlledValue(
     initialMonth,
-    props.month ? startOfMonth(props.month) : undefined
+    // initialMonth is always computed from props.month if provided
+    props.month ? initialMonth : undefined
   );
 
   useEffect(() => {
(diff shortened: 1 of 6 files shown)


src/useCalendar.ts
```diff
@@ -96,7 +96,8 @@ export function useCalendar(
   const initialMonth = getInitialMonth(props, dateLib);
   const [firstMonth, setFirstMonth] = useControlledValue(
     initialMonth,
-    props.month ? startOfMonth(props.month) : undefined
+    // initialMonth is always computed from props.month if provided
+    props.month ? initialMonth : undefined
```
gpbl, line 100: Good catch! I always thought that `useControlledValue` call was hiding some bugs.

Worth to add a test to `getInitialMonth.test.ts`. Would you add one? Thanks!

rodgobbi: For sure, just pushed more tests 👍 
I think `getInitialMonth.test.ts` was testing well in most of the use cases with the `month` prop, but in any case, I expanded existing tests and added more.
Let me know if you think it would be useful to add tests for the `useCalendar` specifically. We may not need it because it's indirectly tested by other test cases.

gpbl (approved):
Looking great, thanks @rodgobbi 🙇

PR #2673, opened 2025-01-26
performance: add sideEffects property to package.json

rodgobbi:
Idea from backlog:
https://github.com/users/gpbl/projects/3/views/1?pane=issue&itemId=92153364

## What's Changed

Add `sideEffects` property to the `package.json` to optimize the bundling of the lib.

CSS files can be imported without using any exports but should not be excluded, because of that the CSS files are marked as having side effects.
From what I tested, even using `"sideEffects": false` doesn't cause any problem to Rollup, the styles are correctly bundled, but for Webpack to bundle the CSS files we need to list them in the "sideEffects" property, hence the value is `"sideEffects": ["**/*.css"]`. ([Webpack docs](https://webpack.js.org/guides/tree-shaking/))

This PR has a basic Webpack setup to test how it would bundle the library.
I left it for testing but it can be removed if it doesn't bring much value.

The screenshots below show the output of using Webpack with different values for the `sideEffects` property.

`"sideEffects": false`

sideEffects false webpack

(The CssModules works fine due to the way the CSS is imported and used)
sideEffects false webpack CssModules

`"sideEffects": ["**/*.css"]`

sideEffects css webpack

## Notes

@gpbl I checked if there are any side effects based on the definition of "side effect" from Webpack and Rollup and I couldn't find any besides the CSS files. I used Cursor to scan all files in the repo with different prompts to be sure.
Let me know if you can think of any issues or any specific bundling scenario that could cause any issues.

I also checked the bundle size with the `sideEffects` property and without the property, both for Rollup and for Webpack, and in both cases, there's no difference in size using the property.

I concluded this is because the library is mainly composed of a component that is always used when importing the `DayPicker` and all the different dependencies across the library are always used when the `DayPicker` is imported.

Other extra features provided by the package are exported from different entry points, for example, the Persian calendar, so it doesn't affect who is not importing it.
The same happens to the `react-day-picker/locale` entry point, which reexports `date-fns/locale`, and I checked that `date-fns` has the `"sideEffects": false` property, so importing partially the `locale`s doesn't bloat the bundle size.

@gpbl This idea could be scrapped because it doesn't bring much value, we have more information after the tests I did.
But there's also the argument that almost all of the library code has no side effect, except the CSS files, so it makes sense to use the property and have it optimized for bundlers for any future iterations that could benefit from it.

### DIFF
--- a/examples-app/README.md
+++ b/examples-app/README.md
@@ -1,3 +1,5 @@
 # Examples app
 
 This small app run the examples in DayPicker and can be used to test the library.
+
+The primary development and build tool is Vite, but there's also a Webpack setup to test how the library would work using Webpack.
--- a/examples-app/package.json
+++ b/examples-app/package.json
@@ -6,8 +6,10 @@
   "scripts": {
     "dev": "vite",
     "build": "tsc -b && vite build",
+    "preview": "vite preview",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
-    "preview": "vite preview"
+    "webpack:dev": "webpack serve --mode development",
+    "webpack:build": "tsc -b && webpack --mode production"
   },
   "dependencies": {
     "react": "^19.0.0",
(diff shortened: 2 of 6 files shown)


gpbl (commented):
Thanks @rodgobbi great work and super useful analysis 💯 

I would however keep the webpack packages out of this PR - as we want it easy to rollback in case there are unexpected issue caused by this new setting. Would you agree?

rodgobbi:
> Thanks @rodgobbi great work and super useful analysis 💯
> 
> I would however keep the webpack packages out of this PR - as we want it easy to rollback in case there are unexpected issue caused by this new setting. Would you agree?

Makes sense, it causes the repo to be bloated.
I'll remote it and request another review.
I'll first try to find way to document the webpack setup for future tests when it's needed. I may do this outside the repo in the end.

rodgobbi:
FYI @gpbl I confirmed that the `date-fns/locales` re-export is tree-shakable, so I'll close this [idea from the backlog](https://github.com/users/gpbl/projects/3/views/1?pane=issue&itemId=93210064).

I checked that the bundle size incrementally increases as I use more locale imports:
image

image

image

PR #2679, opened 2025-02-02
New: custom select dropdown example

rodgobbi:
## What's Changed

- Add an example with Custom Select `Dropdown` using a simplified version of `shadcn/ui`'s [Select](https://ui.shadcn.com/docs/components/select).
- Update [Custom Components](https://daypicker.dev/guides/custom-components) docs, adding the new example for reference.

## Notes

Using `shadcn/ui`'s [Select](https://ui.shadcn.com/docs/components/select) requires adding `@radix-ui/react-select`, which brings a lot of dependencies, but are only needed for development to use in the new example.
Another possible approach would be implementing a custom `Select` that is very simplified (also missing a lot of accessibility features already provided by `radix-ui`) for the example.

## Demo

CustomDropdown

https://github.com/user-attachments/assets/0fb12574-049a-40de-bcac-b7a4642546b5


examples/CustomDropdown/CustomDropdown.tsx
```diff
@@ -0,0 +1,83 @@
...
+  SelectStyles
+} from "./Select";
+
+const ContainerContext = createContext<HTMLDivElement | null>(null);
+
+export function CustomSelectDropdown(props: DropdownProps) {
+  const { options, value, onChange, "aria-label": ariaLabel } = props;
+  const container = use(ContainerContext);
+
+  const handleValueChange = (newValue: string) => {
```
gpbl, line 21: The idea is to add a `onValueChange` to the props:

```suggestion
  const { options, value, onValueChange, "aria-label": ariaLabel } = props;
  const container = use(ContainerContext);

  const handleValueChange = onValueChange
```

gpbl (approved):
Look great to me, thanks for digging into that select library.

- We should implement an `onValueChange` as we suggested in https://github.com/gpbl/react-day-picker/discussions/2676#discussioncomment-11984571 and then update the example.
- Keyboard navigation seems broken, but I can't say if it’s an issue with the original library.

rodgobbi:
> Keyboard navigation seems broken, but I can't say if it’s an issue with the original library.

Good catch @gpbl
I noticed this but missed writing a note about it.
The issue only happens in the docs, when the example is mounted in the doc window sandbox, I don't know if the shadow-dom sandbox causes some issue to radix, but running the example normally on the browser works fine.

PR #2680, opened 2025-02-06
Fix deprecation suggestion of select event handler types

timothyis:
## What's Changed

When I was updating from v8 -> v9 I came across the deprecation notice for the `SelectSingleEventHandler`, which gave me the deprecation notice before this PR. When using the suggestion, I was met with:

```
ts: Type '"single"' has no properties in common with type '{ mode?: Mode | undefined; required?: boolean | undefined; }'.
```

Implementing the change I've made on this PR made the warning disappear, though you can also do `SelectHandler<{ required: true }>` and I'm not sure if that's correct. So maybe a type change needs to be made also.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

### DIFF
--- a/src/types/deprecated.ts
+++ b/src/types/deprecated.ts
@@ -116,20 +116,20 @@ export type InternalModifier =
 
 /**
  * @ignore
- * @deprecated This type will be removed. Use `SelectHandler<"single">` instead.
+ * @deprecated This type will be removed. Use `SelectHandler<{mode: "single"}>` instead.
  */
 export type SelectSingleEventHandler = PropsSingle["onSelect"];
 
 /**
  * @ignore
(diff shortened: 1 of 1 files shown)


rodgobbi:
Hi @timothyis

Thanks for the PR!

It's indeed a mistake of the docs, as the [generic type for the `SelectHandler`  type](https://github.com/gpbl/react-day-picker/blob/main/src/types/selection.ts#L48) needs to extend `{ mode?: Mode | undefined; required?: boolean | undefined }`.

Could you fix the linter in your changes?
So the CI passes and we can merge the changes.

gpbl:
Thanks @timothyis!

PR #2684, opened 2025-02-16
feat: month animation

rodgobbi:
Idea from the board:
https://github.com/users/gpbl/projects/3/views/1?pane=issue&itemId=9360383

## Description

Add a feature to animate the calendar changing the displayed month. The feature can be enabled by setting the prop `animate` to `true`.

The core idea in the approach of this implementation is to imperatively manipulate the DOM when the displayed months change, because doing it with only React APIs is very convoluted and has bad performance.
All the DOM manipulation happens inside a `useLayoutEffect` to set up the animation right before the browser renders.

There is a new set of CSS classes that are imperatively applied to the DOM to trigger the animations that are based on the calendar changing to a month that is after the previous month or before it because depending on this condition, the calendar weeks should animate left or right, and is inverted when the direction is `rtl`.

The animation is disabled when navigating the calendar with the keyboard arrows because focusing on the month being animated can cause issues with the animation. Animation can be worse for a11y, therefore disabling for keyboard navigation seems better.

## Demo

https://github.com/user-attachments/assets/d7173390-9f4c-4861-a1e7-802e48f43c54

https://github.com/user-attachments/assets/4579b9e5-6b65-4940-9e0b-48d0055ff0e5

https://github.com/user-attachments/assets/83d7d28e-ceb1-4e39-93d3-6ce2fa4d90e0


examples-app/src/App.css
```diff
@@ -48,3 +48,40 @@
 .read-the-docs {
   color: #888;
 }
+@keyframes slideInLeft {
```
gpbl, line 51: Here and in the code should use "previous"/"next" instead of "left/right" and consider the `dir="rtl"` case (e.g. arabic calendar).

src/DayPicker.tsx
```diff
@@ -256,6 +263,118 @@ export function DayPicker(props: DayPickerProps) {
 
   const dataAttributes = getDataAttributes(props);
 
+  const rootRef = useRef<HTMLDivElement>(null);
+  const rootElementSnapshot = useRef<HTMLElement>(null);
+  const previousMonths = useRef(months);
+
+  useLayoutEffect(() => {
```
gpbl, line 270: This is nice - could we place it in a new `useTransition` and unit test it?

src/DayPicker.tsx
```diff
@@ -256,6 +263,118 @@ export function DayPicker(props: DayPickerProps) {
...
+      previousMonths.current[0].date
+    );
+    const slideLeft = dateLib.isAfter(
+      months[0].date,
+      previousMonths.current[0].date
+    );
+    previousMonths.current = months;
+    if (rootRef.current) {
+      const previousMonthElement =
+        rootElementSnapshot.current?.querySelector(`[data-month]`);
```
gpbl, line 286: For supporting multiple months, consider adding a value , such as `data-month="yyyy-mm"`.

src/DayPicker.tsx
```diff
@@ -256,6 +263,118 @@ export function DayPicker(props: DayPickerProps) {
...
+          if (currentMonthElement.contains(previousMonthElement)) {
+            currentMonthElement.removeChild(previousMonthElement);
+          }
+        };
+
+        // animate old displayed month
+        previousMonthElement.style.pointerEvents = "none";
+        previousMonthElement.style.position = "absolute";
+        previousMonthElement.style.overflow = "hidden";
+        previousMonthElement.setAttribute("aria-hidden", "true");
```
gpbl, line 336: Is `aria-hidden` added for safety, or it is actually required?

src/DayPicker.tsx
```diff
@@ -256,6 +263,118 @@ export function DayPicker(props: DayPickerProps) {
...
+            weeksElement.style.animation = "";
+          }
+          currentMonthElement.style.overflow = "";
+          if (currentMonthElement.contains(previousMonthElement)) {
+            currentMonthElement.removeChild(previousMonthElement);
+          }
+        };
+
+        // animate old displayed month
+        previousMonthElement.style.pointerEvents = "none";
```
gpbl, line 333: Would this cause the focus being "lost"? Consider adding a test case.

src/DayPicker.tsx
```diff
@@ -256,6 +263,118 @@ export function DayPicker(props: DayPickerProps) {
...
+        }
+
+        const previousWeeksElement =
+          previousMonthElement.querySelector(`[data-weeks]`);
+        if (
+          previousWeeksElement &&
+          previousWeeksElement instanceof HTMLElement
+        ) {
+          previousWeeksElement.style.animation = slideLeft
+            ? "slideOutLeft 0.3s forwards"
```
gpbl, line 354: In a future version, the transition could be customizable via props. 

It would be useful to try a up/down animation (as opposed to left-right) to see how different transitions would actually work, or which challenges could they pose to the developer.

src/components/Root.tsx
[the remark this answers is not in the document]
rodgobbi: Good point, no idea how we need to treat the `data-` attributes being added in this PR, but we can make sure it's not a breaking change by only passing it when the animation is enabled.

src/DayPicker.tsx
rodgobbi: Indeed, I'm thinking of extracting this to a custom hook to isolate this logic, I'll see how easy it is to unit test it.
FYI, [useTransition](https://react.dev/reference/react/useTransition) is a native hook from React so I'll pick another name.

src/DayPicker.tsx
rodgobbi: Just to be safe, and it's technically correct, the previous month's DOM will be only mounted for the visual animation and should not represent anything semantically.

src/DayPicker.tsx
rodgobbi: I checked that `pointer-events` don't affect focus, but fortunately the focus most of the time works fine in this approach, the problem is the focus being applied to the next month when sliding in from the right, for some reason this causes issues to all animations, but semantically the focus is on the correct element when changing month.

src/DayPicker.tsx
rodgobbi: Indeed good point,
I'll work on an API that allows it already and will test it by trying to write a custom components example that has the animation up and down, writing such example is a good validation of the API allowing customization.

src/DayPicker.tsx
gpbl: Oh cool - I'd better skip this attribute if we are not sure it is needed. What do you think?

src/components/Root.tsx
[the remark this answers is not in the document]
rodgobbi: Just update the implementation to only apply the props when animate is true, to avoid any breaking changes, this is especially needed for the `Root` ref prop.

src/DayPicker.tsx
rodgobbi: Done, animation works for multiple months

src/DayPicker.tsx
rodgobbi: I can remove it, but I still think it's more semantically correct having it because the old month being animated out of the page is just for visualization purposes, it should not have any semantic meaning to the page while it's animated.

src/style.css
```diff
@@ -304,3 +303,89 @@
...
+  animation: rdp-animation_fade_out_keyframes var(--rdp-animation_duration) var(--rdp-animation_timing) forwards;
+}
+
+.rdp-animation_new_month_caption_is_before {
+  animation: rdp-animation_fade_in_keyframes var(--rdp-animation_duration) var(--rdp-animation_timing) forwards;
+}
+
+.rdp-animation_old_month_caption_is_after {
+  animation: rdp-animation_fade_out_keyframes var(--rdp-animation_duration) var(--rdp-animation_timing) forwards;
+}
```
gpbl, line 391: Missing EOF new line: 

```suggestion
}

```

src/types/props.ts
```diff
@@ -245,6 +245,13 @@ export interface PropsBase {
    * @see https://daypicker.dev/docs/customization#showweeknumber
    */
   showWeekNumber?: boolean;
+  /**
+   * Experimental. Animate navigating between months.
+   *
+   * @experimental
```
gpbl, line 251: Interesting take. Not sure if this should be classified as "experimental".

```suggestion
   * Animate navigating between months.
```

src/DayPicker.tsx
```diff
@@ -256,6 +256,186 @@ export function DayPicker(props: DayPickerProps) {
...
+
+        const monthAnimationClass = isAfterPreviousMonth
+          ? classNames[AnimationClass.animation_new_month_is_after]
+          : classNames[AnimationClass.animation_new_month_is_before];
+
+        currentMonthElement.style.position = "relative";
+        currentMonthElement.style.overflow = "hidden";
+        const monthCaptionElement = currentMonthElement.querySelector(
+          `[data-month-caption-container]`
+        );
```
gpbl, line 352: Suggesting some refactoring for readability (maybe after we move the code to its own Hook and test it):

- `monthCaption` → just use `caption` in any instance (reasoning: there are no other "captions" in DayPicker, so the "month" prefix is superfluous)
- once we’ve moved the code to another hook, it’s evident that it’s intended for animation, so we can omit this "descriptor".
- `*Element` → `*El`

```suggestion
        const captionClassName = isAfterPreviousMonth
          ? classNames[Animation.animation_new_month_caption_is_after]
          : classNames[Animation.animation_new_month_caption_is_before];

        const monthClassName = isAfterPreviousMonth
          ? classNames[Animation.animation_new_month_is_after]
          : classNames[Animation.animation_new_month_is_before];

        currentMonthEl.style.position = "relative";
        currentMonthEl.style.overflow = "hidden";

        const captionEl = currentMonthEl.querySelector(`[data-caption-container]`);
```

gpbl (commented):
This looks so great, I love how you nailed it!

Next steps to me would be:

- Move the `useLayoutEffect` into a `useAnimation` hook. 
- Add a test for the new `useAnimation`.
- Improve readability with some refactoring.

src/UI.ts
```diff
@@ -100,6 +100,50 @@ export enum SelectionState {
   selected = "selected"
 }
 
+/** The animation classes used by DayPicker for month transitions. */
+export enum AnimationClass {
```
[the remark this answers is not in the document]
rodgobbi: Understandable, some things are really difficult to name properly

src/UI.ts
```diff
@@ -100,6 +100,50 @@ export enum SelectionState {
...
+  /**
+   * Applied to the new mounted displayed month when it is after the old
+   * displayed month.
+   */
+  animation_new_month_is_after = "animation_new_month_is_after",
```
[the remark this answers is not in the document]
rodgobbi: Nice, I think it's good!
The naming is something I wanted more feedback, exit and enter indeed feels more intuitive.

src/types/props.ts
rodgobbi: All good for me, this part I just copied and pasted from your other PR, so I left the way it was

src/DayPicker.tsx
rodgobbi: Thnks for the feedback 👍 
One last point, I just chose `monthCaption` to follow the component name `MonthCaption`, I think it's easier to associate with it by following the exact name.
I'll change in any case, let me know if you change your mind.

src/UI.ts
```diff
@@ -100,6 +100,26 @@ export enum SelectionState {
   selected = "selected"
 }
 
+/** CSS classes used for animating months and captions. */
+export enum Animation {
+  /** Applied to the entering month when it is after the exiting month. */
+  animation_enter_month_weeks_is_after = "animation_enter_month_weeks_is_after",
```
rodgobbi, line 106: @gpbl FYI, I changed the name of these classes to `month_weeks` to follow the same patter of the `month_caption` and be more explicit that it's being applied to the `Weeks` component.

src/style.module.css
```diff
@@ -146,10 +149,6 @@
   fill: var(--rdp-accent-color);
 }
 
-.root[dir="rtl"] .nav .chevron {
-  transform: rotate(180deg);
```
rodgobbi, line 150: These are duplicated styles that I removed, you can see the same thing applied in the next lines.

gpbl: Good find! Thanks!

gpbl (commented):
Fanstastic work @rodgobbi !

The work is done here and we may want to polish a bit the code before merging. 

My suggestions:

- `useAnimation`: Remove the return value and clean up the arguments.
- Shorten CSS class and keyframe names to avoid cumbersome customization.
- Rename animation-related data attributes like `data-month-container` → `data-animated-month`.

If you prefer, I can propose my suggested changes in a separate branch.

src/DayPicker.tsx
```diff
@@ -317,16 +327,19 @@ export function DayPicker(props: DayPickerProps) {
               formatters,
               dateLib
             );
-
             return (
               <components.Month
+                data-month-container={props.animate ? "true" : undefined}
```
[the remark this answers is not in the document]
rodgobbi: good idea 👍 
it makes more clear it's for the animation

src/useAnimation.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { nextButton, previousButton } from "../test/elements";
+
+import { DayPicker } from "./DayPicker";
+
+jest.setSystemTime(new Date(2025, 1, 10));
+
+const getMonthContainers = () => [
+  ...document.querySelectorAll(`[data-month-container]`)
+];
+const getMonthCaptionContainers = () => [
```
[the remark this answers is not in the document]
rodgobbi: The reason I didn't think of putting these queries in `/test/elements.ts` is because they are only used in the animation test, for me it would be only useful to have them there if they were reused somewhere, what do you think of waiting for when we have it used in more than one place to refactor it to the `/test/elements.ts`?
To avoid making the module bloated with things that are only used in one place.

> why are you spreading the array here?

It's because `querySelectorAll` returns a `NodeList`, which doesn't have the `Array` methods, so the spread converts it to an array.

`NodeList` is an `iterable`, so we could use `for ...of` to iterate over it if preferred.

src/useAnimation.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+  ...document.querySelectorAll(`[data-weekdays-container]`)
+];
+const getMonthWeeksContainers = () => [
+  ...document.querySelectorAll(`[data-weeks-container]`)
+];
+
+describe("useAnimation", () => {
+  describe("animate prop is falsy", () => {
+    it("should not change the default props", () => {
+      render(<DayPicker />);
```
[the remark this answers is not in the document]
rodgobbi: make sense, I'll rephrase a bit because `not render any data-animated element` looks like the element will not be rendered at all.

src/useAnimation.ts
```diff
@@ -0,0 +1,201 @@
...
+      previousMonths.length === 0 ||
+      months.length !== previousMonths.length ||
+      // skip animation if a day is focused because it can cause issues to the animation and is better for a11y
+      focused ||
+      animatingRef.current
+    ) {
+      return;
+    }
+
+    const isSameMonth = dateLib.isSameMonth(
```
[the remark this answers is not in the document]
rodgobbi: One straightforward case, when the component mounts, so this prevents the calendar triggering the animation.

But in general this is a good safety check to have, many different dependencies may change, very likely a `CalendarDay` or `CalendarMonth` object reference may change, but if the first month is the same, it means the calendar didn't change months, so we should not trigger an animation, otherwise the day picker would animate the same month.

Do you think it would be helpful to add more comments?

In general for this feature I added quite a lot of comments because it's full of edge cases and caveats, so I concluded it's good to give all the context of the tricky caveats that are hard to grasp just by reading the code.

src/useAnimation.ts
```diff
@@ -0,0 +1,201 @@
...
+      months[0].date,
+      previousMonths[0].date
+    );
+
+    if (isSameMonth) {
+      return;
+    }
+
+    const previousMonthEls = [
+      ...(previousRootElSnapshot?.querySelectorAll(`[data-month-container]`) ??
```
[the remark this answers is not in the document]
rodgobbi: actually it's fine to use the original object, it's just because `querySelectorAll` returns a `NodeList` and we don't have access to array methods with it, I wrote also about this in this previous [comment](https://github.com/gpbl/react-day-picker/pull/2684#discussion_r1975101327).

Lemme know if you'd like to add a comment or I could try to use `for...of` directly

src/style.css
```diff
@@ -304,3 +303,89 @@
...
+
+.rdp-animation_enter_month_weeks_is_after {
+  animation: rdp-animation_slide_in_right_keyframes var(--rdp-animation_duration) var(--rdp-animation_timing) forwards;
+}
+
+.rdp-animation_exit_month_weeks_is_before {
+  animation: rdp-animation_slide_out_left_keyframes var(--rdp-animation_duration) var(--rdp-animation_timing) forwards;
+}
+
```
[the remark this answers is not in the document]
rodgobbi: Indeed sounds better being shorter.
But a couple of points to discuss about the new format you suggested:

1. What about adopting the same terms already established by other libs like [react-transtion-group](https://reactcommunity.org/react-transition-group/css-transition), instead of `in` and `out`, we stick with `enter` and `exit`. You suggested this naming in the previous comments, and I think it's great because we reuse the same pattern already used by other libs. We could check if there are other commonly used libs that use another naming pattern.

2. About the naming of the keyfram animation, I think it's not ideal to use `prev` or `next`, the keyframe animation itself represents sliding in from left or right, or sliding out to the left or right, and it doesn't change, but with the CSS classes and `rtl` attribute we use different keyframes based on the case, the keyframe will still represent left or right transitions, but the class represents how to animate the elements based on the month entering the DOM being next or previous (after/before), and the text direction preference.

src/style.css
```diff
@@ -304,3 +303,89 @@
 .rdp-focusable {
   cursor: pointer;
 }
+
+@keyframes rdp-animation_slide_in_left_keyframes {
```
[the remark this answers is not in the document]
rodgobbi: Discussed this in point 2 on the https://github.com/gpbl/react-day-picker/pull/2684#discussion_r1975155793

src/style.css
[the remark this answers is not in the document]
gpbl: @rodgobbi, you’re absolutely right. Let’s maintain the usage of “enter” and “exit” instead of “in” and “out,” and “left” and “right” for the keyframe names.

src/useAnimation.test.tsx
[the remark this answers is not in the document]
gpbl: @rodgobbi Agree, thanks for the explanation!

src/useAnimation.ts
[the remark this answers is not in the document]
gpbl: It's fine this way then, I just wanted be sure there was a reason for it. Thanks for your explanation!

src/useAnimation.ts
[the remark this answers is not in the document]
gpbl: @rodgobbi now it makes sense to me. Thanks for clarifying!

gpbl (approved):
Looking great! Thanks @rodgobbi for your patience addressing my reviews.

Issue #2630, opened 2024-12-18
Current focus can be lost if next or previous month buttons become disabled


mrdaniellewis:
## To reproduce

```tsx
import { DayPicker } from "react-day-picker";
import "react-day-picker/style.css";

export default function App() {
  return (
    mode="single"
    disabled={{ after: new Date() }}
    endMonth={new Date()}
  );
}
```

CodeSandbox: https://codesandbox.io/p/sandbox/daypicker-bug-report-forked-cff2m5?workspaceId=ws_RGtEKXawCzJL1meZNuGGuQ

Using the keyboard:

1. tab to the previous month button
2. click
3. tab to the next month button
4. click

## Actual Behaviour

The next month button becomes disabled and focus is lost.

## Expected Behaviour

Something should be focused.  I'm not 100% sure the correct pattern:

Option 1: Do not disable the button, mark with `aria-disabled="true"`, and `tabIndex={-1}` to take it out of tab order.  The button will remain focused, but will act as disabled.
Option 2: Focus something nearby, eg the other month navigation button.

https://adrianroselli.com/2023/08/where-to-put-focus-when-deleting-a-thing.html

I'm not sure how much of an issue this is.  I believe it is an issue and it can cause screen-readers to move the current focus to the body throwing the user to the top of the page.  However, I'm not completely sure this is that catastrophic an issue with the latest browsers and screen-readers.

rodgobbi:
@gpbl 
Have you started working on it already?
If not,  I can work on it.
If you already know how to fix this accessibility issue, let me know, otherwise, I'll ask a friend who specializes in accessibility for some directions.
And if you think there are other potential accessibility improvements let me know because I can research more while fixing the issue.

gpbl:
This is indeed an accessibility bug and we have to find a fix for it 🕵 

Can we use `aria-disabled` instead of `disabled`? These two attributes have been causing some trouble in DayPicker, and I don't understand the real difference between them. 

On the other side, this could be a breaking change. Developers using the `[disabled]` selector in their CSS would need to change it to `[aria-disabled="true"]`.

rodgobbi:
@gpbl 
Besides the breaking change, I think it also makes sense for the button to be `disabled`, for all types of users, and we get the correct accessibility by default from the browser.

I'll research how to handle the focus for this type of use case, of disabling the element focused on the page, and share what I learn here.

gpbl:
Some ideas :

[...]

PR #2685, opened 2025-02-16
fix: focus lost when navigation button is disabled

gpbl:
Fixes issue #2630 by replacing the `disabled` attribute with `aria-disabled`, to prevent the focus loss when navigating to the last month.

After some research, I believe this change won’t cause any other accessibility issues.

Breaking change (style): This change may affect the style of those who rely on the `[disabled]/:disabled` CSS selectors. To fix this, use `aria-disabled` instead.

```diff
- .rdp-button_next[disabled="true"],
- .rdp-button_next:disabled,
+ .rdp-button_next[aria-disabled="true"],
- .rdp-button_previous:disabled,
- .rdp-button_previous[disabled="true"],
+ .rdp-button_previous[aria-disabled="true"] {
   /* Existing custom style */
}
```

https://github.com/user-attachments/assets/986d55eb-18b0-431f-9eab-86e1a39e4262

### DIFF
--- a/src/components/Nav.tsx
+++ b/src/components/Nav.tsx
@@ -1,4 +1,8 @@
-import React, { type MouseEventHandler, type HTMLAttributes } from "react";
+import React, {
+  type MouseEventHandler,
+  type HTMLAttributes,
+  useCallback
+} from "react";
 
 import { UI } from "../UI.js";
 import { useDayPicker } from "../useDayPicker.js";
@@ -31,15 +35,39 @@ export function Nav(
     labels: { labelPrevious, labelNext }
   } = useDayPicker();
 
+  const handleNextClick = useCallback(
+    (e: React.MouseEvent<HTMLButtonElement>) => {
+      if (nextMonth) {
+        onNextClick?.(e);
+      } else {
+        e.preventDefault();
+        e.stopPropagation();
+      }
+    },
+    [nextMonth, onNextClick]
+  );
+
+  const handlePreviousClick = useCallback(
+    (e: React.MouseEvent<HTMLButtonElement>) => {
+      if (previousMonth) {
+        onPreviousClick?.(e);
+      } else {
+        e.preventDefault();
+        e.stopPropagation();
+      }
+    },
+    [previousMonth, onPreviousClick]
+  );
+
   return (
     <nav {...navProps}>
       <components.PreviousMonthButton
         type="button"
         className={classNames[UI.PreviousMonthButton]}
         tabIndex={previousMonth ? undefined : -1}
-        disabled={previousMonth ? undefined : true}
+        aria-disabled={previousMonth ? undefined : true}
         aria-label={labelPrevious(previousMonth)}
-        onClick={props.onPreviousClick}
+        onClick={handlePreviousClick}
       >
         <components.Chevron
           disabled={previousMonth ? undefined : true}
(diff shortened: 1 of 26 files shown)


rodgobbi:
@gpbl
I discussed the https://github.com/gpbl/react-day-picker/issues/2630 issue with my friend, who has experience with a11y, and he mentioned that the core issue is not that the page loses focus when the button becomes disabled but that the action no longer makes sense and has no effect.
To illustrate with an example, with this PR, even if the button is still focused and not fully disabled, a user using AT would still read the description of the button as "Go to the Next Month", they would try to press it and not get any feedback.

It seems like this PR would not fully fix the a11y issues.
May I do more research and come back to you in this PR with more information?

My friend recommended checking the [Adobe Spectrum](https://react-spectrum.adobe.com/react-spectrum/index.html) design system, which has great accessibility, and they have a [DatePicker](https://react-spectrum.adobe.com/react-spectrum/DatePicker.html), I'll study how it behaves with AT to provide more feedback to the `React DayPicker`.

Let me know what you think.

gpbl:
I agree that further research is required here.

At the very least, I would defer to the developer to address these kinds of issues. However, the issue of losing focus appears to be the most significant one to me. 

>  the core issue is not that the page loses focus 

Well, this is the accessibility issue we need to address. Is it indeed an issue, or are we mistaken? 

> the action no longer makes sense and has no effect.

Shouldn’t this be communicated via an aria-live region somehow? Which is worse - losing focus or having action become unavailable? :thinking:

rodgobbi:
@gpbl thanks for the patience

> Well, this is the accessibility issue we need to address. Is it indeed an issue, or are we mistaken?

I'll do some further research, I'm not an expert in a11y, so I'll need some time to come up with an answer that I feel confident to give.

> Shouldn’t this be communicated via an aria-live region somehow?

My initial impression is yes, I'll check what other applications and libs do.

> Which is worse - losing focus or having action become unavailable?

My friend gave the opinion that both are not enough for a11y. I also asked what if the focus changes to the other button to navigate the months to the other direction, but he mentioned it's also not great if the users using AT are not informed of what happened.

src/components/Nav.tsx
```diff
@@ -31,15 +35,39 @@ export function Nav(
...
+    [nextMonth, onNextClick]
+  );
+
+  const handlePreviousClick = useCallback(
+    (e: React.MouseEvent<HTMLButtonElement>) => {
+      if (previousMonth) {
+        onPreviousClick?.(e);
+      } else {
+        e.preventDefault();
+        e.stopPropagation();
```
rodgobbi, line 56: I think it's better internally only ignore the button click and let it bubble up, meaning not stop its propagation and default behavior, so any developer doesn't lose such events.
E.g.: a developer renders a custom button which has a tooltip saying the day picker reached the maximum allowed months, they may need this event to propagate to show the tooltip.

rodgobbi (commented):
I researched and tested this implementation more, and it's great!
I tested the screen reader on different examples of buttons with `aria-disabled` and with the example of this PR specifically, the screen reader will announce the button is "inactive" due to `aria-disabled`, providing the same experience as users visually seeing the button become disabled.

Great article about disabled buttons: https://css-tricks.com/making-disabled-buttons-more-inclusive/

src/components/Nav.tsx
gpbl: Agree! Thanks for the heads up!

PR #2694, opened 2025-03-01
docs: update for new `animate` prop

gpbl:
- added docs and playground
- renamed keyframes
- add `AnimationCustom` example (still not working well for weeks animation)


src/style.module.css
```diff
@@ -342,47 +342,43 @@
...
+@keyframes rdp_caption_prev_exit {
+  from {
+    opacity: 1;
+  }
+  to {
+    opacity: 0;
+  }
 }
 
-@keyframes rdp-fade_in {
```
rodgobbi, line 385: The same as the comment before, `fade_in` is used in 2 classes, `caption_next_enter` and `caption_prev_enter`, so it's more confusing naming it `caption_next_enter` than what it represents visually, which is a `fade_in` animation.

src/style.css
```diff
@@ -342,47 +342,43 @@
   }
 }
 
-.rdp-weeks_next_enter {
```
rodgobbi, line 345: Does it make sense to flip the words in the class names?
Naming the classes like `weeks_enter_next`?

I don't know if it's just for me how I think logically about it, but for me to reason about what is happening in the animation, I need to thnk first about the "entering" or "exiting" month, than if it's before or after, so I usually read "enter" -> "next" to understand the meaning of the class.

Also, what do you think of changing the naming `prev` and `next` to use the naming from `date-fns` for comparing dates, which is `after` and `before`.
IMO reusing the same concept used for other date libs makes it easier to understand.
So with my suggestion, the class I referenced in this comment would be named `rdp-weeks_enter_after`.

rodgobbi (changes requested):

src/style.module.css
```diff
@@ -306,7 +306,7 @@
   cursor: pointer;
 }
 
-@keyframes rdp-slide_in_left {
+@keyframes rdp-weeks_prev_enter {
```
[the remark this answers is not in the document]
gpbl: Yeah I agree, after a second thought you are right here! I restored the original names.

src/style.css
gpbl: I do indeed prefer using “before/after”!

However, I’m not convinced that it should be used as a suffix. To me, “before” and “after” are adjectives that modify “weeks” or “caption.” I’m also not entirely sure what it means in this context. :)

I renamed it using before/after in [commit `1744724`](https://github.com/gpbl/react-day-picker/pull/2694/commits/174472439c951913002d0009982b54abf084843a).

src/style.module.css
```diff
@@ -342,47 +342,43 @@
...
+  animation: rdp-weeks_next_enter var(--rdp-animation_duration)
+    var(--rdp-animation_timing);
 }
 
 .weeks_next_exit {
-  animation: rdp-slide_out_right var(--rdp-animation_duration)
-    var(--rdp-animation_timing) forwards;
-}
-
-[dir="rtl"] .weeks_next_enter {
```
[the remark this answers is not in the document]
gpbl: I’m not sure what’s wrong with the screenshot above :) Isn't correct that they slide left when clicking the next month? I’ll leave this part the, for further investigation in the future.

gpbl:
@rodgobbi I’ve removed the “Advanced Animation” example. To properly customize animations, we’ll need to modify the HTML structure since the HTML Table isn’t working well with overflows.

I’ve also adjusted the class names based on your feedback - I appreciate your help there!

src/style.module.css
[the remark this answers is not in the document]
rodgobbi: I tested [AirBnB with a `rtl` language to confirm](https://ar.airbnb.com/), and the animation of the calendar is the same as a `ltr` language when you are clicking the same "left" or "right" buttons, what changes is navigating to the month after or before, but the animation adapts to always follow the visual direction of the button.
Therefore clicking the "left" button slides things from left to right, and the "right" button slides things from right to left, both for `ltr` and for `rtl`.

src/style.css
rodgobbi: All good, thanks for updating it 👍 
Yeah, the other part is more personal taste, I find my way to read the classes to create a mental model of how they are used by the DayPicker.

rodgobbi (dismissed):

rodgobbi:
> @rodgobbi I’ve removed the “Advanced Animation” example. To properly customize animations, we’ll need to modify the HTML structure since the HTML Table isn’t working well with overflows.
> 
> I’ve also adjusted the class names based on your feedback - I appreciate your help there!

Thanks!

And it makes sense to remove the up and down example, it's tricky to do it with the current HTML structure, adding any wrapper for trying to properly set `overflow` for up and down breaks the layout.

Fortunately, the default animation has no issues.

Issue #2699, opened 2025-03-07
Style days as selected also when they are disabled


gpbl:
### Discussed in https://github.com/gpbl/react-day-picker/discussions/2698

<div type='discussions-op-text'>

<sup>Originally posted by **YongThePsibor** March  7, 2025</sup>
react-day-picker is a wonderful picker, I love the excellent design of it. 

recently, I came across a situation where I could not display the selected day in a disabled picker.

I mark the picker as disabled because I have no permission to modify the value.

But the picker has a selected value, I want to display it.

Can anybody tell me how to achieve that?

[Here is a stackblitz show the situation.](https://stackblitz.com/edit/vitejs-vite-a9njjzun?file=src%2FDayPicker.tsx)</div>

PR #2700, opened 2025-03-08
fix: render selected days with selected modifier when disabled

rodgobbi:
Fixes issue https://github.com/gpbl/react-day-picker/issues/2699 by always applying the `selected` modifier if a day is part of the `selected` date.

Notes:
This fix reverts [this change](https://github.com/gpbl/react-day-picker/pull/2171/files#diff-16d85a5debbc177a4de5d5687ff4f6a4d5ed25ad909b77b50d6c22014a551727L75) from a previous fix PR.
What I understood from the context is that the main issue the referred PR fixes is that the disabled days were able to be clicked and selected, so displaying them as `selected` seems correct while ignoring click events coming from them.

Demo:

Tested rendering duplicated `DayPicker`s by rendering them `disabled`.

Screenshot 2025-03-08 at 11 11 49

Added new example for testing the modifier customizations are applied:

Screenshot 2025-03-08 at 11 25 45

### DIFF
--- a/examples/DisabledSelectedDate.test.tsx
+++ b/examples/DisabledSelectedDate.test.tsx
@@ -1,0 +1,37 @@
+import React from "react";
+
+import { gridcell } from "@/test/elements";
+import { render } from "@/test/render";
+
+import { DisabledSelectedDate } from "./DisabledSelectedDate";
+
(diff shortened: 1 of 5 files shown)


gpbl (approved):
That’s great! I have no idea why we accepted that previous behavior. 🤷

PR #2710, opened 2025-03-11
fix(animation): fix issues when navigating months during animation

rodgobbi:
This PR fixes `DayPicker` minor bugs caused when trying to navigate months while animating the month transition.
In the same PR, improve tests and refactor `useAnimation` implementation.

Issues:
- Navigation month buttons are not clickable during animation.
- Navigating months during animation causes visual bug in the animation of the next month transition.

The videos below compare the `DayPicker` before and after the fix, running the animation at 25% the default speed.

Before

 https://github.com/user-attachments/assets/c2e8c3d4-1626-4d67-ac79-e6676d0d3402

After 

https://github.com/user-attachments/assets/2dd2a25e-11dd-4744-a6cd-dbebdf41df07

### DIFF
--- a/src/DayPicker.tsx
+++ b/src/DayPicker.tsx
@@ -303,6 +303,7 @@ export function DayPicker(props: DayPickerProps) {
         >
           {!props.hideNavigation && (
             <components.Nav
+              data-animated-nav={props.animate ? "true" : undefined}
               className={classNames[UI.Nav]}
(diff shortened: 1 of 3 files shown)


gpbl (approved):
Looks good, thanks @rodgobbi ! I run some code cleanup (see last 2 commits).

Issue #2715, opened 2025-03-15
When using screen reader, navigation buttons not announcing month/year when CaptionLayout is dropdown


erstrauch:
Thanks for reporting a bug for DayPicker. Please provide here a brief description of the issue you’ve encountered. 

## To reproduce

```tsx
import { DayPicker } from "react-day-picker";
import "react-day-picker/style.css";

export default function App() {
  return <DayPicker mode="single"  captionLayout="dropdown" />;
}
```

## Actual Behavior

When utilizing a screen reader, using the navigation buttons doesn't announce the month/year when the captionLayout is dropdown

## Expected Behavior

Using the navigation buttons should announce the month/year when the captionLayout is dropdown, just like it does when the captionLayout is label

PR #2716, opened 2025-03-15
fix: add visually hidden text to announce month/year

rodgobbi:
Fixes https://github.com/gpbl/react-day-picker/issues/2715

This PR adds a [visually hidden](https://github.com/reach/reach-ui/blob/main/packages/visually-hidden/src/reach-visually-hidden.tsx) text when the `captionLayout` is of any dropdown type. This text announces to screen readers when the month and year change, following the same pattern for the `CaptionLabel` component used in the default layout (without dropdown).

These changes fix the issue when using the previous or next month buttons.
When using the select dropdown though, the screen reader will announce the new option that gets selected, e.g. July or 2025, and will announce the month and year because the visually hidden text changes.
This in turn causes the screen reader to announce for example "July, July 2025" after selecting July with the dropdown, and after selecting a year, for example, it will announce "2025, July 2025".

### DIFF
--- a/src/DayPicker.tsx
+++ b/src/DayPicker.tsx
@@ -362,7 +362,7 @@ export function DayPicker(props: DayPickerProps) {
                           value={dateLib.getMonth(calendarMonth.date)}
                         />
                       ) : (
-                        <span role="status" aria-live="polite">
+                        <span>
                           {formatMonthDropdown(calendarMonth.date, dateLib)}
                         </span>
                       )}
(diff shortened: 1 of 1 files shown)


src/DayPicker.tsx
```diff
@@ -380,10 +380,32 @@ export function DayPicker(props: DayPickerProps) {
                           value={dateLib.getYear(calendarMonth.date)}
                         />
                       ) : (
-                        <span role="status" aria-live="polite">
+                        <span>
                           {formatYearDropdown(calendarMonth.date, dateLib)}
                         </span>
                       )}
+                      <span
```
rodgobbi, line 387: @gpbl should we extract this to a separate component to keep the `DayPicker` body shorter?

gpbl: Before, we had a “visuallyHidden” CSS class for this purpose. However, I prefer the inline style in this case for backward compatibility. 

I agree that we should extract parts of this component into subcomponents in the near future!

src/DayPicker.tsx
```diff
@@ -362,7 +362,7 @@ export function DayPicker(props: DayPickerProps) {
                           value={dateLib.getMonth(calendarMonth.date)}
                         />
                       ) : (
-                        <span role="status" aria-live="polite">
+                        <span>
```
gpbl, line 365: We should remove these elements, as they've been added only for a the live region. Thoughts?

gpbl (commented):
Looks good to me! I could try with VoiceOver and indeed this is a much better experience.

What about removing the extra `span`?

src/DayPicker.tsx
rodgobbi: Makes sense 👍 
Wanted to check what would be best for now.
We can wait for another place using the same pattern to extract to a separate component.

src/DayPicker.tsx
rodgobbi: They are still used to display text when the layout is not purely `dropdown`, like `dropdown-months` and `dropdown-years`:

image

So they are still needed to display the calendar information for users without screen reader.

gpbl (approved):

src/DayPicker.tsx
gpbl: I meant to remove the `span`, not its content -  but out of the scope of this PR!

PR #2717, opened 2025-03-17
fix: Fix importing of whole locale from date-fns

binhpv:
## What's Changed

Importing directly to the en-US locale instead of import from "date-fns/locale".
When using react-day-picker and, I found out that the bundle size increased quite a bit. After tweaking import of react-day-picker, look like this single line that import "date-fns/locale" is causing issue. When replace with a direct import to en-US, the bundle size reduced significantly, because only en-US is imported.

Note: normally one would look into https://date-fns.org/v4.1.0/docs/webpack for how to solve bundle size. However this is not necessary if you are not doing dynamic import. In my project we import the exact locale we need, and expect that only those locale will be bundle. However this was broken due to this line in react-day-picker.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/classes/DateLib.ts
+++ b/src/classes/DateLib.ts
@@ -40,7 +40,7 @@ import type {
   Interval
 } from "date-fns";
 import type { Locale } from "date-fns/locale";
-import { enUS } from "date-fns/locale";
+import { enUS } from "date-fns/locale/en-US";
 
 import { endOfBroadcastWeek } from "../helpers/endOfBroadcastWeek.js";
 import { startOfBroadcastWeek } from "../helpers/startOfBroadcastWeek.js";


rodgobbi (commented):
Thanks for the PR @binhpv 👍 

Some considerations I want to share:

`date-fns` is tree-shakable, so if your bundler setup is properly configured to tree-shake, named imports from `date-fns/locale` do not cause any issues to the bundle size.

I have confirmed this before: https://github.com/gpbl/react-day-picker/pull/2673#issuecomment-2646415726.

And I also tested your changes to be sure, the screenshots below are the output of using webpack to bundle the examples app used in the project:

`date-fns/locale` import:

date-fns:locale

---

`date-fns/locale/en-US` import:

date-fns:locale:en-US

---

Therefore, it may be that your Webpack setup is not tree-shaking `react-day-picker`.

In any case, we could merge these PR changes to help bundler setups that are not tree shaking as expected, but most of the lib is assuming tree shaking works, so this issue would happen again if importing any `locale` from the [re-exported locales](https://github.com/gpbl/react-day-picker/blob/main/src/locale.ts).

cc @gpbl

gpbl:
I agree this is a good change. Thanks @binhpv !

PR #2727, opened 2025-04-08
fix: calculateFocusTarget logic

### LINKED ISSUE #2725: `autoFocus` not focusing the selected day if after today's date
There is a small bug when using the `autoFocus` prop in the `<DayPicker />` component regarding re-focusing on the selected date. 

If the selected date comes after the current date (today) and is in the same month, the focus will target the todays date rather than the selected date. When selecting months outside of the current month, it will refocus on the current selected day.

The documentation for `autoFocus` might suggests that it will target either Today or the selected day but the behavior is not consistent with version 8 when using the the depreciated `initialFocus` prop. I was unsure if this is the intended behavior.

This might be due to the `calculateFocusTarget` [function](https://github.com/gpbl/react-day-picker/blob/main/src/helpers/calculateFocusTarget.ts#L36) will flag the `found` to `true` and exit the loop when the today day is before any of the other priorities.

## To reproduce

I created some sandboxes using the [example Dialog component](https://github.com/gpbl/react-day-picker/blob/main/examples/Dialog.tsx) using versions 8 and 9 (and React 18 and 19). 

1. Tab into opening calendar,
2. Select day later than **todays date** but in the same month,
3. Tab back into opening calendar,
4. Focus should be on Todays date (it should be on the current selected day).

- [V8 with React 18: selection correctly targets selected date](https://codesandbox.io/p/sandbox/unruffled-goldstine-jrmm4v)
- [V9 with React 19: selection focus on today](https://codesandbox.io/p/sandbox/day-picker-focus-selected-v9-4vt3z8)
- [V9 with React 18: selection focus on today](https://codesandbox.io/p/devbox/day-picker-focus-selected-v9-forked-jqrtq5)

## Actual Behavior

When the calendar opens and the prop `autoFocus` is passed into the `<DayPicker />` component, the screen will focus on todays date.

After **selecting a date after todays date but in the same month** and when the calendar re-opens, the focus is on todays date (instead of the selected date).

## Expected Behavior

When the calendar opens and the prop `autoFocus` is passed into the `<DayPicker />` component, the screen will focus on todays date.

After **selecting a date after todays date but in the same month** and when the calendar re-opens, the focus is on the selected date.

rodgobbi:
Fix #2725

## Changes

Fix the `calculateFocusTarget` logic.  
Implement a logic that iterates over all days shown in the calendar, and computes the final target day based on the priority of the valid focusable days.

### DIFF
--- a/src/helpers/calculateFocusTarget.test.ts
+++ b/src/helpers/calculateFocusTarget.test.ts
@@ -1,0 +1,167 @@
+import { DayFlag } from "../UI.js";
+import { CalendarDay } from "../classes/index.js";
+import type { Modifiers } from "../types/index.js";
+
+import { calculateFocusTarget } from "./calculateFocusTarget";
+
+const todayModifiers = {
+  [DayFlag.today]: true
+};
+
+const todayHiddenModifiers = {
+  [DayFlag.today]: true,
+  [DayFlag.hidden]: true
+};
+
+const focusedModifiers = {
+  [DayFlag.focused]: true
+};
+
+const focusedDisabledModifiers = {
+  [DayFlag.focused]: true,
+  [DayFlag.disabled]: true
+};
+
+const outsideModifiers = {
+  [DayFlag.outside]: true
+};
+
+const hiddenModifiers = {
+  [DayFlag.hidden]: true
+};
+
+const disabledModifiers = {
+  [DayFlag.disabled]: true
+};
+
+const month = new Date(2021, 0, 1);
+const day1 = new CalendarDay(new Date(2021, 0, 1), month);
+const dayToday = new CalendarDay(new Date(2021, 0, 2), month);
+const daySelected = new CalendarDay(new Date(2021, 0, 3), month);
+const dayLastFocused = new CalendarDay(new Date(2021, 0, 4), month);
+const dayFocusedModifier = new CalendarDay(new Date(2021, 0, 5), month);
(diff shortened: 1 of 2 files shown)


gpbl:
Nice work @rodgobbi! Should we add a test?

gpbl:
> > Nice work @rodgobbi! Should we add a test?
> 
> @gpbl for sure 👍 I opened a draft to check the logic before committing to write tests. Lemme know if you have any change suggestions. It looks like this is the complete fix, but I'll test it better with a calendar with multiple months, also date range, and other use cases to be safe.

Looking good to me. I wouldn't spend that much time on the test, but I'd like at least to cover the use case for #2725 in the unit test for this file.

rodgobbi:
> Looking good to me. I wouldn't spend that much time on the test, but I'd like at least to cover the use case for https://github.com/gpbl/react-day-picker/issues/2725 in the unit test for this file.

I see, thanks for the directions @gpbl, I was considering doing some integration tests with `examples` modules, but focusing on the unit tests is much faster indeed.

I also see now the helper doesn't have any tests, adding them will be valuable.

I'll finish the tests tomorrow 👍

rodgobbi:
@gpbl finished tests.

Did some manual tests with multiple months and date range also.

rodgobbi:
@gpbl  good call indeed, better explain better with the test description.

PR #2734, opened 2025-04-12
fix: fix left/right icon alignment

AlecRust:
## What's Changed

In [Chevron.tsx](https://github.com/gpbl/react-day-picker/blob/main/src/components/Chevron.tsx) different SVG code is used for each icon direction needed (rather than e.g. rotating the same icon).

The "up" vs "down" icons seem to use almost identical code, with the only changes being the direction it's pointing in.

However the "right" arrow compared with the "left" arrow has more `points` code, and it's also positioned slightly lower.

**This PR updates the `points` code of the right arrow to better match the left arrow and fix the misalignment.**

I've zoomed in to make this comparison more obvious, but I've noticed the slight misalignment of the left/right arrows since using this library at the normal zoom level.

**Before:**
before

**After:**
after

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update


gpbl:
Nice find @AlecRust ! Thanks a lot!

PR #2737, opened 2025-04-13
fix: prevent timezone override when initialMonth is Date type

lovebuizel:
## What's Changed

When month passed as Date type, it would override the provided timezone settings. This fix ensures proper timezone handling.

image
In `react-day-picker\examples\TimeZone.tsx`, when passing a Date type for the month parameter, the timezone setting was ignored:

Expected:
- First day should be: "Tue Apr 01 2025 00:00:00 GMT+1400 (Line Islands Time)"
- Current day displays correctly

Actual:
- First day shows: "Tue Apr 01 2025 00:00:00 GMT+0800 (台北標準時間)"
- Current day displays incorrectly

image

image

After fix:
image

image

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/helpers/getInitialMonth.ts
+++ b/src/helpers/getInitialMonth.ts
@@ -1,3 +1,5 @@
+import { TZDate } from "@date-fns/tz";
+
 import { type DateLib } from "../classes/DateLib.js";
 import { type DayPickerProps } from "../types/props.js";
 
@@ -23,7 +25,8 @@ export function getInitialMonth(
     today = dateLib.today(),
     numberOfMonths = 1,
     endMonth,
-    startMonth
+    startMonth,
+    timeZone
   } = props;
   let initialMonth = month || defaultMonth || today;
   const { differenceInCalendarMonths, addMonths, startOfMonth } = dateLib;
@@ -37,5 +40,8 @@ export function getInitialMonth(
   if (startMonth && differenceInCalendarMonths(initialMonth, startMonth) < 0) {
     initialMonth = startMonth;
   }
+  // When timeZone is provided, convert initialMonth to TZDate type to ensure proper timezone handling
+  initialMonth = timeZone ? new TZDate(initialMonth, timeZone) : initialMonth;
+
   return startOfMonth(initialMonth);
 }


gpbl:
Thanks @lovebuizel for your contribution! I believe the behavior of DayPicker is correct here. To pass a time-zoned date as the initial month, you should initialize the date with `TZDate`:

```tsx
import { DayPicker, TZDate } from "react-day-picker";

export function TimeZone() {
  const timeZone = "Pacific/Kiritimati";
  return (
    <DayPicker 
      // make sure you use `TZDate` here
      month={new TZDate(2024, 1, 0, timeZone)} 
    />
  );
}
```

See: https://daypicker.dev/docs/time-zone#working-with-time-zoned-dates

We might want to further clarify this in the documentation.

lovebuizel:
Thank you @gpbl for your response! I have a concern regarding the timezone configuration. I believe the timezone should be specified in only one place to avoid potential conflicts. For example, if I pass `timeZone='Pacific/Kiritimati'` as a parameter, but simultaneously provide a month parameter as a TZDate object with `timeZone='Asia/Taipei'`, it becomes unclear which timezone should take precedence.

Interestingly, I've noticed that when I don't explicitly pass the timeZone parameter but provide a month as a TZDate object with its own timezone, the timezone from the TZDate object is correctly applied. This behavior essentially renders the separate timeZone parameter redundant and potentially misleading.

This situation not only makes debugging challenging but also presents a potential issue where type checking doesn't catch the inconsistency when passing both a timezone parameter and a month parameter that isn't a TZDate type.

gpbl:
I see your point @lovebuizel and I agree with your feedback here. 

I wonder what happens with the other props, such as `selected`, where we also pass dates DayPicker 🤔 Should we apply the same treatment to them?

gpbl:
@lovebuizel, definitely your PR goes in the right direction, but without tests it's hard to say if it fixes _all the cases_ when using `Date` instead of `TZDate`.

For example, while I can confirm your fix solves the case when selecting days, it does not when rendering the month passing a simple `Date` to `month` or `defaultMonth`:

```tsx
export function TimeZone() {
  const timeZone = "Pacific/Kiritimati";
  return (
    <DayPicker
      defaultMonth={new Date(2024, 0, 0)} // Peru Time: -5 from GMT
      mode="single"
      timeZone={timeZone}
    />
  );
}
```

Here I'd expect the calendar render the January 2024, but since I'm in a timezone with `-5` from GMT hours, it still renders the previous month:

Screenshot 2025-04-14 at 6 02 47 AM

I guess you couldn't see this issue because you are at +8 from GMT.

To move forward, we need to add tests to make sure the whole calendar works with simple `Date`s objects within the specified timezone, and not just when selecting days. 

I can investigate better the issue during the next days.

gpbl (approved):

lovebuizel:
@gpbl, I agree we need to add tests for this. However, when users provide a timezone, they should be responsible for ensuring that the Date values they pass are already correctly adjusted according to the timezone they specified. This is because we cannot determine whether the Date values provided by users have already taken timezone into account or not. Therefore, we should not automatically convert the Date values to TZDate using their specified timezone.

PR #2743, opened 2025-04-19
docs: prevent propagation on input keydown

### LINKED ISSUE #2724: docs: typing "/" in input fields opens the global search
it's at website/docs/guides/input-fields.mdx file or https://daypicker.dev/guides/input-fields

## To reproduce

Just go to https://daypicker.dev/guides/input-fields and try to add manually a date you want, e.g. 12/18/2004, as the pattern suggests.

## Actual Behavior

The global search thing is triggered by "/" press, but I wanted to actually input that thing.

## Expected Behavior

I want to be able to demo the input thing with slashes, as docs suggest.

## Screenshots

Image

gpbl:
Fixes #2724 causing search appearing when typing `/`.

### DIFF
--- a/examples/Input.tsx
+++ b/examples/Input.tsx
@@ -60,6 +60,10 @@ export function Input() {
         value={inputValue}
         placeholder="MM/dd/yyyy"
         onChange={handleInputChange}
+        onKeyDown={(e) => {
+          // Fixes https://github.com/gpbl/react-day-picker/issues/2724 causing search appearing when typing '/'
+          e.stopPropagation();
+        }}
       />
       <div style={{ marginBlock: "1em" }}>
         <DayPicker


examples/Input.tsx
```diff
@@ -60,6 +60,10 @@ export function Input() {
         value={inputValue}
         placeholder="MM/dd/yyyy"
         onChange={handleInputChange}
+        onKeyDown={(e) => {
+          // Fixes https://github.com/gpbl/react-day-picker/issues/2724 causing search appearing when typing '/'
+          e.stopPropagation();
```
rodgobbi, line 65: @gpbl 

What do you think of calling `stopPropagation` only when the key typed is `/`?

This way other use cases like `cmd/ctrl + k` are not blocked.

gpbl: Oh well why should someone call that keyboard shortcut while typing into the field? I think we'd push it too far :)

gpbl: I agree, I chose the laziest path :) It seemed unnecessary to spend more time on it. I didn't even know about the shortcut itself. It looks like they addressed the bug in some release: not sure when the fix will land on Docusaurus.

> someone tries to use the shortcut while the focus is in the input

At this point, I really hope they report a bug!

PR #2750, opened 2025-04-26
fix: add timezone to `Date` props

gpbl:
In #2658, we fixed the `initialMonth` prop to account the timezone when using a `Date` object to set the initial month. In this PR, I am applying the same fix to other props that rely on `Date`.

Relevant comment about tests failing due to timezone overrides:

https://github.com/gpbl/react-day-picker/commit/814cb76b584e4be35a64d807ab0de01e99400cdd#commitcomment-155917971


rodgobbi (approved):

gpbl:
Thanks @rodgobbi !

PR #2755, opened 2025-05-04
feat: add `navLayout` prop

### LINKED ISSUE #2629: Tab order is not in visual order when captionLayout is "dropdown"
## To reproduce

```tsx
import "react-day-picker/style.css";

  return <DayPicker mode="single" captionLayout="dropdown" />;
}
```

CodeSandbox: https://codesandbox.io/p/sandbox/daypicker-bug-report-forked-mthnh4

Use tab / shift + Tab to move though the focusable elements.

## Actual Behaviour

The tabbing order does not match visual order

Tab order: Previous month, next month, Month dropdown, Year dropdown, calendar grid
Visual order: Month dropdown, Year dropdown, Previous month, next month, calendar grid

That is the previous and next months are focused before the month and year dropdowns; however visually the month and year dropdowns are displayed before the previous and next buttons.

This wasn't the case with version 8.

Possible failure of https://www.w3.org/TR/WCAG22/#focus-order.  It feels confusing if navigating with a keyboard.

## Expected Behaviour

The tabbing order should match the visual order:

## Workaround

Overriding the `<Months>` component with this monstrosity (and modifying the css) can be used to workaround this issue.

```jsx
function Months({ className, children }) {
  const [
    nav,
    {
      props: {
        children: [caption, grid],
      },
    },
  ] = Children.toArray(children);
  return (
    <div className={className}>
      {caption}
      {nav}
      {grid}
    </div>
  );
}
```

gpbl:
This PR introduces the new `navLayout` prop, which fixes #2629 and provides a new layout option for navigation:

- **`around`**: Navigation buttons are displayed on either side of the caption.
- **`after`**: Navigation buttons are displayed after the caption

Screen Shot 2025-05-04 at 08 10 34

Screen Shot 2025-05-04 at 08 22 06

This change also accounts for cases where the number of months displayed is greater than one.

Screen Shot 2025-05-04 at 08 21 54

### Known Issues

- During transitions, the navigation buttons appear duplicated. This issue can likely be resolved by modifying the `useAnimation` hook.

https://github.com/user-attachments/assets/d5faeb63-59f2-41fa-8eb8-6f699fd815b0

### Notes on Reusing the `Nav` Component

I considered adding `skipNextButton` or `skipPrevButton` props to the `Nav` component to reuse it. However, I am not sure if as having multiple `<nav>` elements with the same label is correct. Reusing `Nav` would potentially result in a breaking change.


gpbl:
@rodgobbi still refining the change, wanted to share with you this proof-of-concept - what do you think?

rodgobbi:
@gpbl 

Lemme fix the animation issues, I'll push changes to your branch.

rodgobbi:
@gpbl 

I could not repro the issue with the animation:

https://github.com/user-attachments/assets/1ce88c40-9fda-46ca-8558-290ed6ef5055

Could you share what props you used to trigger the issue?

Also, the https://github.com/gpbl/react-day-picker/issues/2629 issue still happens with the default `navLayout`. Should we keep track of it to try to fix it in the future?

rodgobbi:
@gpbl 

Sorry, you can ignore my last comment, I was able to repro it, I just need to explicitly use the `navLayout` prop equal to `after`.

rodgobbi (commented):
@gpbl 

The approach looks good indeed.
I understand better now, using explicit values for the prop fixes https://github.com/gpbl/react-day-picker/issues/2629.

And I imagine we could make the `after` value the default one in the upcoming breaking change.

My only suggestion is to try to move the new nav layout elements out of the `months.map`, as they don't repeat per month in the calendar, therefore conceptually they should be out of the `months.map` logic.
And I expect that doing so would simplify the CSS styles and perhaps fix the issue happening with the animation.

gpbl:
Thank you, @rodgobbi, for your quick review and suggestions! Apologies for not being clear about how to test this (I use the playground in the website).

> And I imagine we could make the `after` value the default one in the upcoming breaking change.

Yes, exactly! That is the plan.

> I think we could move this out of the `months.map` scope, as it is absolutely positioned. I expect we could mount one button before the map and another one after.  
> If you tried this already, let me know if there are any issues.

The reason I placed the navigation buttons inside the loop is to align them with the caption when displaying multiple months across multiple rows. I can't make assumptions about how large is that container...

Screenshot 2025-05-04 at 10 42 06 AM

Moving the buttons outside the `months` container makes it challenging to position them correctly without altering the element tree.

I believe the animation glitch is because these elements are not being hidden while duplicating the container during the transition (the test failure seems to confirm that). If you have some time to look at it I'd be happy to let you fix it!

rodgobbi:
@gpbl 

I see, I think I was testing the branch too soon, even before you pushed some changes required for the nav buttons to be aligned with the caption.
It makes sense to mount with the month element to align it properly.

For the `after` layout, I see different issues, one of which is a bug is that multiple buttons are being mounted, one for each month; we should mount only one.
The glitch also happens because of the styles manually set to the month container during the animation, which causes the `relative`  container for the nav buttons to change from the root element to the month in the calendar.

What do you think of rendering only the `after` nav buttons outside the `months.map`? After the calendar months.
I think this would fix the glitch and the current issue of multiple buttons rendered, and it also makes more sense to be outside the `map` logic.

I can test this and push the changes if it works.

gpbl:
Thanks @rodgobbi I got what you mean.

For the `around` layout - I could fix the animation glitch by changing `useAnimation` (I think I'd like to give a review to these query selectors soon or later)

> For the after layout, I see different issues, one of which is a bug is that multiple buttons are being mounted, one for each month; we should mount only one.

Fixed! 

> The glitch also happens because of the styles manually set to the month container during the animation, which causes the relative container for the nav buttons to change from the root element to the month in the calendar.

I'm not sure if I can replicate this again - I'll double check.

> What do you think of rendering only the after nav buttons outside the months.map? 

Also in this case, we want it right after the caption to preserve the tab behavior.

gpbl:
I confirm the glitch with multiple months is still there: 

https://github.com/user-attachments/assets/6202e189-944b-491c-a86a-7052080a2a6d

I will give a look to it tomorrow with a fresh mind!

rodgobbi:
@gpbl  thanks for checking, lemme try to fix it and I'll message you again with the results 👍

rodgobbi:
@gpbl
The line that is causing the issue in the animation is this [one](https://github.com/gpbl/react-day-picker/blob/main/src/useAnimation.ts#L164).

I need to think more about how to work around that. It's very tricky to support the current `after` nav layout UX with the current animation logic.

On the flip side, the glitch happens when there are multiple months wrapping to the next line, and in the same scenario the tabbing flow is also problematic:

https://github.com/user-attachments/assets/8ec2179f-b882-4c78-9fa1-e4dc481d27ce

I wasn't able to reproduce the glitch happening to the first displayed month like in https://github.com/gpbl/react-day-picker/pull/2755#issuecomment-2849474596.

What do you think of always anchoring the `after` nav buttons to the last displayed month?

Like the last nav button for the `around` layout.
This would keep the tabbing flow consistent following the visual order, and this glitch would not happen because the nav buttons' position during animation would be the same as the non-animating state.

gpbl:
> What do you think of always anchoring the after nav buttons to the last displayed month?
> Like the last nav button for the around layout.

Thanks, that's a good idea! I updated the CSS - now it seems working as expected. I'm adding some tests before I think it is ready to be merged.

Aside, the tabbing behavior with multiple months is interesting - should we focus on each month when tabbing?

gpbl:
@rodgobbi sorry, I meant a different feature - i.e. tabbing into each month - unrelated to the new nav layout. 

The tabbing may not be perfect with multiple months, but I believe with the new `after` layout we could fix the wrong tab order for most cases...

rodgobbi:
> The tabbing may not be perfect with multiple months, but I believe with the new after layout we could fix the wrong tab order for most cases...

Indeed. I think it's possible to cover all use cases easily by default, better to leave it for the developers using the day picker to customize for their needs.

@gpbl, I'll review this PR again as soon as I have time.

src/components/Nav.tsx
```diff
@@ -19,6 +19,7 @@ export function Nav(
     onNextClick?: MouseEventHandler<HTMLButtonElement>;
     previousMonth?: Date | undefined;
     nextMonth?: Date | undefined;
+    /** The component to render the previous month button. */
```
rodgobbi, line 22: Maybe this comment was left behind?

src/useAnimation.ts
```diff
@@ -210,6 +210,15 @@ export function useAnimation(
           previousWeekdaysEl.style.opacity = "0";
         }
 
+        const animatedButtons = previousMonthEl.querySelectorAll(
+          "[data-animated-button]"
```
rodgobbi, line 214: Is it possible to remove this logic due to the latest changes pushed to the branch?

If we still need, `data-animated-button` seems a bit too generic, as there are other buttons rendered by the day picker, what do you think of adding a bit more context like `data-animated-nav-button`?
Also, add the `data` attribute to all types of layouts to keep it consistent.

We can add an assertion to the `useAnimation` test (I can do that if we need to keep this logic).

gpbl:
@rodgobbi thanks again for the review. I've removed the left over code in [`a53541c` (#2755)](https://github.com/gpbl/react-day-picker/pull/2755/commits/a53541c013c58a37fcdc6911988aaf6793ef4966). To me this is ready to go!

rodgobbi (approved):

PR #2770, opened 2025-06-03
feat(accessibility): Shift+Arrows to navigate months/years

mhwice:
### What's Changed

I added a Shift modifier to the keyboard navigation. Specifically,

Shift + ArrowLeft moves the focus to the previous month.
Shift + ArrowRight moves the focus to the next month.
Shift + ArrowUp moves the focus to the previous year.
Shift + ArrowDown moves the focus to the next year.

This feels quite natural and can be useful for those without PageUp/Down buttons on their keyboards.

### DIFF
--- a/examples/Keyboard.test.tsx
+++ b/examples/Keyboard.test.tsx
@@ -78,6 +78,24 @@ describe.each(["ltr", "rtl"])("when text direction is %s", (dir: string) => {
         });
       }
     });
+    describe("when the Shift + Arrow Left is pressed", () => {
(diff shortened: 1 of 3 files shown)


rodgobbi:
Sorry @gpbl, it took me a while to see this PR.
Looks good indeed!

gpbl:
> Sorry @gpbl, it took me a while to see this PR. Looks good indeed!

Ah thanks for checking it out! I merged it assuming you were busy :)

PR #2777, opened 2025-06-12
docs(translation): Fix wrong import on example

gonzalosalmeron:
# Pull Request

Thank you for the great library! This PR fixes a naming conflict in the code examples.

## What's Changed

- Resolved variable naming conflicts in documentation examples when importing locales from both `react-day-picker/locale` and `date-fns/locale`.

## Type of Change

- [x] Documentation update

## Additional Notes

This change improves clarity for developers using both libraries with localized calendars.

### DIFF
--- a/website/docs/docs/translation.mdx
+++ b/website/docs/docs/translation.mdx
@@ -30,7 +30,7 @@ The `labels` prop allows you to customize the [ARIA labels](../guides/accessibil
 
 ```tsx
 import { format } from "date-fns";
-import { it } from "react-day-picker/locale";
+import { it } from "date-fns/locale";
 
 function ItalianLabels() {
   return (
@@ -88,7 +88,7 @@ The following labels are optional and should work out of the box in most languag
 To set the text direction to right-to-left, use the `dir` prop with the value `rtl`.
 
 ```tsx
-import { arSA } from "react-day-picker/locale";
+import { arSA } from "date-fns/locale";
 
 <DayPicker locale={arSA} dir="rtl" />;
 ```


rodgobbi (commented):
Hi @gonzalosalmeron
Thanks for the PR 👍 
But how does this improve documentation?
Removing will hide from developers that the `react-day-picker` re-exports `date-fns/locale` for convenience.

gonzalosalmeron:
Hi @gpbl @rodgobbi, thanks for the clarification! 🙏

I totally understand the intention of re-exporting date-fns/locale for convenience — however, after checking the published package, it seems the re-export isn’t actually working as expected.

I couldn’t find date-fns/locale included anywhere in the package (neither in the build output nor in the exports). I’ve attached some screenshots to show what I found — so even though the export was added in #2474 , it doesn’t appear to be accessible to consumers of the package right now.

Happy to help further or suggest a fix if needed! 🚀

image
image
image

gpbl:
Thanks for checking this. I verified the published v9.1.1+ and v10 packages expose `react-day-picker/locale`, including `dist/esm/locale.js`, `dist/cjs/locale.js`, and per-locale files.

The screenshots appear to come from an older v8 package layout (`dist/index.js`). In v9/v10, locales should be imported from `react-day-picker/locale` because they include DayPicker-specific labels in addition to date-fns locale data.

If an app needs both imports, aliasing is the better fix:

```tsx
import { it as dayPickerIt } from "react-day-picker/locale";
import { it as dateFnsIt } from "date-fns/locale";
```

Since the docs have also been reorganized on main, this PR is now stale, so I’m going to close it.

Issue #2776, opened 2025-06-12
Setting a `defaultMonth` the next year when using the `dropdown` mode does not render the calendar.


drac94:
Hey, not sure if this is a bug or a feature but setting a default month in the future (next year) when using the dropdown mode results in the calendar not being rendered, to solve it I need to add and end month but not sure if this is the correct way.

## To reproduce

- Just add a component with a default month the year after you are right now

```tsx
import { DayPicker } from "react-day-picker";
import "react-day-picker/style.css";

export default function App() {
  return  (
    <DayPicker
        hideNavigation
        captionLayout="dropdown"
        defaultMonth={new Date("2026/07/05")}
      />
  )
}
```

CodeSandbox: https://codesandbox.io/p/sandbox/4zg6k5

## Actual Behavior

Calendar is not rendered

## Expected Behavior

Calendar should be rendered.

PR #2783, opened 2025-06-29
fix: setting defaultMonth the next year when using the dropdown mode does not render the calendar

rodgobbi:
Fix https://github.com/gpbl/react-day-picker/issues/2776

## Description

The issue happens because the valid months of the calendar are calculated internally in the [`useCalendar` hook](https://github.com/gpbl/react-day-picker/blob/main/src/useCalendar.ts#L101) when the `layout` of the day picker is a `dropdown` type, but other inner logic parts of the day picker use the `startMonth` and `endMonth` props as the calendar limits, which in this use case is `undefined`.

## Changes

This PR unifies the source of truth for the start and end months of the day picker, using the `navStart` and `navEnd` returned by the [`getNavMonths` in the `useCalendar` hook](https://github.com/gpbl/react-day-picker/blob/main/src/useCalendar.ts#L101) as the source of truth.
This also means that from now on, we should **NOT** use the `startMonth` and `endMonth` props directly, and instead use the `navStart` and `navEnd` values.

### DIFF
--- a/examples/InvalidMonthDefaultLimit.test.tsx
+++ b/examples/InvalidMonthDefaultLimit.test.tsx
@@ -1,0 +1,11 @@
+import React from "react";
+
+import { grid } from "@/test/elements";
+import { render } from "@/test/render";
+
+import { InvalidMonthDefaultLimit } from "./InvalidMonthDefaultLimit";
+
+test("should display calendar in December", () => {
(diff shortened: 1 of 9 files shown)


gpbl (approved):
Well done @rodgobbi, thanks!

PR #2803, opened 2025-07-27
Move from eslint/prettier to biomejs

gpbl:
In this PR, I'm migrating from ESLint/Prettier to [Biome](http://biomejs.dev) for linting and formatting the codebase.

### Motivation

Upgrading to ESLint 9 has been painful, and I wanted a simpler, faster setup with fewer dependencies. This was also a good opportunity to explore newer dev tools.

### Pros

- Fast and sensible defaults
- No need for plugins or extra dependencies
- Simple config and setup
- Good documentation

### Cons

- Not as widely adopted as ESLint/Prettier
- Doesn’t support formatting JSDoc: https://github.com/biomejs/biome/discussions/741
- GitHub annotations appear duplicated (might be related to the workflow configuration)

### Changes

- Removed ESLint and Prettier dependencies, added Biome
- Applied formatting and linting to `src` and `example-app` (most diffs are for trailing commas)
- Used `@biome-ignore` comments for breaking changes

The `website` package is excluded in this PR to keep the diff manageable; it will be updated in a follow-up.


rodgobbi (approved):
@gpbl 
awesome, looks great!
So, the experience I had with `biomejs` is that it's very limited and opinionated compared to `eslint`, but if it covers all our use cases, it feels much better with how much faster and leaner it is.

gpbl:
Thanks @rodgobbi for the review! Yeah, I don't feel ESLint fits anymore for a small repo with few maintainers. It's too much work for something that should be invisible and require minimal maintenance. Let's see how Biome works. I'm all for supporting smaller projects like that!

package.json
```diff
@@ -165,7 +165,7 @@
     "build:cjs": "tsc --project tsconfig-cjs.json && echo '{ \"type\": \"commonjs\" }' > dist/cjs/package.json",
     "build:esm": "tsc --project tsconfig-esm.json",
     "build:css": "./scripts/build-css.sh ./src/style.css ./src/style.module.css",
-    "lint": "eslint .",
+    "lint": "biome check",
```
rodgobbi, line 168: Would it help to add a separate `biome check --write` script?

The same can be achieved with `pnpm run lint --write`

gpbl: Yeah I prefer adding the `--write` when necessary...

PR #2811, opened 2025-07-31
repo: add prettier dependency and update biome config

gpbl:
Biome.js, added in #2803, doesn't support formatting Markdown documents or JSDocs:  
https://github.com/biomejs/biome/issues/3718  
https://github.com/biomejs/biome/discussions/741

In this PR, I'm updating the repo to use Prettier for those cases.  
Not ideal, but at least we can format them using VSCode or the `pnpm format` script.

I also fixed some wrong `// biome` comments, and tweaked the config to be more inclusive.


rodgobbi (dismissed):
So sad we can't use just biome 😢 
I have issues with VS Code if I try to have both enabled and have the editor format the file for me.

PR #2814, opened 2025-08-03
feat: Update year and month dropdown order based on locale

chuganzy:
# Pull Request Template

## What's Changed

Currently dropdown order is hardcoded, however the ideal order changes based on locale.
This PR makes use of `Intl.DateTimeFormat.prototype.formatToParts()` API to get the best matching order.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Tips for a good PR

Thanks

## Additional Notes

Add any extra comments or questions here.

### DIFF
--- a/examples/Dropdown.test.tsx
+++ b/examples/Dropdown.test.tsx
@@ -1,3 +1,4 @@
+import { ja } from "date-fns/locale";
 import React from "react";
 
 import { grid, monthDropdown, yearDropdown } from "@/test/elements";
@@ -11,19 +12,29 @@ const today = new Date(2015, 6, 1);
 beforeAll(() => jest.setSystemTime(today));
 afterAll(() => jest.useRealTimers());
 
-beforeEach(() => {
-  render(<Dropdown />);
-});
-
 test("should display the month dropdown", () => {
+  render(<Dropdown />);
   expect(monthDropdown()).toBeInTheDocument();
 });
 
 test("should display the year dropdown", () => {
+  render(<Dropdown />);
   expect(yearDropdown()).toBeInTheDocument();
 });
 
+test("should display the month and year dropdowns in the correct order", () => {
+  const { rerender } = render(<Dropdown />);
+  const comparePosition = () =>
+    monthDropdown().compareDocumentPosition(yearDropdown());
+
+  expect(comparePosition()).toBe(Node.DOCUMENT_POSITION_FOLLOWING);
+
+  rerender(<Dropdown locale={ja} />);
+  expect(comparePosition()).toBe(Node.DOCUMENT_POSITION_PRECEDING);
+});
+
 test("should disable the months before startMonth", () => {
+  render(<Dropdown />);
   const disabledMonth = [
     "January",
     "February",
@@ -40,6 +51,7 @@ test("should disable the months before startMonth", () => {
 });
 
 test("should disable the months after endMonth", async () => {
+  render(<Dropdown />);
   await user.selectOptions(yearDropdown(), "2025");
   const disabledMonth = ["November", "December"];
   for (const month of disabledMonth) {
@@ -51,6 +63,7 @@ test("should disable the months after endMonth", async () => {
 
 describe("when choosing a month", () => {
   test("should display the month", async () => {
+    render(<Dropdown />);
     const monthName = "December";
     await user.selectOptions(monthDropdown(), monthName);
     expect(grid()).toHaveAccessibleName(`${monthName} 2024`);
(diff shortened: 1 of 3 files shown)


src/DayPicker.tsx
```diff
@@ -425,42 +437,52 @@ export function DayPicker(initialProps: DayPickerProps) {
...
-                          options={dropdownYears}
-                          style={styles?.[UI.Dropdown]}
-                          value={dateLib.getYear(calendarMonth.date)}
-                        />
-                      ) : (
-                        <span>
-                          {formatYearDropdown(calendarMonth.date, dateLib)}
-                        </span>
-                      )}
+                      {dropdownTypes.map((type) => {
```
rodgobbi, line 440: This new feature will also adapt the month and year labels, not just the `dropdowns`.
Could you rename `dropdownTypes` to a more generic name that refers to the date-time format of the `locale`?

src/DayPicker.tsx
```diff
@@ -425,42 +437,52 @@ export function DayPicker(initialProps: DayPickerProps) {
...
+                        const uiKey =
```
rodgobbi, line 441: For me, it's better to be explicit, as in the previous version.
Could you refactor this to have `if` statements that return the component with the explicit props applied to it, like:
```tsx
if (type === month) {
  return // return the same code block as before for month
}
if (type === year) {
  return // return the same code block as before for year
}
```

src/DayPicker.tsx
```diff
@@ -425,42 +437,52 @@ export function DayPicker(initialProps: DayPickerProps) {
...
+                          type === "month" ? dropdownMonths : dropdownYears;
+                        const valueGetter =
+                          type === "month" ? dateLib.getMonth : dateLib.getYear;
+                        const formatLabel =
+                          type === "month"
+                            ? formatMonthDropdown
+                            : formatYearDropdown;
+
+                        return (
+                          <Fragment key={type}>
```
rodgobbi, line 462: `Fragment` is not needed, `key` can be passed to the `Dropdown` component, and `span`.

examples/Dropdown.test.tsx
```diff
@@ -11,19 +12,29 @@ const today = new Date(2015, 6, 1);
...
 
+test("should display the month and year dropdowns in the correct order", () => {
+  const { rerender } = render(<Dropdown />);
+  const comparePosition = () =>
+    monthDropdown().compareDocumentPosition(yearDropdown());
+
+  expect(comparePosition()).toBe(Node.DOCUMENT_POSITION_FOLLOWING);
+
+  rerender(<Dropdown locale={ja} />);
+  expect(comparePosition()).toBe(Node.DOCUMENT_POSITION_PRECEDING);
```
rodgobbi, line 33: This new feature will also change the way date text is displayed, so we can assert that the text is rendered in the right order using `toHaveAccessibleName`, like the other tests.
But to do so, we also need to update the accessible name of the calendar, could you update the [`labelGrid`](https://github.com/gpbl/react-day-picker/blob/main/src/labels/labelGrid.ts) to have the same dynamic logic based on `locale`?
Please extract the logic to compute the month/year order based on `locale` to a separate function.

rodgobbi (changes requested):
@chuganzy thanks for the PR!
The idea is great 🚀
I left some change requests.

Could you also improve the PR description?
Adding more context and if possible, screenshots of how the day picker would look for you.

gpbl:
This is an interesting take. I have some questions:

- Which locales require displaying the year before the month? Does the test cover these locales?
- This sounds like a breaking change. Should we add a prop to toggle this feature instead?
- Why apply this only to the dropdown and not also to the caption?

chuganzy:
Apologies for leaving this open for a while..

> Which locales require displaying the year before the month? Does the test cover these locales?

Many Asian countries such as Japan, China, Korea and more, and yes I added a test that covers this.

> This sounds like a breaking change. Should we add a prop to toggle this feature instead?

Let me come up with the way.

> Why apply this only to the dropdown and not also to the caption?

Because for captions and labels we already have a way to override, and it is necessary for these locales to override these to properly localize them, so I initially thought it can be decoupled, but indeed I can come up with the way to take care of it along with the previous question.

gpbl:
@chuganzy Thanks so much for your insights here!

I see that the issue isn’t just about the drop-downs, but also affects the formatting of the caption in [formatCaption](https://github.com/gpbl/react-day-picker/blob/329fe4ad08d85ea06f017f0d464b9962ba937d18/src/labels/labelGrid.ts#L7) and the grid label in [labelGrid](https://github.com/gpbl/react-day-picker/blob/329fe4ad08d85ea06f017f0d464b9962ba937d18/src/labels/labelGrid.ts).

I think it makes sense to take a broader approach and add a function to [DateLib](https://github.com/gpbl/react-day-picker/blob/329fe4ad08d85ea06f017f0d464b9962ba937d18/src/classes/DateLib.ts#L210) that tells us whether the calendar should display year-first or month-first.

Since this requires some familiarity with the codebase, I’m happy to take this on.

chuganzy:
Great, then I will close mine! Thanks!

gpbl:
@chuganzy the change is available in v9.11 - see for example 

https://daypicker.dev/playground?locale=zh-CN

does it work for you?

chuganzy:
Thanks, and it looks good! Technically, year format is not 100% accurate still (can see the difference between Intl format vs `date-fns` when using label caption vs dropdown caption), but indeed that can be decoupled!

So, looks like no flag is added and was just released as a minor update?

gpbl:
Oh thanks, @chuganzy! I thought the kanji 年 only applied to the year when displayed with the month. Sure, I can update the format so it’s also included when the year is shown alone. Thanks for the heads up!

PR #2833, opened 2025-09-26
fix: timezone-matcher-support

lovebuizel:
## What's Changed

- Add convertMatcherToTimeZone utility for timezone-aware date conversions
- Integrate timezone conversion for disabled, hidden, and custom modifiers
- Ensure consistent date comparisons across different timezones

### Problem Description:
When using a timezone that converts today to a different day, the disabled and hidden properties do not update according to the timezone conversion. For example, when today is September 26th in Taipei timezone and I convert to Hawaii timezone (which becomes September 25th), the September 25th date on the DayPicker still remains disabled, but I expect that day should not be disabled.

image

```
import React, { useState } from "react";

import { DayPicker, TZDate } from "react-day-picker";

export function TimeZone() {
  const timeZone = "Pacific/Honolulu";
  const [selected, setSelected] = useState<Date | undefined>(
    TZDate.tz(timeZone),
  );
  const now = new Date();
  return (
    <div>
      <h3>{TZDate.tz("Asia/Taipei", now).toString()}</h3>
      <h3>{TZDate.tz(timeZone, now).toString()}</h3>
      <DayPicker
        mode="single"
        timeZone={timeZone}
        selected={selected}
        onSelect={setSelected}
        disabled={{
          before: now,
        }}
        footer={
          selected
            ? selected.toString()
            : `Pick a day to see it is in ${timeZone} time zone.`
        }
      />
    </div>
  );
}
```

### Solution:
Implemented timezone-aware conversion for disabled, hidden, and custom modifiers properties in DayPicker. The fix ensures that all date matchers are properly converted to the specified timezone before being applied, maintaining consistent date comparisons across different timezones.

image

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/DayPicker.tsx
+++ b/src/DayPicker.tsx
@@ -30,6 +30,7 @@ import { useFocus } from "./useFocus.js";
 import { useSelection } from "./useSelection.js";
 import { rangeIncludesDate } from "./utils/rangeIncludesDate.js";
 import { isDateRange } from "./utils/typeguards.js";
+import { convertMatchersToTimeZone } from "./utils/convertMatcherToTimeZone.js";
 
 /**
  * Renders the DayPicker calendar component.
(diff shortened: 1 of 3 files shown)


lovebuizel:
Thanks @gpbl for the feedback, I understand! Looking forward to your implementation.

PR #2896, opened 2026-02-07
feat: add vazirmatn font for persian example

rezaaa:
### Summary
- Add the [Vazirmatn](https://fonts.google.com/specimen/Vazirmatn) font (a popular Persian/Arabic font) to the Persian calendar example for proper RTL typography
- Load the Vazirmatn font via Google Fonts in the Docusaurus config
- Wrap the Persian `DayPicker` example in a `div` with `fontFamily: "Vazirmatn, sans-serif"`

New Standard Font:

image

Old Font:

image

### Changed files
- `examples/Persian.tsx` — Wrap `DayPicker` in a div with Vazirmatn font family
- `website/docusaurus.config.ts` — Add Vazirmatn Google Fonts stylesheet

Thanks


rezaaa:
@gpbl That makes sense 👍 I agree it should be tied to the locale rather than applied manually.
I look into it.

rezaaa:
@gpbl Applied font globally to all RTL (right-to-left) components across the examples app, and shadow DOM components.

PR #2977, opened 2026-05-02
fix: correct dropdown navigation in multi-month calendars

### LINKED ISSUE #2741: Dropdown should change the month where it is displayed
Dropdown should change the month where it is displayed, while now it changes the DayPicker starting month. 

### Discussed in https://github.com/gpbl/react-day-picker/discussions/2740

<div type='discussions-op-text'>

<sup>Originally posted by **duc-q** April 16, 2025</sup>

- When selecting a Month in the "To" or 2nd Month dropdown, the new month will be assigned to the "From" or 1st Calendar. 
Selecting "June"

image

Have a look at the left calendar as June is selected.

image

Is this a bug or it's intended like that? Should the new month affect only on the selected calendar and adjust the other calendar accordingly?

</div>

hackgray47-eng:
When using captionLayout="dropdown" with numberOfMonths > 1, selecting a month or year from the Nth calendar's dropdown was incorrectly setting firstMonth to the selected value directly, causing the first calendar to jump instead of the intended one. Offset the target by -displayIndex before calling goToMonth so the selected calendar shows the chosen month.

Fixes #2741

### DIFF
--- a/.changeset/multi-month-dropdown-navigation.md
+++ b/.changeset/multi-month-dropdown-navigation.md
@@ -1,0 +1,6 @@
+---
+"react-day-picker": patch
+"@daypicker/react": patch
+---
+
+fix: keep multi-month dropdown selections on the edited calendar.
--- a/packages/react-day-picker/src/DayPicker.test.tsx
+++ b/packages/react-day-picker/src/DayPicker.test.tsx
@@ -472,6 +472,50 @@ test("places the year dropdown before the month dropdown for year-first locales"
   );
 });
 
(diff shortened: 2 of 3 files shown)


gpbl:
Thanks for the fix @hackgray47-eng.

I added coverage for `reverseMonths`: it uses the chronological month offset instead of the visual `displayIndex`.

gpbl (approved):

hackgray47-eng:
> Thanks for the fix @hackgray47-eng.
> 
> I added coverage for `reverseMonths`: it uses the chronological month offset instead of the visual `displayIndex`.

Thank you.

PR #2997, opened 2026-05-11
fix: add `@types/react` as optional peer dependency

### LINKED ISSUE #2994: `@types/react` needs to be listed as peer dependency
## To reproduce

This package depends on types from React, but doesn't declare `@types/react` as a dependency or peer dependency.

Example:
https://github.com/gpbl/react-day-picker/blob/839f270f039cbe2ac2450ff9bdef26f08771b770/packages/react-day-picker/src/types/shared.ts#L1

This workaround (with a lazy types version) solves the issue for now:
```yaml
# pnpm-workspace.yaml
packageExtensions:
  react-day-picker:
    peerDependencies:
      "@types/react": "*"
    peerDependenciesMeta:
      "@types/react": { optional: true }
```

## Actual Behavior

In pnpm's stricter modes (i.e. `enableGlobalVirtualStore`), TypeScript can't resolve the type.

The underlying issue is that in this mode, pnpm doesn't symlink the the types package, as it's not declared as a dependency. This is likely also an issue in with Yarn Berry.

## Expected Behavior

The package should declare this relationship, allowing pnpm to link this.

**I'm happy to create a PR for this fix, and to mark it as an optional peer dependency.**

## Screenshots

Error in VS Code:
Image

With the workaround applied:
Image

mrmckeb:
This PR adds `@types/react` as an optional peer dependency, satisfying the requirements of stricter package manager configurations.

Closes #2994

### DIFF
--- a/.changeset/public-moles-knock.md
+++ b/.changeset/public-moles-knock.md
@@ -1,0 +1,11 @@
+---
+"@daypicker/buddhist": patch
+"@daypicker/ethiopic": patch
+"@daypicker/hebrew": patch
+"@daypicker/hijri": patch
+"@daypicker/persian": patch
+"@daypicker/react": patch
+"react-day-picker": patch
+---
+
+Added `@types/react` as an optional peer dependency to fix type resolution under strict package managers.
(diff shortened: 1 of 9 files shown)


gpbl (approved):

gpbl:
Thanks @mrmckeb !

Issue #2690, opened 2025-02-28
The `selected` prop is not working without setting `onSelect`


jiwon79:
## To reproduce
**stackblitz Link**: https://stackblitz.com/edit/vitejs-vite-rdyvfyjg?file=src%2FApp.tsx

```tsx
import { useState } from 'react';
import { DayPicker } from 'react-day-picker';
import 'react-day-picker/style.css';

const DAY_IN_MS = 24 * 60 * 60 * 1000;

function App() {
  const [range, setRange] = useState({
    from: new Date(),
    to: new Date(),
  });

  const changerPrev = () => {
    setRange({
      from: new Date(range.from.getTime() - DAY_IN_MS),
      to: range.to,
    });
  };

  return (
    <>
      <button onClick={() => changerPrev()}>prev</button>
      {/* it doesn't work */}
      <DayPicker mode="range" selected={range} />
      {/* it work */}
      <DayPicker mode="range" selected={range} onSelect={() => void 0} />
    </>
  );
}

export default App;
```

## Actual Behavior

In the first case, the selected value does not change.
In the second case, the selected value changes.

## Expected Behavior

In both the first and second cases, the selected value should change.

## Screenshots

<video src="https://github.com/user-attachments/assets/46eefca4-d775-4b41-9418-6992e979b5e7" />

rodgobbi:
Hi @jiwon79 
Thanks for the feedback!

@gpbl 
I agree with @jiwon79, I noticed this behavior and was planning to discuss at some point.
Internally the `DayPicker` only uses the `selected` prop if the `onSelect` is defined.
IMO, it's not ideal because developers may want to use other ways to control the `selected` date.
This GH issue has one example, and another one that comes to my mind is using the `onDayClick` event handler instead of `onSelect` to compute the `selected` value.

What do you think of removing this behavior and only using the `selected` prop if it's defined?

The issue with this change of behavior is that I think it's a breaking change, isn't it?

jiwon79:
Yes, it’s a BC, but I think we have to removing this behavior.

In all input-like components, component is uncontrolled when value is undefined. The current `DayPicker` range mode's behavior is hard to predict for someone using it for the first time.

devdpontes:
Also bumped into this issue. I'm migrating from v7 to v9. Couldn't figure out why the selected prop wasn't working as before.

gpbl:
Thanks for the feedback @devdpontes. 

[...]

PR #2998, opened 2026-05-12
feat: add defaultSelected selection prop

gpbl:
Adds `defaultSelected` so apps can initialize uncontrolled selections without relying on `selected` as initial state. Keeps existing `selected` behavior intact for v10 compatibility; read-only controlled selections can use `selected` with a no-op `onSelect`.

## What Changed

- Added `defaultSelected` to single, multiple, and range selection prop types.
- Initialized uncontrolled selection hooks from `defaultSelected`, falling back to existing `selected` behavior when omitted.
- Normalized `defaultSelected` through the existing `timeZone` conversion path.
- Updated selection docs, focused tests, and changeset coverage.

### DIFF
--- a/.changeset/gentle-ranges-start.md
+++ b/.changeset/gentle-ranges-start.md
@@ -1,0 +1,5 @@
+---
+"react-day-picker": minor
+---
+
+Add `defaultSelected` to initialize uncontrolled single, multiple, and range selections.
--- a/apps/website/docs/selections/multiple-mode.mdx
+++ b/apps/website/docs/selections/multiple-mode.mdx
@@ -16,13 +16,16 @@ Set the `mode` prop to `"multiple"` to enable the selection of multiple dates in
 
 ## Multiple Mode Props
 
-| Prop Name  | Type                                                                             | Description                                       |
-| ---------- | -------------------------------------------------------------------------------- | ------------------------------------------------- |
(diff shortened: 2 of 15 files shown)


rodgobbi (commented):
@gpbl IMO uncontrolled mode is not useful for `DayPicker`.
The consumers want to have access to the selected dates to use for something else.

Using `DayPicker` uncontrolled is just a calendar day picker that changes state, but that state is not used for anything else.

gpbl:
> @gpbl IMO uncontrolled mode is not useful for `DayPicker`. The consumers want to have access to the selected dates to use for something else.
> 
> Using `DayPicker` uncontrolled is just a calendar day picker that changes state, but that state is not used for anything else.

Fair point, and sorry for not providing enough context in this draft PR. The goal is to introduce the `defaultSelected` prop so that, in a future major release, we can address https://github.com/gpbl/react-day-picker/issues/2690.

I am still exploring the right solution here, so any feedback is very welcome!

rodgobbi:
@gpbl I see, I also revisited the discussion in the issue.
Keep in mind the existing issue with `undefined` representing an empty selection.

What do you think of trying this together with [another prop](https://github.com/gpbl/react-day-picker/issues/2690#issuecomment-2783882848) that changes the empty selection from `undefined` to `null`?

PR #3004, opened 2026-07-19
fix: clamp displayed month when endMonth changes

wanxiankai:
## Summary
- Clamp the internally displayed month when navigation bounds change after mount.
- Keep dropdown captions and the month grid visible when a new endMonth falls before the current month.
- Add a regression test for issue #2912.

Fixes #2912

## Tests
- pnpm exec jest packages/react-day-picker/src/DayPicker.test.tsx --selectProjects src --runInBand --no-watchman
- pnpm run lint
- pnpm run typecheck
- TZ=UTC pnpm exec jest --selectProjects examples --selectProjects src --selectProjects packages --selectProjects scripts --runInBand --no-watchman

### DIFF
--- a/packages/react-day-picker/src/DayPicker.test.tsx
+++ b/packages/react-day-picker/src/DayPicker.test.tsx
@@ -533,6 +533,34 @@ describe("when the `month` is changed programmatically", () => {
   });
 });
 
+test("clamps the displayed month when endMonth changes before the current month", () => {
+  const { rerender } = render(
+    <DayPicker
+      captionLayout="dropdown"
+      defaultMonth={new Date(2024, 8, 1)}
+      endMonth={new Date(2024, 11, 1)}
+    />,
+  );
+
+  expect(grid("September 2024")).toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


rodgobbi (changes requested):
Hi @wanxiankai , thanks for the contribution! 🙏 

But it's better to avoid `useEffect`, it's a common source of bugs with `React`.

Thanks to your PR I thought of a solution that doesn't rely on `useEffect`.
I'll prepare a PR to fix the issue and keep you posted. 👍

gpbl:
Do we actually need to change this? Can't the date be controlled upstream? 🤔

rodgobbi:
> Do we actually need to change this? Can't the date be controlled upstream? 🤔

@gpbl but we would be forcing consumers to always control the month, and developer errors can still put the calendar in an unusable state.

I'm following your idea here https://github.com/gpbl/react-day-picker/issues/2671#issuecomment-2613980971 to always make sure the calendar is in a usable state.

PR #3005, opened 2026-07-26
fix: changing `startMonth` and `endMonth` makes the calendar disappear

rodgobbi:
fix #2912 

### Description

The logic has gaps where invalid state, e.g. uncontrolled displayed month state ends outside the valid `startMonth` and `endMonth` range when they change through props, results in the calendar being unusable.
In the reported issue #2912, it disappears.

### What's changed

This PR changes the logic to always calculate the displayed first month in the calendar consiring all edge cases and constraints.

Currently, for uncontrolled "selected month" state, it only calculates a valid first month on mount.
This PR calculates it in every render.

### Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update


packages/react-day-picker/src/useCalendar.ts
```diff
@@ -95,24 +95,27 @@ export function useCalendar(
...
   const [firstMonth, setFirstMonth] = useControlledValue(
-    initialMonth,
-    // initialMonth is always computed from props.month if provided
-    props.month ? initialMonth : undefined,
+    props.defaultMonth || today,
+    props.month || undefined,
   );
 
-  // biome-ignore lint/correctness/useExhaustiveDependencies: change the initial month when the time zone changes.
-  useEffect(() => {
```
rodgobbi, line 106: @gpbl  removed this `useEffect`, lemme if how to verify if it's still needed.

gpbl: This effect is still needed :) I added a test to cover the use of the `useEffect`, and rebased your PR to confirm that removing it breaks an existing feature.

gpbl (commented):
Thanks for working on this! Sorry for the delay here...

#3006 addresses the same issue, reusing `getInitialMonth`: should we close this PR in favor of #3006?

rodgobbi:
> Thanks for working on this! Sorry for the delay here...
> 
> #3006 addresses the same issue, reusing `getInitialMonth`: should we close this PR in favor of #3006?

Indeed, we can close #3006.
This PR (#3005) solves the same issue.

> This effect is still needed :) I added a test to cover the use of the useEffect, and rebased your PR to confirm that removing it breaks an existing feature.

Thanks for checking 👍 
I added the `useEffect` back and updated to the new logic and functions.

It's ready for review again. 👌

PR #3007, opened 2026-08-18
fix: import date-fns locales from subpaths

huuyafwww:
Thanks for maintaining DayPicker — we use it in an SSR app and traced part of our startup cost to the translated locale wrappers.

Each wrapper currently imports from `date-fns/locale`. That works after tree-shaking, but Node and SSR setups that keep DayPicker external evaluate the whole date-fns locale barrel. Importing the matching date-fns subpath keeps the exported locale and translated labels unchanged without loading that barrel.

With a fresh ESM build loaded directly by Node, importing `react-day-picker` went from 943 to 421 loaded files. Importing `react-day-picker/locale/ja` went from 841 to 325. The date-fns locale barrel is no longer resolved in either path.

This applies the same direct-locale pattern as #2717 to all 95 wrappers, adds a scoped lint guard, and leaves the public `react-day-picker/locale` barrel unchanged. The existing tests, timezone test, typecheck, build, built-package tests, and package dry run pass. One caveat: the built-package tests hit the existing 1880/1879 boundary assertion under my local Asia/Tokyo timezone; all 311 pass with `TZ=UTC`.

Would you be open to this change? Happy to adjust the shape if you would prefer a smaller guard.

### DIFF
--- a/.changeset/fix-locale-subpath-imports.md
+++ b/.changeset/fix-locale-subpath-imports.md
@@ -1,0 +1,6 @@
+---
+"react-day-picker": patch
+"@daypicker/react": patch
+---
+
+fix: avoid loading all date-fns locales when importing a locale subpath.
--- a/biome.json
+++ b/biome.json
@@ -40,6 +40,23 @@
   },
   "overrides": [
     {
+      "includes": ["packages/react-day-picker/src/locale/*.ts"],
+      "linter": {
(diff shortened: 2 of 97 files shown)


gpbl:
Thanks for the detailed PR @huuyafwww! LGTM!

gpbl (approved):

PR #3012, opened 2026-08-26
Keep focus in the grid when a controlled month change removes the focused day

### LINKED ISSUE #3009: Focus is lost when a controlled month change removes the focused day
In `react-day-picker` 10.0.1, changing the controlled `month` while a day has focus removes that day from the DOM and leaves `document.activeElement` on `<body>`. This breaks keyboard continuity for controlled calendars whose value can change externally, such as after undoing a date selection.

This also occurs on the current `main` branch. It was found while working on [WordPress/gutenberg#81635](https://github.com/WordPress/gutenberg/pull/81635).

## To reproduce

```tsx
import { useState } from "react";
import "react-day-picker/style.css";

const march = new Date(2026, 2, 15);
const november = new Date(2026, 10, 15);

  const [month, setMonth] = useState(march);
  const [selected, setSelected] = useState<Date | undefined>(march);

  return (
    <div
      onKeyDown={(event) => {
        if (event.key.toLowerCase() === "j") {
          setMonth(november);
          setSelected(november);
        }
      }}
    >
      <p>Focus a day in the calendar, then press J.</p>
      <DayPicker
        mode="single"
        month={month}
        selected={selected}
        onMonthChange={setMonth}
        onSelect={setSelected}
      />
    </div>
  );
}
```

1. Use Tab to focus a day in March.
2. Press <kbd>J</kbd> to update the controlled month and selection to November.
3. Inspect the focus position or press an arrow key.

## Actual Behavior

November is displayed and November 15 is selected, but focus moves to `<body>`. Arrow keys no longer move between days.

The focus state in `useFocus` still refers to the day from the previous month. When that day is removed from the DOM, the browser has no replacement focus target.

## Expected Behavior

When a controlled month change removes the focused day, DayPicker should keep focus on an enabled day in the newly displayed month. The selected day is the natural target when it is enabled. DayPicker should not move focus when focus was outside the day grid before the update.

## Screenshots

Not applicable.

raphyabak:
## Summary

Fixes #3009.

When the `month` prop changes under external control (e.g. an app updates both `month` and `selected` together, such as after undoing a date selection), the previously focused day can be removed from the DOM entirely. `useFocus`'s `focusedDay` state kept referring to that removed day, so no day had the `focused` modifier after the update, `DayButton` never called `.focus()` on anything, and focus silently fell back to `<body>` — breaking keyboard continuity (arrow keys stop moving between days) for any controlled calendar whose value can change externally.

```tsx
// Focus a day in March, then press J:
setMonth(november);
setSelected(november);
// November is displayed and November 15 is selected, but
// document.activeElement is <body>, and arrow keys do nothing.
```

## Fix

Added an effect in `useFocus` that detects when `focusedDay` is no longer among the currently displayed `calendar.days` and moves focus to `calculateFocusTarget`'s result instead — the same helper already used for `autoFocus`, which prioritizes the selected day, then today, then the first focusable day. Focus is left untouched when it wasn't in the grid to begin with, per the issue's expected behavior.

## Test plan

- Added two tests under `DayPicker.test.tsx`:
  - Focus a day in the initial month, then a controlled `month`+`selected` update (with `onSelect` wired up, so the selection is genuinely controlled): asserts focus lands on the newly selected day rather than `<body>`.
  - Same update without ever focusing a day first: asserts focus stays on `<body>` (i.e. the fix doesn't grab focus that wasn't in the grid).
- Verified the regression: reverting the source change turns the first new test from passing into failing, with `document.activeElement` on `<body>` — exactly the reported bug.
- Full package suite passes: `npx jest packages/react-day-picker` — 492 passing.
- `biome check` and `tsc --noEmit` both pass on the changed files.

### DIFF
--- a/packages/react-day-picker/src/DayPicker.test.tsx
+++ b/packages/react-day-picker/src/DayPicker.test.tsx
@@ -531,6 +531,67 @@ describe("when the `month` is changed programmatically", () => {
       expect(monthDates[monthDates.length - 1]).toEqual(new Date(2023, 1, 1));
     });
   });
+
+  describe("when the focused day is removed by the month change", () => {
+    test("moves focus to a day in the new month instead of losing it", () => {
+      const march = new Date(2024, 2, 15);
+      const november = new Date(2024, 10, 15);
+      const onSelect = jest.fn();
+
+      const { rerender } = render(
+        <DayPicker
+          mode="single"
+          month={march}
+          selected={march}
+          onSelect={onSelect}
+        />,
+      );
+
+      act(() => dateButton(march).focus());
+      expect(activeElement()).toBe(dateButton(march));
+
+      // A controlled update to both `month` and `selected`, as would happen
+      // e.g. after an external action changes the selection (see #3009).
+      rerender(
+        <DayPicker
+          mode="single"
+          month={november}
+          selected={november}
+          onSelect={onSelect}
+        />,
+      );
+
+      // Focus should land on the newly selected day, which calculateFocusTarget
+      // prioritizes above other days in the new month.
+      expect(activeElement()).toBe(dateButton(november));
+    });
+
+    test("does not move focus when it wasn't in the grid to begin with", () => {
+      const march = new Date(2024, 2, 15);
+      const november = new Date(2024, 10, 15);
+      const onSelect = jest.fn();
+
+      const { rerender } = render(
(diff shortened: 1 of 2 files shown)


raphyabak:
Closing this — #3009 was already fixed and merged independently via #3010. Thanks for taking care of it!

PR #3017, opened 2026-09-10
fix(types): allow undefined in ClassNames for exactOptionalPropertyTypes

### LINKED ISSUE #3015: Types: allow undefined in ClassNames values (classNames breaks under exactOptionalPropertyTypes)
### Summary

`ClassNames` types every slot as a required `string`, so the `classNames` prop (`classNames?: Partial<ClassNames>`) becomes `{ [key]?: string }`. Under `exactOptionalPropertyTypes: true`, a value of type `string | undefined` — which is what most class-name builders return (`clsx`, `cva`, `tailwind-variants`, and similar) — is **not** assignable to an optional-`string` property, so `classNames` fails to type-check with `TS2375`.

The sibling `Styles` type already allows `undefined` in its values; `ClassNames` does not. Aligning the two would fix the incompatibility and remove the inconsistency.

### Version

- `@daypicker/react` `10.0.1` (core `react-day-picker` `10.0.1`)
- TypeScript with `"exactOptionalPropertyTypes": true`

### Current types

```ts
// react-day-picker/dist/esm/types/shared.d.ts
export type ClassNames = {
  [key in UI | SelectionState | DayFlag | Animation]: string;
};

export type Styles = {
  [key in UI | SelectionState | DayFlag]: CSSProperties | undefined; // ← already allows undefined
};

// react-day-picker/dist/esm/types/props.d.ts
classNames?: Partial<ClassNames>;
```

### Reproduction

```tsx
// tsconfig.json → "compilerOptions": { "exactOptionalPropertyTypes": true }
import { DayPicker } from "@daypicker/react";

// what clsx()/cva()/tailwind-variants and similar helpers return:
const maybeClass: string | undefined = Math.random() > 0.5 ? "text-red-500" : undefined;

export function Example() {
  return <DayPicker classNames={{ root: maybeClass }} />;
  // TS2375: Type '{ root: string | undefined }' is not assignable to type
  // 'Partial<ClassNames>' with 'exactOptionalPropertyTypes: true'.
}
```

Without `exactOptionalPropertyTypes`, `?: string` widens to `string | undefined`, so this compiles. With it enabled, an optional `?: string` means "absent **or** a string, never an explicit `undefined`", so a `string | undefined` value is rejected.

### Proposed change

Allow `undefined` in the `ClassNames` values, mirroring `Styles`:

```ts
export type ClassNames = {
  [key in UI | SelectionState | DayFlag | Animation]: string | undefined;
};
```

This is a types-only change (no runtime impact) and makes `classNames` work under `exactOptionalPropertyTypes` and with the `string | undefined` values that class-name utilities commonly produce.

jafari1995:
## Summary
`ClassNames` typed every slot as `string`, so with `exactOptionalPropertyTypes` you couldn't pass `string | undefined` from helpers like clsx/cva. `Styles` already allows `undefined` — this lines `ClassNames` up with that.

Fixes #3015

## Test plan
- [x] Updated the `ClassNames` value type to `string | undefined`
- [x] Added a compile-time check in `props.test.tsx` using `string | undefined`
- [ ] CI typecheck


jafari1995:
Pushed a follow-up for the CI failures:

- Keep \`ClassNames\` values as \`string\` (avoids breaking internal \`classList\` usage)
- Widen the \`classNames\` prop to allow \`string | undefined\` under \`exactOptionalPropertyTypes\`
- Filter out \`undefined\` when merging with defaults

Local \`pnpm typecheck\` is green.

